PR #103, opened 2014-09-09
Setup the right version range of spring framework packages that  spring-data-redis can import

WillemJiang:
org.springframework.beans.*;version="[4.0.6, 4.0.6)", means we can only import the package with version is >= 4.0.6 and version < 4.0.6.  It makes spring-data-redis  bundle cannot import any version of spring beans packages. 

So I just change the bundle template file to let spring-data-redis can work with a rang of spring framework in the OSGi platform.


christophstrobl:
Should be fixed by moving to a Maven based build (see: b6e90752e7059453f0d6ac50fddb2332c52e1042) with placeholder resolvers.

PR #104, opened 2014-09-18
Update lettuce to 3.x

mp911de:
* Update lettuce redis client to latest snapshot, adopt changed API and use API for commands which were emulated using LUA script
* Implement Sentinel support using lettuce library, adopt finer grained exceptions and aligned exception behavior to match other libraries
* Added shutdown-timeout property in order to control shutdown duration of the netty framework which reduces test run duration
* Documentation updated
* Contains also a fix to use psetex instead of setex (typo)
* Switched `MinimumRedisVersionRule` to Jedis since Jedis has a much faster shutdown behavior than lettuce

Reference: DATAREDIS-348


mp911de:
@christophstrobl @thomasdarimont still interested in the lettuce 3.x upgrade or should I abandon this PR?

thomasdarimont:
Hi Mark,

sorry for the radio silence - we worked on different projects for a while and Christoph currently is and I will be on holiday for the next 2 weeks.
Would you be so kind to rebase this PR again on the latest SD Redis master?
I just tried to apply the PR as a patch but it failed due to some code shifting here and there that took place in the meantime :-(.

Although we focus on using the jedis driver in SD Redis internally, it would be helpful to have an updated lettuce  alternative at hand, especially given that the lettuce driver we currently use was last updated in 2013...
Since we still have Lettuce support in SD Redis and your fork is the most active and well maintained (and comes with support for Redis cluster) I'd vote to definitely upgrade to this.

@christophstrobl @olivergierke thoughts?

Cheers,
Thomas

olivergierke:
I am not too much in the details but it worries me that this PR seems to introduce a lot of dependencies to the Lattice driver in types that didn't have driver dependencies before. Is that by purpose?

mp911de:
I'm not quite sure which types you're talking about. Do you mean netty or the massive change within the import sections? 

I could not find types in which imports to `com.lambdaworks.redis` were added and that did not contain previously imports to `com.lambdaworks.redis`.

olivergierke:
Looks like the imports are ordered differently. They showed additions but the removals further below (which I oversaw). All good.

mp911de:
Rebased this PR to the most recent master. Added support for HLL and ZRANGEBYLEX

christophstrobl:
thanks @mp911de!

mp911de:
This PR could be closed, lettuce 3.0 was merged to master with #144. WDYT?

christophstrobl:
fixed in #144

PR #121, opened 2015-01-19
DATAREDIS-578 - made default hash/list/set/value/zset ops public

efenderbosch:
The behavior of the Default Hash/Set.List/Value/ZSet Operations can only be customized by creating classes in the org.springframework.data.redis.core package.

Making these classes public will allow people to customize the behavior using their own package structure.


mp911de:
Thanks for your PR. After considering this change for a while, we came to the conclusion that we won't expose our default Template API implementations. These contain too much API that isn't intended for outside use (related ticket: [DATAREDIS-578](https://jira.spring.io/browse/DATAREDIS-578)).

PR #122, opened 2015-01-28
DATAREDIS-369 - Avoid resource leak in RedisCache.

thomasdarimont:
In order to make sure that time-outed keys are removed from the set of known cache keys, we probe each key in the known cache keys set (setName) for existence during put / putIfAbsent operations and remove the keys from he known cache key set if redis doesn’t contain the key anymore.


thomasdarimont:
We use the following lua script to perfom the removal of potentially "dead" keys from the known cache keyset:

```lua 
local keys = redis.call('zrange', ARGV[1], 0, -1);
local keysCount = table.getn(keys);
local deletedCount = 0;
if(keysCount > 0) then
	for _, key in ipairs(keys) do
		if (not (redis.call('exists', key) == 1)) then
		   redis.call('zrem', ARGV[1], key);	
		   deletedCount = deletedCount + 1
		end;
	end;
end;
return deletedCount;
```

christophstrobl:
I do not think it's a good idea to do the cleanup on _put_ as the script execution takes quite some time and throttles performance by a factor of ~10.

src/main/java/org/springframework/data/redis/cache/RedisCache.java
```diff
@@ -84,190 +202,395 @@ public Object getNativeCache() {
...
-		System.arraycopy(keyBytes, 0, result, prefix.length, keyBytes.length);
+		@Override
+		public Void doInLock(RedisConnection connection) throws DataAccessException {
+
+			byte[] wildCard = "*".getBytes();
+			byte[] prefixToUse = Arrays.copyOf(metadata.getKeyPrefix(), metadata.getKeyPrefix().length + wildCard.length);
+			System.arraycopy(wildCard, 0, prefixToUse, metadata.getKeyPrefix().length, wildCard.length);
 
-		return result;
+			connection.eval(REMOVE_KEYS_BY_PATTERN_LUA.getBytes(), ReturnType.INTEGER, 0, prefixToUse);
```
christophstrobl: Use `StringRedisSerializer` to ensure property serialization to binary representation.

src/main/java/org/springframework/data/redis/cache/RedisCache.java
```diff
@@ -84,190 +202,395 @@ public Object getNativeCache() {
...
+			super(metadata);
+			this.metadata = metadata;
 		}
 
-		byte[] result = Arrays.copyOf(prefix, prefix.length + keyBytes.length);
-		System.arraycopy(keyBytes, 0, result, prefix.length, keyBytes.length);
+		@Override
+		public Void doInLock(RedisConnection connection) throws DataAccessException {
+
+			byte[] wildCard = "*".getBytes();
```
christophstrobl: Use `StringRedisSerializer` to ensure property serialization to binary representation.

src/main/java/org/springframework/data/redis/cache/RedisCacheKey.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * @author Christoph Strobl
+ * @since 1.5
+ */
+public class RedisCacheKey {
+
+	private final Object keyElement;
+	private byte[] prefix;
+	@SuppressWarnings("rawtypes")//
+	private RedisSerializer serializer;
```
christophstrobl: Favor usage of an `Accessor` over directly using the the `RedisSerializer`.

PR #123, opened 2015-01-30
DATAREDIS-328 - RedisCacheManager should not instantiate caches in setCacheNames().

thomasdarimont:
We now construct the caches for the configured cache-names in `afterPropertiesSet()`.
Previously the caches were created in the setter which lead to unwanted property-set order dependencies.


thomasdarimont:
There is still one test case failing -  org.springframework.data.redis.cache.RedisCacheManagerUnitTests > testPuttingCacheManagerIntoStaticModeShouldNotRemoveAlreadyRegisteredCaches FAILED

Let's have a talk about this later

thomasdarimont:
Merged into 
1.5.x (master) via 0ea0c4bf673de563415a588bc4fdcd7d5cef9e7e
1.4.x via 76c07f58323ad1d11270028c4c9b434c8b743d6f
1.3.x via 90a96fe40eb6691e115ce9bd35a0b2e0fd20c00f

PR #127, opened 2015-02-13
DATAREDIS-374 - Adapt JedisConnectionFactory for changes in Jedis 2.7.

thomasdarimont:
Previously we didn’t support Jedis 2.7 since the {{getTimeout}} / {{setTimeout}} Methods on {{JedisShardInfo}} to configure the {{SoTimeout}} were replaced by {{getSoTimeout}} and {{setSoTimeout}} respectively.

We now detect the appropriate Methods via Reflection in JedisConnectionFactory.

Tested against Jedis 2.7 branch with Redis 3.0.0.RC4.


christophstrobl:
Am I wrong or is jedis 2.7 still under development... Besides that the change https://github.com/xetorthio/jedis/commit/3032c9410d9e8515a75ba6a516a3669324aec879 involved more than just `JedisShardInfo` but also `Connection` where the timeout methods have been renamed as well. Did you check for potential invocations of those?

HeartSaVioR:
We're forcusing our plan to 3.0.0 and we don't discuss about releasing 2.7 yet.
We can maintain 2.x so 2.7 could be released sometimes, but seems like not several weeks.

christophstrobl:
thanks @HeartSaVioR !

PR #129, opened 2015-03-04
DATAREDIS-315 - Add support for RedisCluster.

christophstrobl:
Allow cluster to be configured using `RedisClusterConfiguration` via `RedisConnectionFactorfy`.

Jedis is currently the only supported driver for cluster freatures. 

This is an intial version get the conversation going.


src/test/java/org/springframework/data/redis/connection/jedis/JedisClusterConnectionTests.java
```diff
@@ -0,0 +1,880 @@
...
+	static final byte[] KEY_1_BYTES = JedisConverters.toBytes(KEY_1);
+	static final byte[] KEY_2_BYTES = JedisConverters.toBytes(KEY_2);
+
+	static final String VALUE_1 = "value1";
+	static final String VALUE_2 = "value2";
+
+	static final byte[] VALUE_1_BYTES = JedisConverters.toBytes(VALUE_1);
+	static final byte[] VALUE_2_BYTES = JedisConverters.toBytes(VALUE_2);
+
+	JedisCluster nativeConnection;
```
thomasdarimont: We should use `jedisCluster` here instead of `nativeConnection`.

thomasdarimont:
We should add test cases that show how to use `RedisClusterConnection` in combination with a `RedisTemplate`.

HeartSaVioR:
Just 2 cents, Jedis changes strategy of handling resource cleanup (including returning to pool) to just call close() on Jedis, JedisCluster instance directly. Old strategy has been discarded completely.
I wish Spring Data Redis to follow up new strategy, and report any issues with new strategy.

christophstrobl:
thanks @HeartSaVioR (especially for merging cluster-revised branch into jedis!) already saw that and tried to do so but this broke quite some tests - that's why we switched back to the reflective call - still I will have to check what happens exactly when we use `close()`...

PR #132, opened 2015-03-18
DATAREDIS-383 - Improve JavaDoc of RedisCacheManager.

snicoll:
`setCacheNames` allow to restore the mode to dynamic with a null/empty collection argument but the javadoc does not state that.


thomasdarimont:
Thanks cherry-picked your change via d0a0a6a!

PR #134, opened 2015-03-20
DATAREDIS-332 - @Transactional operation should use the same connection.

thomasdarimont:
We only unbind the current redisConnectionFactory from the current thread if the 
transaction is not active anymore. 
Previously we performed the unbinding in RedisTemplate.execute(RedisCallback<T>, boolean, boolean) overtime which lead to the fact that the next redis operation within the same transaction 
used a different redis connection, which could lead to exhaustion of the connection pool.
We also ensure that we correctly unbind the RedisConnectionFactory resource from the TransactionSynchronizationManager on transaction completion in RedisTransactionSynchronizer.afterCompletion(int).
Added test case supplied from issue.


thomasdarimont:
CP'ed 
into master via cf4fa9e4fba68d5a857603d7a3229723a3b71518
into 1.5.x via 4d9b71dcc2af7807bc4bb76e37248a2cc5522b2b
into 1.4.x via d696a2feffec58fcd6aa641768271c5dcfe1ea33

PR #136, opened 2015-04-07
DATAREDIS-390 - Improve Support for JSON Serialization.

thomasdarimont:
Introduced GenericJackson2RedisSerializer that is capable of serialising arbitrary java objects into redis without a priority knowledge of the used types. This is achieved by encoding the actual type information with the value. Previously users had to use Jackson2RedisSerializer which only supported serialising one type.


christophstrobl:
fixed in #145

jingege:
We do need this!But it's likely not accepted, pity~

christophstrobl:
@jingege A `Jackson2` based `RedisSerializer` suitable for more than one type has already been added via 35424bb9da606f5d9b40460da793bc7fdd9a9d5c.

PR #137, opened 2015-04-08
DATAREDIS-378 - Add support for ZRANGEBYLEX. 

christophstrobl:
We now support `ZRANGEBYLEX` on `RedisConnection` and `StringRedisConnection` when using the Jedis driver. The upper and lower bounds can be defined using the `Range` type which will be converted to the according binary representation including prefix and/or infinite operators.

```
range().gt("a")          => (a +
range().gte("a")         => [a +
range().lt("z")          => - (z
range().lte("z")         => - [z
range().gte("a").lt("z") => [a (z
```


thomasdarimont:
Merged into master via Merged via ff37da51977f0d86b09e176ffa5785f5ab7533fc.

PR #138, opened 2015-04-27
DATAREDIS-352 - Enhance range options for ZSET.

christophstrobl:
We now allow usage of `Range` and `Limit` types (newly introduced for `ZRANGEBYLEX`) on mostly all `ZSET` range operations.


src/main/java/org/springframework/data/redis/connection/jedis/JedisConnection.java
```diff
@@ -2194,33 +2212,120 @@ public Long zInterStore(byte[] destKey, byte[]... sets) {
...
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * @see org.springframework.data.redis.connection.RedisZSetCommands#zRangeByScoreWithScores(byte[], org.springframework.data.redis.connection.RedisZSetCommands.Range, org.springframework.data.redis.connection.RedisZSetCommands.Limit)
+	 */
+	@Override
+	public Set<Tuple> zRangeByScoreWithScores(byte[] key, Range range, Limit limit) {
+
+		Assert.notNull(range, "Range cannot be null for ZRANGEBYSCOREWITHSCORES.");
```
thomasdarimont: we should check limit here as well.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnection.java
```diff
@@ -2194,33 +2212,120 @@ public Long zInterStore(byte[] destKey, byte[]... sets) {
...
+
+		Assert.notNull(range, "Range cannot be null for ZRANGEBYSCORE.");
+
+		byte[] min = JedisConverters.boundaryToBytesForZRange(range.getMin(), JedisConverters.toBytes("-inf"));
+		byte[] max = JedisConverters.boundaryToBytesForZRange(range.getMax(), JedisConverters.toBytes("+inf"));
+
 		try {
 			if (isPipelined()) {
-				pipeline(new JedisResult(pipeline.zrangeByScore(key, min, max)));
+				if (limit != null) {
```
thomasdarimont: Perhaps we should introduce a Null Object like: Limit.ALL or Limit.NONE to get rid of those null checks.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnection.java
```diff
@@ -2194,33 +2212,120 @@ public Long zInterStore(byte[] destKey, byte[]... sets) {
...
+	/*
+	 * (non-Javadoc)
+	 * @see org.springframework.data.redis.connection.RedisZSetCommands#zRangeByScoreWithScores(byte[], org.springframework.data.redis.connection.RedisZSetCommands.Range, org.springframework.data.redis.connection.RedisZSetCommands.Limit)
+	 */
+	@Override
+	public Set<Tuple> zRangeByScoreWithScores(byte[] key, Range range, Limit limit) {
+
+		Assert.notNull(range, "Range cannot be null for ZRANGEBYSCOREWITHSCORES.");
+
+		byte[] min = JedisConverters.boundaryToBytesForZRange(range.getMin(), JedisConverters.toBytes("-inf"));
```
thomasdarimont: could create a constant for -inf and +inf in JedisConverters.

thomasdarimont:
cp'ed via 283a9638f2bbe2854e8773c21ee0159ebd01f4db.

PR #139, opened 2015-04-29
Improve compatibility with jedis 2.7.1

quaff:
jedis 2.7.1 changed method signature casue spring data redis NPE

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 slf4jVersion=1.7.10
 junitVersion=4.12
 jredisVersion=06052013
-jedisVersion=2.7.0
+jedisVersion=2.7.2
 springVersion=4.1.6.RELEASE
 springDataBuildVersion=1.7.0.BUILD-SNAPSHOT
 log4jVersion=1.2.17
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnection.java
@@ -68,6 +68,7 @@ import redis.clients.jedis.Jedis;
 import redis.clients.jedis.Pipeline;
 import redis.clients.jedis.Protocol;
 import redis.clients.jedis.Protocol.Command;
+import redis.clients.jedis.ProtocolCommand;
 import redis.clients.jedis.Queable;
 import redis.clients.jedis.Response;
 import redis.clients.jedis.ScanParams;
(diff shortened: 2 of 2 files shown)


HeartSaVioR:
Need to explain why signature changes happen though I didn't participate to implement Jedisque.

First of all, Connection is internal class of Jedis, so we don't recommend outers to use it directly.
We didn't think about reflection, and it's internal class so we think it should be safe.

Jedisque has its command set, so in order to use Jedis from Jedisque, we introduced an interface -  ProtocolCommand which replaces Command so that "command set" could be extended.
In order to let Connection able to send new command set, sendCommand methods signature need to be changed.

HeartSaVioR:
@christophstrobl @thomasdarimont 
Could you take a look at this PR? Or did Spring-data-redis team already take care of it?

christophstrobl:
@HeartSaVioR will have a look. thanks for reminder!

HeartSaVioR:
@christophstrobl 
We found that we're breaking semver (https://github.com/xetorthio/jedis/issues/1043#issuecomment-119036187), and we're discussing about revert the change. 
When we revert and release 2.7.3, spring-data-redis could be safe without applying this PR.
Sorry for annoying and thanks!

christophstrobl:
@HeartSaVioR thanks - yeah I think it's different in the jedis 2.8 branch - will wait and see how things turn out.

marcosnils:
Hi, Jedis 2.7.3 has been just release to fix this issue. Can someone here try and let us knows how it goes?

quaff:
@marcosnils , jedis 2.7.3 works fine, thanks.

marcosnils:
@quaff amazing. @christophstrobl you can go ahead and close this issue :)

christophstrobl:
@marcosnils will need to apply a slightly modified fix so one is able to run sd-redis with _2.7.2_ or _2.7.3_.  Currently we've several tests (unrelated to this issue) failing using _2.7.3_

christophstrobl:
@marcosnils @HeartSaVioR @quaff created #149 to tackle upgarde issues. We now perform a safe lookup of either `ProtocolCommand` or `Protocol$Command` for usage with `Connection#sendCommand` which allows us to use 2.7.1, 2.7.2 and 2.7.3.

Concerning the test failures I found that 
```java
JedisPool pool = new JedisPool();

Jedis jedis = pool.getResource();
Transaction transaction = new Transaction(jedis.getClient());

jedis.multi();
transaction.set("foo".getBytes(), "bar".getBytes());
transaction.exec();

pool.returnResource(jedis);
```

Will cause `returnResource(jedis)` to fail with 
```
redis.clients.jedis.exceptions.JedisException: Could not return the resource to the pool
Caused by: redis.clients.jedis.exceptions.JedisDataException: ERR DISCARD without MULTI
```

Changing to `Transaction transaction = jedis.multi()` fixed it.

PR #141, opened 2015-06-01
DATAREDIS-402 - RedisCache should not expire known keys in case ttl of element is not set.

christophstrobl:
We now skip calling `EXPIRE` on known keys set in case element is eternal.


thomasdarimont:
Cp'ed into master aa04e73dbddd48105c2ffe037d05b898297fbec7

PR #143, opened 2015-06-04
DATAREDIS-399 Make RedisCache easier for extending

marko-stankovic:
This change is setting visibility of RedisCache inner classes to
protected, thus making them more accessible for subclassing. Also,
generation of RedisCacheKey object has been moved to separate method so
that it can be reused from subclasses.

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -210,6 +210,14 @@ public class RedisCache implements Cache {
 				template.getKeySerializer());
 	}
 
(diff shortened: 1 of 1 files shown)


marko-stankovic:
I have also made comment in related ticket on what is a motivation for this change https://jira.spring.io/browse/DATAREDIS-399?focusedCommentId=116317&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-116317

mp911de:
Thank you for your PR. We started working on Spring Data Redis 2.0 where `RedisCache` will be improved. It does not make much sense merging this PR to `master` as things are going to change anyway.

PR #145, opened 2015-06-16
DATAREDIS-392 - Add Gerneric Jackson2 RedisSerializer. 

christophstrobl:
We introduced `GenericJackson2JsonRedisSerializer` holding a preconfigured `ObjectMapper` which writes class type informations into the JSON structure. This enables polymorphic deserialization and allows `RedisCacheManager` to operate upon a `RedisTemplate` storing data in JSON format, which had until this only been possible using the `JdkSerializer`.

----
Alternative Implementation for #136


src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	/**
+	 * Setting a custom-configured {@link ObjectMapper} is one way to take further control of the JSON serialization
+	 * process. For example, an extended {@link SerializerFactory} can be configured that provides custom serializers for
+	 * specific types.
+	 * 
+	 * @param mapper must not be {@literal null}.
+	 */
+	public GenericJackson2JsonRedisSerializer(ObjectMapper mapper) {
```
olivergierke: Do we need an additional overload that takes an `ObjectMapper` *and* the type key?

src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
```diff
@@ -0,0 +1,122 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectMapper.DefaultTyping;
+import com.fasterxml.jackson.databind.ser.SerializerFactory;
+
+/**
+ * @author Christoph Strobl
+ * @since 1.6
+ */
+public class GenericJackson2JsonRedisSerializer implements RedisSerializer<Object> {
```
olivergierke: Does that type need to be registered anywhere or is it the users' task to do so?

src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
christophstrobl: which benefit would we offer by adding that one? If we offer that we'd need to take care of potential configuration already done on the mapper before handing it to the serializer. I'd rather leave it completely to the user to configure the mapper when using this option.

src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
christophstrobl: it's up to the user to configure that one on the `RedisTemplate`. By default it's the `JdkSerializationRedisSerializer`.

kgdngitfk:
I have read the source code and the data store into redis,I think the GenericJackson2JsonRedisSerializer
seri serialize(Object val) convert val to JSON with the key "@class": "com.xx.xx.Data" which value use for 
deserialize() to the corresponding java Type,isn't it?

christophstrobl:
@kgdngitfk yes. It's using an `ObjectMapper` having `defaultTyping` enabled.

PR #147, opened 2015-07-17
Move serialization out of doInRedis calls.

jayanderson:
This allows the pooled connection to be returned quicker so more threads can be
serviced.

This change adds RedisCacheKeyBuilder and makes RedisCacheKey fully immutable. This eager serialization of the key avoids calling it multiple times and ensures this work is done while not holding onto a connection.

The putIfAbsent call was not changed to deserialize outside the redis call. It would require a bit more rearrangement. Let me know if you'd like me to give that a go.

See also https://jira.spring.io/browse/DATAREDIS-414.


jebaney:
+1

christophstrobl:
fixed in #153

PR #149, opened 2015-07-21
DATAREDIS-412 - Assert compatibility with jedis 2.7.1, 2.7.2 and 2.7.3.

christophstrobl:
Bump jedis version to 2.7.3 and fix `Connection#sendCommand` method args type differences between versions.

We now also lazily init `Transaction` on first call of `RedisConnection#multi()` since pre initialization causes issues releasing jedis connections and returning them to pool.

Supersedes: #139


HeartSaVioR:
@christophstrobl Really great! Thanks for fixing this.

PR #161, opened 2015-12-29
Fix a typo

tsachev:
Misleading error message. There is no JMS involed.


christophstrobl:
That's merged. Thanks @tsachev!

PR #170, opened 2016-02-16
DATAREDIS-316 - Add EX/PX, NX/XX options to SET command.

christophstrobl:
We now support `EX`/`PX` and `NX`/`XX` arguments along with the `SET` command for both [xetorthio/jedis](https://github.com/xetorthio/jedis) and [mp911de/lettuce](https://github.com/mp911de/lettuce). 

Jedis 2.8 does not support all combinations of `EX`/`PX` and `NX`/`XX`. To work around those limitations we delegate to the according set methods to execute related commands or fail fast if there is no way for delegation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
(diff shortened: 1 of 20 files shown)


mp911de:
Merged with 1151bcf13689

PR #172, opened 2016-02-22
DATAREDIS-433 - Document dependencies for Redis serializers.

mp911de:
Add maven dependencies examples for Redis serializers. Fix code highlighting syntax.


mp911de:
Won't fix...

PR #173, opened 2016-02-24
DATAREDIS-468 - Guard muli/exec bocks in RedisCache to allow usage with cluster.

christophstrobl:
We now guard `muli` / `exec` blocks by checking the connection type. This allows to skip those commands when running in cluster.


mp911de:
Merged with 0f035982b209bba667bf1921269d0db557940df3

PR #174, opened 2016-02-25
DATAREDIS-467 - Fix cluster multi key commands.

christophstrobl:
We moved away from returning raw map types and now return dedicated objects for command executions in cluster environment. This allows to maintain node information and collecting results from each and every callback.


mp911de:
Merged with 02742e3b4921cdda886a9455839bc5f691b2d0af

PR #180, opened 2016-03-17
DATAREDIS-480: Added support for all SSL options

balazsmaria:
----------------------
branch [[lettuce-connection-factory-no-verify-option]]


mp911de:
Thanks, that's merged with b3243ff.

PR #182, opened 2016-03-24
DATAREDIS-469 - Throw DataRetrievalFailureException for atomic numbers when key is removed.

christophstrobl:
We now throw `DataRetrievalFailureException` in case of `get()` when the underlying key storing the value of an `AtomicInteger`, `AtomicLong` or `AtomicDouble` is removed from Redis.


mp911de:
Merged and backported.

PR #189, opened 2016-04-11
DATAREDIS-492 - Handle serializing arrays and collections

gregturn:
When serializing complex structures that include lists AND arrays, Spring Data Redis breaks down by assuming all "collectionLike" structures implement the Collection interface. This commit includes a test case that reproduces this problem along with a patch to handle both collections and arrays that both identify as "collectionLike".


christophstrobl:
thanks @gregturn thats fixed.

PR #191, opened 2016-04-25
DATAREDIS-471 - Add support for partial updates.

christophstrobl:
We now allow partial update of domain types via PartialUpdate. The according expiration times and secondary index structures are updated accordingly.

In some cases it is not necessary to load and rewrite the entire entity just to set a new value within it. A session timestamp for last active time might be such a scenario where you just want to alter one property.
`PartialUpdate` allows to define `set`, `delete` actions on existing objects while taking care of updating potential expiration times of the entity itself as well as index structures.

```java
PartialUpdate<Person> update = new PartialUpdate<Person>("e2c7dcee", Person.class)
  .set("firstname", "mat")                                                           <1>
  .set("address.city", "emond's field")                                              <2>
  .del("age");                                                                       <3>

template.update(update);

update = new PartialUpdate<Person>("e2c7dcee", Person.class)
  .set("address", new Address("caemlyn", "andor"))                                   <4>
  .set("attributes", singletonMap("eye-color", "grey"));                             <5>

template.update(update);

update = new PartialUpdate<Person>("e2c7dcee", Person.class)
  .refreshTtl(true);                                                                 <6>
  .set("expiration", 1000);

template.update(update);
```

```bash
<1> Set the simple property _firstname_ to _mat_
<2> Set the simple property _address.city_ to _emond's field_ without having to pass in the entire object. This does not work when a custom conversion is registered.
<3> Remove the property _age_.
<4> Set complex property _address_.
<5> Set a map/collection of values removes the previously existing map/collection and replaces the values with the given ones.
<6> Automatically update the server expiration time when altering time to live.
```

**NOTE:** Updating complex objects as well as map/collection structures requires further interaction with Redis to determine existing values which means that it might turn out that rewriting the entire entity might be faster.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/PartialUpdate.java
+++ b/src/main/java/org/springframework/data/redis/core/PartialUpdate.java
@@ -114,6 +114,7 @@ public class PartialUpdate<T> {
 		PartialUpdate<T> update = new PartialUpdate<T>(this.id, this.target, this.value, this.refreshTtl,
 				this.propertyUpdates);
 		update.propertyUpdates.add(new PropertyUpdate(UpdateCommand.SET, path, value));
+
 		return update;
 	}
 
@@ -130,6 +131,7 @@ public class PartialUpdate<T> {
 		PartialUpdate<T> update = new PartialUpdate<T>(this.id, this.target, this.value, this.refreshTtl,
 				this.propertyUpdates);
 		update.propertyUpdates.add(new PropertyUpdate(UpdateCommand.DEL, path));
+
 		return update;
 	}
 
--- a/src/main/java/org/springframework/data/redis/core/PartialUpdate.java
+++ b/src/main/java/org/springframework/data/redis/core/PartialUpdate.java
@@ -1,0 +1,236 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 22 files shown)


src/main/java/org/springframework/data/redis/core/PartialUpdate.java
```diff
@@ -0,0 +1,232 @@
...
+	/**
+	 * Remove the value reachable via given {@literal path}.
+	 *
+	 * @param path path must not be {@literal null}.
+	 * @return a new {@link PartialUpdate}.
+	 */
+	public PartialUpdate<T> del(String path) {
+
+		Assert.hasText(path, "Path to remove must not be null or empty!");
+		propertyUpdates.add(new PropertyUpdate(UpdateCommand.DEL, path));
```
mp911de: `PartialUpdate` changes its own state and creates a new instance.

src/main/java/org/springframework/data/redis/core/convert/MappingRedisConverter.java
```diff
@@ -357,7 +378,118 @@ public void write(Object source, final RedisData sink) {
...
+
+			String path = pUpdate.getPropertyPath();
+
+			if (UpdateCommand.SET.equals(pUpdate.getCmd())) {
+
+				KeyValuePersistentProperty targetProperty = getTargetPropertyOrNullForPath(path, update.getTarget());
+
+				if (targetProperty == null) {
 
+					writeInternal(entity.getKeySpace(), pUpdate.getPropertyPath(), pUpdate.getValue(),
```
christophstrobl: add a type check and deny eg. setting `foo-bar` as `String` for a `Long` property.

src/main/java/org/springframework/data/redis/core/RedisKeyValueAdapter.java
```diff
@@ -393,6 +394,109 @@ public Long doInRedis(RedisConnection connection) throws DataAccessException {
 		return count != null ? count.longValue() : 0;
 	}
 
+	public void update(final PartialUpdate<?> update) {
```
mp911de: Old index references remain in index-helper index `(entity):(key):idx` when updating/deleting `@Indexed` properties

mp911de:
That's merged with addcdfe.

PR #192, opened 2016-04-26
DATAREDIS-501 - Use application context ClassLoader as default for JdkSerializationRedisSerializer in RedisTemplate.

christophstrobl:
We now pick up the `ClassLoader` from the `ApplicationContext` and use the latter as default in `RedisTemplate` for initializing the `JdkSerializationRedisSerializer`. We only do this in case the default serializer has not been set explicitly.

----

Should also be back ported to `1.7.x` which might require updating javadoc `@since` which currently states 1.8.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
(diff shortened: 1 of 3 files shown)


mp911de:
That's merged with f5ecd3a.

PR #194, opened 2016-04-29
DATAREDIS-503 - Add HashMapper implementation based on MappingRedisConverter.

christophstrobl:
We now support mapping of simple and complex types to Redis _HASH_ structures applying the same structure as the Repository support. This allows `Object` to hash mapping and its direct usage via `RedisTemplate` without the need of explicitly having to use the repository abstraction.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>1.8.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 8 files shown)


christophstrobl:
@mp911de I left the generic type out by intention as it suggests the mapper is only supposed to to be usable with one specific type which is not the case. The returned type is actually defined via `_class` in the `Map` itself. Unfortunately we cannot change the `HashMapper` interface without breaking to much.

mp911de:
Rolled back the generic type change. Thanks for checking.

mp911de:
That's merged with c1476b7.

PR #197, opened 2016-05-10
DATAREDIS-423 - Add support for Jackson2 based HashMapper.

christophstrobl:
Added FasterXML Jackson `ObjectMapper` based `HashMapper` implementation that allows flattening. 

```java
class Person {
  String firstname;
  String lastname;
  Address address;
}

class Address {
  String city;
  String country;
}
```

```bash
### NON Flat 
firstname:Jon
lastname:Snow
address:{ city : Castle Black, country : The North }

###  Flat 
firstname:Jon
lastname:Snow
address.city:Castle Black
address.country:The North
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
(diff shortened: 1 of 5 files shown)


mp911de:
Merged with d097a64.

PR #198, opened 2016-05-18
DATAREDIS-512 - Skip repository index update checks on insert.

christophstrobl:
We now skip checking for existing index values on insert.


mp911de:
Merged with 46f07d7056e7b3cd1f15ba1c060f007f2ab40692.

PR #199, opened 2016-05-19
DATAREDIS-513 - Fix RedisServerCommands.time() failure when in pipeline mode.

christophstrobl:
We fixed a glitch in Jedis/Lettuce RedisConneciton `TIME` command when used in pipeline or transaction mode.

----

Should be back ported to `1.7.x` and `1.6.x`.


mp911de:
Merged and backported.

PR #200, opened 2016-05-25
DATAREDIS-514 - Unify geoRadius capitalization. 

christophstrobl:
Unify capitalization of `geoRadius` to match other `geo` prefixed operations.


mp911de:
Merged with 2b51966.

PR #201, opened 2016-06-14
DATAREDIS-510 - Fix caching of null values. 

christophstrobl:
We now no longer add empty `byte[]` as cache value but return `null` instead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>1.8.0.BUILD-SNAPSHOT</version>
+	<version>1.8.0.DATAREDIS-510-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -686,10 +686,14 @@ public class RedisCache implements Cache {
 				connection.multi();
 			}
 
-			connection.set(element.getKeyBytes(), element.get());
+			if (element.get().length == 0) {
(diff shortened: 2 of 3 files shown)


mp911de:
That's merged and backported to `1.7.x`.

djwoodward:
I'm curious why this change was made? We have to use 1.7.1 and are unable to update to 1.7.2. because this breaks caching of null values.

gregturn:
@djwoodward How does this break you? The unit tests demonstrate that caching null with the API will then yield a null when fetched. It simply does so without stuffing some awkward value into Redis. Is there a different behavior you are leaning on right now other than fetching nulls from a cache?

djwoodward:
The behavior we are seeing is with 1.7.1 when a cached method returns null it issues a cache put operation and we see an add command in the redis console. With 1.7.2 no cache put is issued and nothing is ever added to the cache. Effectively not caching nulls.

djwoodward:
@gregturn Here is a simple example which demonstrates the behavior difference between 1.7.1 and 1.7.2. This used to cache null, until a cache evict or additional cache put. Now if this method returns null and a cache delete is sent to redis (even though it's not actually in the cache), and nothing is cached. In this example the ``someExpensiveCall()`` method is repeatedly called while our intentions are to cache it's value.
```
    @Cacheable(cacheNames="PizzaCache")
    public Pizza retrievePizza() {
        // Might return null or a Pizza object, either way this call is expensive
        Pizza pizza = repository.someExpensiveCall();
        return pizza;
    }
```

ahilsend:
I agree with @djwoodward . We have the same issue right now. Just updated to spring boot 1.4, but had to go back a version for spring-data-redis to 1.7.1 because of this issue. 

We have the exact same use case, we want to cache some expensive operations ... which are allowed to return null in the end. In that case we don't want to redo that operation all the time. Thus writing null into the cache is a valid use-case.

PR #202, opened 2016-06-16
DATAREDIS-505 - Allow configuration of Redis Keyspace events config value.

christophstrobl:
We now allow setting the desired `notify-keyspace-events` server configuration via `@EnableRedisRepsoitories`. Use `null` or an empty String for not touching the server’s config. The default is `Ex`.


mp911de:
That's merged and backported to `1.7.x`.

PR #205, opened 2016-06-23
DATAREDIS-526

oudb:
The redis command TTL and PTTL should return negative number(-1 or -2) when the key does not exist or the key exist but has no associated expire.
But the function getExpire like this: redisTemplate.getExpire(key1, TimeUnit.SECONDS),the return value will be 0 and redisTemplate.getExpire(key1) will be -1 or -2.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
+++ b/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
@@ -715,11 +715,14 @@ public class RedisTemplate<K, V> extends RedisAccessor implements RedisOperation
 		return execute(new RedisCallback<Long>() {
 
 			public Long doInRedis(RedisConnection connection) {
+				Long expire;
 				try {
-					return timeUnit.convert(connection.pTtl(rawKey), TimeUnit.MILLISECONDS);
+					expire = connection.pTtl(rawKey);
+					return expire < 0 ? expire : timeUnit.convert(expire, TimeUnit.MILLISECONDS);
 				} catch (Exception e) {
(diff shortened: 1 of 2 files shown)


oudb:
@christophstrobl Hi,could you help me review the code?

mp911de:
Can you please sign, if you haven't done already, the [CLA form](https://support.springsource.com/spring_committer_signup) for `Spring Data Access` with project lead `Oliver Gierke`?

oudb:
@mp911de Yes,I have signed.my cla Number: 143520151016081625 (Duobiao Ou)

src/main/java/org/springframework/data/redis/core/RedisTemplate.java
```diff
@@ -715,11 +715,14 @@ public Long getExpire(K key, final TimeUnit timeUnit) {
 		return execute(new RedisCallback<Long>() {
 
 			public Long doInRedis(RedisConnection connection) {
```
mp911de: `getExpire(K, TimeUnit)` has a bug if it's used with transactions/pipelining. It throws in the first place a `NullPointerException`. If the call is guarded against `null` access the results return the original `ttl`/`pTtl` value and not the converted one. I think we should move the conversion into the connections as that's the place where we're able to perform a conversion. Adding `pTtl(K, TimeUnit)` and `ttl(K, TimeUnit)` to `RedisKeyCommands` should fix that bug.

mp911de:
That's merged and polished with 5a3a7549856011970f80a8f2d64b66274064e748. We also discovered a bug when using transactions/pipelining and fixed it together with this PR.

PR #207, opened 2016-07-05
DATAREDIS-530 - Fix PartialUpdate removing existing indexes.

christophstrobl:
We now make sure to leave existing indexes untouched when using `PartialUpdate`. 
We also fixed a glitch, where index values have not been removed correctly when saving entities with `null` values, along the way.


mp911de:
That's merged with f5a506c.

PR #208, opened 2016-07-08
DATAREDIS-523 - Read back TTL into property.

christophstrobl:
We now read back `TTL` value from Redis for explicitly `@TimeToLive` annotated properties.


mp911de:
That's merged with fd6e4d1.

PR #212, opened 2016-07-26
DATAREDIS-539 - Replace call to deprecated addCache in RedisCacheManager.

christophstrobl:
We now rely on `getMissingCache` and `loadCaches` for initialization instead of explicitly calling deprecated `addCache` method on `AbstractCacheManager`. This changes the behavior of `createAndAddCache` which has been deprecated as of now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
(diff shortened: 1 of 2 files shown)


mp911de:
@snicoll do you mind having a look at the PR?

snicoll:
LGTM

mp911de:
That's merged with b47fc35.

PR #216, opened 2016-08-25
DATAREDIS-547 - Fix query execution when derived criteria is empty.

christophstrobl:
We now make sure to pipe finder queries without any criteria to the according find all method. This allows usage of `PagingAndSortingRepository.findAllBy(Pageable page)` as well as finders without any criteria like `findTop3By()`.

---

Depends on: spring-projects/spring-data-keyvalue#22


mp911de:
That's merged.

PR #217, opened 2016-08-30
DATAREDIS-531 - Fix deferred ScanCursor command execution using Jedis.

mp911de:
Previously, `ScanCursor` held a reference to the Jedis connection which was used during ScanCursor creation. Cursors can be used outside RedisTemplate that leads to releasing/closing the initial connection. The retained reference used a released connection.

We now check whether the connection is still open. If so, we reuse the connection. If the connection is closed, we use the associated connection factory to obtain a connection to invoke `SCAN` and release the reference after command execution.

----

Related ticket: [DATAREDIS-531](https://jira.spring.io/browse/DATAREDIS-531)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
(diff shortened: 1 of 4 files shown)


mp911de:
Superseded by #218.

PR #224, opened 2016-10-06
DATAREDIS-542 - Expire keys in RedisCache.putIfAbsent.

mp911de:
RedisCache now expires keys using putIfAbsent if the key was set. Previously, the key was only expired if the value was already present and the value matched the value stored inside of Redis.

----

Should be backported to _1.6.x_ and _1.7.x_.

Related ticket: [DATAREDIS-542](https://jira.spring.io/browse/DATAREDIS-542)
See also: https://github.com/spring-projects/spring-data-redis/pull/150

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -17,7 +17,6 @@
 package org.springframework.data.redis.cache;
 
 import static org.springframework.util.Assert.*;
-import static org.springframework.util.ObjectUtils.*;
 
 import java.lang.reflect.Constructor;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/redis/cache/RedisCache.java
```diff
@@ -720,20 +720,27 @@ public RedisCachePutIfAbsentCallback(BinaryRedisCacheElement element, RedisCache
...
+			boolean keyMaintenance;
+
+			byte[] keyBytes = element.getKeyBytes();
+			byte[] value = element.get();
+
+			if (connection.setNX(keyBytes, value)) {
+				keyMaintenance = true;
+			} else {
+				existingValue = connection.get(keyBytes);
+				keyMaintenance = ObjectUtils.nullSafeEquals(value, existingValue);
```
mp911de: Not sure about the reason for expiry/key maintenance if the key was not set but the same content.

PR #230, opened 2016-12-11
DATAREDIS-581: waitForLock() cannot be interrupted

ljrmorgan:
As mentioned in https://jira.spring.io/browse/DATAREDIS-581, here is a failing unit test for this issue.

I think the cleanest way to fix this would be to throw a `DataAccessException` after setting the `interrupted` flag on the thread in the catch block in `waitForLock()`, but I wanted to run that solution by you before implementing it.


mp911de:
This was fixed with [DATAREDIS-481](https://jira.spring.io/browse/DATAREDIS-481).

PR #231, opened 2016-12-12
DATAREDIS-582: Make locking in RedisCache optional

ljrmorgan:
Work in progress for making cache locking optional. Commiting to gather feedback. Still to do: add tests, update documentation.

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -50,6 +50,7 @@ import org.springframework.util.ObjectUtils;
  * @author Christoph Strobl
  * @author Thomas Darimont
  * @author Mark Paluch
+ * @author Louis Morgan
  */
 @SuppressWarnings("unchecked")
 public class RedisCache extends AbstractValueAdaptingCache {
@@ -69,7 +70,7 @@ public class RedisCache extends AbstractValueAdaptingCache {
 	 */
 	public RedisCache(String name, byte[] prefix, RedisOperations<? extends Object, ? extends Object> redisOperations,
 			long expiration) {
-		this(name, prefix, redisOperations, expiration, false);
+		this(name, prefix, redisOperations, expiration, false, true);
 	}
 
 	/**
@@ -83,7 +84,23 @@ public class RedisCache extends AbstractValueAdaptingCache {
 	 * @since 1.8
 	 */
 	public RedisCache(String name, byte[] prefix, RedisOperations<? extends Object, ? extends Object> redisOperations,
-			long expiration, boolean allowNullValues) {
+					  long expiration, boolean allowNullValues) {
+		this(name, prefix, redisOperations, expiration, allowNullValues, true);
+	}
+
+	/**
+	 * Constructs a new {@link RedisCache} instance.
+	 *
+	 * @param name cache name
+	 * @param prefix must not be {@literal null} or empty.
+	 * @param redisOperations
+	 * @param expiration
+	 * @param allowNullValues
+	 * @param useLocks
+	 * @since 1.8
+	 */
+	public RedisCache(String name, byte[] prefix, RedisOperations<? extends Object, ? extends Object> redisOperations,
+			long expiration, boolean allowNullValues, boolean useLocks) {
 
 		super(allowNullValues);
 
@@ -92,7 +109,7 @@ public class RedisCache extends AbstractValueAdaptingCache {
 		RedisSerializer<?> serializer = redisOperations.getValueSerializer() != null ? redisOperations.getValueSerializer()
 				: (RedisSerializer<?>) new JdkSerializationRedisSerializer();
 
-		this.cacheMetadata = new RedisCacheMetadata(name, prefix);
+		this.cacheMetadata = new RedisCacheMetadata(name, prefix, useLocks);
 		this.cacheMetadata.setDefaultExpiration(expiration);
 		this.redisOperations = redisOperations;
 		this.cacheValueAccessor = new CacheValueAccessor(serializer);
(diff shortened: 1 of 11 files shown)


kdavisk6:
I'm new here, but if I'm reading the documentation correctly, locking only applies to calls to `<T> T get(final Object key, final Callable<T> valueLoader)`.  Is your intention to use that call without locking?

ljrmorgan:
@kdavisk6 Hi Kevin, there's a more complete description of the issue in the Jira ticket here: https://jira.spring.io/browse/DATAREDIS-582. I probably should have copied that into the PR body, because I think this is a serious issue, and it doesn't seem to have been picked up at all.

Locking at this level is dangerous: there's no guarantee that the call to delete the `~lock` key will succeed, and if it fails then all calls to the cache will hang indefinitely (to make this worse, there is no TTL on the `~lock` key, and `waitForLock()` can't be interrupted, so it really is indefinitely). I _think_ that if you're just calling a single Redis instance then you might be OK because the operations will be wrapped in a transaction, but transactions are not used when talking to a Redis Cluster, which we're using.

The lock is only created in a handful of places (the `get()` method that you referenced, and `clear()` IIRC), but `RedisCache` checks to see if the lock exists before performing most operations. This means you're always making at least twice as many Redis calls as you really need to.

On your question around locking in `get(final Object key, final Callable<T> valueLoader)` - yes, we'd like to call that without locking, or at least, without locking on this level. We have a high cache hit rate, by locking like this we have to make four Redis calls (check if the lock exists, create the lock, get the value from the cache, delete the lock) instead of one for each cache hit, just to try to optimise the case where we get two cache misses for the same key at the same time, which is a pretty rare event for us.

As well as making more calls than needed for each `get()`, by locking like this we serialise all access to a cache across all our application servers, which just doesn't scale. Each `get()` becomes slower because of lock contention (the lock is per cache, not per key within the cache).

The `Cache` interface suggests that this method should use locking "if possible" to minimise calls to `valueLoader`. I think locking per application server would be OK (e.g. using mutexes), but locking across all application servers just impacts performance too much.

We currently mainly use `@Cacheable` to interface with the cache, but we do also make calls directly to the `RedisCache` using the versions of `get()` that don't take a `Callable<T>` (and don't lock). We wanted to move to that method for various reasons, but hit the infinite hang issue (fortunately, before the change made it to production).

Due to the risks and the performance impact, we'd like to be able to opt out of locking entirely. This change allows you to do that, without changing the default behaviour. I've just put this out as a request for comments really, I haven't added any tests for example, but I'd like some advice from the Spring Data Redis team about how best to test this, and whether they'll even consider the change, before spending more time on it.

kdavisk6:
@ljrmorgan Thank you for the explanation.  I noticed the same behavior in my own tests.  I agree, that locking on all `get` calls may not make sense when you are not looking for Write Through semantics.  I do agree, however, that locking is essential when you do require Write Through semantics, as the `<T> T get(final Object key, final Callable<T> valueLoader)` method implies.  

I wonder if the locking mechanism could be refactored from a global singular locking pattern to a read-write locking pattern.  In that scenario, you would still have locking, but instead of waiting for a global lock on all operations, the `get` operation checks for the existence of a `writeLock`.  If one does not exist, the read executes.  If a `writeLock` does exist, the `get` will wait until that `writeLock` clears or expires.  `put`, `evict`, and `clear` operations would create `writeLocks`.  This approach would give cache the safety of allowing write-through semantics when needed and remove the need to create a lock on all `get` calls.  It does not, however, remove the call to Redis to check for the existence of a lock before continuing, but I think the trade off could be worth it.

ssteward54:
I'm running into this problem as well. Once a cache has the ~lock record stuck in there, anything else that accesses the cache just gets stuck in the endless waitForLock loop and hangs the application. I'm seeing this occur using Sentinel, when our cache loading job is running it's causing Redis to failover from master to slave, and since there's no check to ensure the remove lock call succeeded, it leaves it out there and breaks everything else. This is unusable at this point. This PR should definitely be pulled in ASAP.

ljrmorgan:
@christophstrobl @mp911de do you guys have any thoughts on this change? I don't want to pester you, but it would be nice to have an idea of if/when this might be merged, and what changes would be required to get it in. Thanks!

mp911de:
We have to reconsider the implementation of `RedisCache` and `RedisCacheManager`. Especially the locking is too broad and unconditional in some cases. That's something we can only change for Spring Data Redis 2.0 and having a discussion about your requirements is crucial to to address the current issues.

kdavisk6:
I think for our uses at HBR, we have a large cache hit rate and rely on WriteThrough caching to support our autoscaling instances on AWS.  Locking is key for that to work, so I would like the ability to specify when I want that to occur.  The Spring Caching Abstraction `@Cacheable(sync = true)` option looks like the right approach for that, but like the other here in this issue have noticed, the current implementation locks more often than I think is expected.  It feels like the contract put forth by `@Cacheable(sync = true)` is not being adhered too.

My expectation is that locking will only occur when the `sync` flag is set, for the most part.  At the very least, this would be the minimum.  So a short list for us would be:

- Better documentation about how locking is applied, especially in the case of `@Cacheable(sync = true)`
- Locking by element (by key)
- Configurable Lock expiration (if so desired)

These three changes would be a good start I think.  Thoughts?

ljrmorgan:
I think for us we just want to be able to opt-out of locking completely, in the same way that we've opted out of maintaining the `~keys` set. Our monitoring suggests that we're spending more time checking whether keys exist in Redis than we are either setting or getting data from Redis.

I suppose you could also have per-key locking, e.g. the lock for `blog_post:123` could be `{blog_post:123}~lock`, and since those keys would be sharded to the same Redis Cluster node (I think), you could then do everything within a transaction. That should support @kdavisk6's use case, as far as I understand, while avoiding locking the entire cache or potentially locking keys indefinitely.

I think we'd still like to be able to entirely opt-out of locking though

mp911de:
That was fixed with [DATAREDIS-481](https://jira.spring.io/browse/DATAREDIS-481).

PR #232, opened 2017-01-11
DATAREDIS-589 - Fix NPE for non repository related keyspace events.

christophstrobl:
We now no longer rely on `ApplicationEvents` captured in the `RedisKeyValueAdapter` for performing cleanup operations for expired keys, but do this along with the phantom key removal. This removes a flaw when initializing a non repository related `KeyspaceEventListener` publishing events that actually are unrelated to the `Adapter`.

Additionally upgraded test infrastructure to utilize Redis 3.2.6 with disabled protected-mode and enabled keyspace-events.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-REDIS_VERSION:=3.2.0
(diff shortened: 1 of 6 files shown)


mp911de:
That's merged.

PR #240, opened 2017-03-10
DATAREDIS-617 - Remove dependency to org.codehaus.jackson.

christophstrobl:
Remove already deprecated orc.codehaus.jackson dependency and related `RedisSerializer` / `HashMapper` implementations.


mp911de:
That's merged now.

PR #241, opened 2017-03-22
DATAREDIS-618 - RedisServer getNumberOtherSentinels reading wrong pro…

fzilic:
…perty

Changed property key for NUMBER_OTHER_SENTINELS to read "num-other-sentinels" instead of "multi" which is currently used.


mp911de:
Thanks for your PR. That's merged, polished and ported to the according branches now.

PR #244, opened 2017-04-11
DATAREDIS-613 - Stream results instead of returning collections where possible.

christophstrobl:
Allow streaming where possible without breaking Redis API contract. This means we keep the list collection for those operations potentially returning `NULL` to indicate a value is not present but was requested (eg. MGET).


mp911de:
That's merged.

PR #247, opened 2017-04-24
DATAREDIS-626 - Split up RedisConnection and add defaulting to Redis.*Command interfaces.

christophstrobl:
* Split up JedisConnection.
* Split up JedisClusterConnection.
* Split up LettuceConnection.
* Split up LettuceClusterConnection.

Provide method defaulting for current `RedisCommands` via `DefaultedRedisConnection` to provide an easy upgrade path.


mp911de:
That's merged.

PR #248, opened 2017-05-04
DATAREDIS-639 Unify logging framework usage.

mp911de:
* Move off SLF4J to Commons Logging for code that used SLF4J.


mp911de:
Merged.

PR #252, opened 2017-06-20
DATAREDIS-481 - Overhaul RedisCache & RedisCacheManager

christophstrobl:
⚠️ breaking API and behavior changes.

``RedisCache``s now operate directly upon the `RedisConnectionFactory` allowing each `Cache` to use its own `SerializationPair` for converting cache keys and values.

A `RedisCacheManager` with default settings can be obtained via `defaultCacheManager(connectionFactory)`. When unhappy with the defaults `RedisCacheManagerConfigurator` allows to modify those.

```java
RedisCacheManager cm = RedisCacheManager.usingRawConnectionFactory(connectionFactory)
  .withCacheDefaults(defaultCacheConfig())
  .withInitialCacheConfigurations(singletonMap("predefined", defaultCacheConfig().disableCachingNullValues()))
  .transactionAware()
  .createAndGet();
```

The behavior of a single `RedisCache` is defined via its `RedisCacheConfiguration`. Starting from `defaultCacheConfig()` the configuration can be altered as needed.

```java 
RedisCacheConfiguration config = RedisCacheConfiguration.defaultCacheConfig()
  .entryTtl(Duration.ofSeconds(1))
  .disableCachingNullValues();
```

The default `RedisCacheManager` uses a _non locking_ `RedisCacheWriter` for reading & writing bits. While this ensures a max of performance the lack of entry locking can lead to overlapping, non atomic commands, for `putIfAbsent` and `clean` as those methods combine a series of commands sent to Redis. The _locking_ counterpart prevents command overlap by setting an explicit lock key and checking against presence of this key, which leads to additional requests and potential command wait times.


mp911de:
That's merged and polished.

PR #259, opened 2017-07-26
Add COUNT option support for SPOP command. 

christophstrobl:
We now support `SPOP` command with `COUNT` option throughout the imperative, cluster and reactive API.


mp911de:

PR #261, opened 2017-07-27
DATAREDIS-673 - Gracefully return null if key gets deleted after checking for presence

zngj:
I am using spring boot with 1.5.3 + spring-data-redis 1.8.3.RELEASE, and suffer NullPointer frequently,
Per check, it it due to RedisCache.get use exists/get command separately, and the ttl of my key offen between them.
I noticed RedisCache revised in 2.0.
But I need to solve my production issue with 1.8.x, hope it can be merge into 1.8.x so I can use official version again.

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -181,7 +181,8 @@ public class RedisCache extends AbstractValueAdaptingCache {
 			return null;
 		}
 
-		return new RedisCacheElement(cacheKey, fromStoreValue(lookup(cacheKey)));
+		Object storeValue = lookup(cacheKey);
+		return (storeValue != null ? new RedisCacheElement(cacheKey,fromStoreValue(storeValue)) : null);
 	}
 
 	/*


piaolingxue:
👍

piaolingxue:
@zngj some key may exist but value is empty

wreulicke:
@zngj you can rebase on HEAD of 1.8.x? 
In my environment, all tests passed.

src/main/java/org/springframework/data/redis/cache/RedisCache.java
```diff
@@ -181,7 +181,8 @@ public Boolean doInRedis(RedisConnection connection) throws DataAccessException
 			return null;
 		}
 
-		return new RedisCacheElement(cacheKey, fromStoreValue(lookup(cacheKey)));
+		Object storeValue = lookup(cacheKey);
```
mp911de: `lookup(…)` can return `null` values that don't indicate absence but can represent caching of a `null` value. This is because `lookup(…)` applies already deserialization.

We rather should split the actual `GET` from deserialization and inspect the result.

mp911de:
Thanks for your pull request. That's resolved via #310 with a slightly different approach that does not discard cached `null` values.

PR #263, opened 2017-08-05
DATAREDIS-674 - JedisClusterConnection to support zAdd with Set of Tuples.

clemong888:
Zadd more multiple elements in cluster mode was added. 
I have not added any unit tests because i can't seem to find where the test cases of JedisClusterZSetCommands.java are being written or if there are any at all at the moment.

- [◯] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [◯ ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAREDIS).
- [◯ ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [◯ ] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/RedisZSetCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/RedisZSetCommands.java
@@ -15,15 +15,10 @@
  */
 package org.springframework.data.redis.connection;
 
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.Set;
-
-import org.springframework.data.redis.connection.jedis.JedisVersionUtil;
 import org.springframework.data.redis.core.Cursor;
 import org.springframework.data.redis.core.ScanOptions;
 import org.springframework.util.Assert;
+import java.util.Set;
 
 /**
  * ZSet(SortedSet)-specific commands supported by Redis.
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/redis/connection/RedisZSetCommands.java
```diff
@@ -740,4 +745,34 @@ default Long zRemRangeByScore(byte[] key, double min, double max) {
 	 */
 	Set<byte[]> zRangeByLex(byte[] key, Range range, Limit limit);
 
+	/**
+	 * Convert tuples to map of bytes and double.
+	 * Bytes represents the value of the element and double is for the score.
+	 * @param tuples
+	 * @return
+	 */
+	default Map<byte[], Double> zAddArgs(Set<Tuple> tuples) {
```
mp911de: This interface is driver-agnostic and the change would affect users using the Lettuce driver. Rather move the tuple conversion into `JedisConverters`.

clemong888: Thanks for the comment. Yes, it's better to be in JedisConverters. I've made the changes. Please check

mp911de:
Thanks. I'll take the PR from here.

mp911de:
That's merged, polished and back-ported.

clemong888:
Thank you!

PR #265, opened 2017-08-14
DATAREDIS-680 - Incorrect reference on method score

rostykerei:


mp911de:
Thanks for your pull request. That's now merged, polished and back-ported.

PR #266, opened 2017-08-18
DATAREDIS-676 - Pass on configured command timeout to LettuceClusterConnection.

christophstrobl:
We now not only configure the underlying lettuce connection with the command timeout set via the connection factory, but also make sure to pass the option on to the `LettuceClusterConnection`.


mp911de:
That's merged and back ported now.

PR #272, opened 2017-09-05
DATAREDIS-686 - Fix non Java6 method reference in ClusterCommmandExecutor.

christophstrobl:


mp911de:

PR #279, opened 2017-10-02
DATAREDIS-705 - Change return type for set commands from void to Boolean.

christophstrobl:


mp911de:

PR #283, opened 2017-10-06
DATAREDIS-698 - Add support for HSTRLEN. 

christophstrobl:
We now support `HSTRLEN` command throughout `RedisHashCommand` for Lettuce and Jedis in both an imperative and reactive (Lettuce only) manner.

However as Jedis does not natively support `HSTRLEN` via its API we’ve come up with some more reflective invocation allowing to execute commands currently not known by Jedis. We also added this behavior to the cluster implementation which as of now also supports `RedisClusterConnection#execute`.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
@@ -23,10 +23,7 @@ import redis.clients.jedis.JedisCluster;
 import redis.clients.jedis.JedisClusterConnectionHandler;
 import redis.clients.jedis.JedisPool;
 
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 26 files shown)


mp911de:
That's merged now.

src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
```diff
@@ -127,11 +129,51 @@ public JedisClusterConnection(JedisCluster cluster, ClusterCommandExecutor execu
...
 	@Override
 	public Object execute(String command, byte[]... args) {
 
-		// TODO: execute command on all nodes? or throw exception requiring to execute command on a specific node
-		throw new UnsupportedOperationException("Execute is currently not supported in cluster mode.");
+		Assert.notNull(command, "Command must not be null!");
+		Assert.notNull(args, "Args must not be null!");
+
+		return clusterCommandExecutor
+				.executeCommandOnArbitraryNode((JedisClusterCommandCallback<Object>) client -> JedisClientUtils.execute(command,
```
pavel-hp: Hello! @christophstrobl 
I have a question. Why you have chosen `executeCommandOnArbitraryNode`?
What If someone decided to use this API to run another command like:
`connection.execute("FLUSHDB", "ASYNC".getBytes("UTF-8")), true);`
Don't you think that in this case execution behaviour would be different?

PR #294, opened 2017-11-09
DATAREDIS-693 - Add support for UNLINK.

christophstrobl:
We now support `UNLINK` for both Jedis & Lettuce throughout the single node and cluster connection. Reactive support is only available for Lettuce.


mp911de:

PR #311, opened 2018-02-15
DATAREDIS-767 - Add flag to disable in flight Cache creation. 

christophstrobl:
We now provide a flag to disable the in flight cache creation for missing caches. This gives eg. a `CompositeCacheManager` the possibility to chime in.


mp911de:
That's merged, polished, and ported to master.

PR #313, opened 2018-02-15
DATAREDIS-715 - Allow computation of cache key prefix.

christophstrobl:
We now allow, in addition to a static key prefix, the computation of prefixes based on the cache name. This change introduces an alternative to the in 2.x removed `org.springframework.data.redis.cache.RedisCacheKey`, and gives users more control over cache key generation.

```java
// static key prefix
RedisCacheConfiguration.defaultCacheConfig()
    .prefixKeysWith("( ͡° ᴥ ͡°)");

// computed key prefix
RedisCacheConfiguration.defaultCacheConfig()
    .computePrefixWith(cacheName -> "¯\_(ツ)_/¯" + cacheName);
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>2.0.4.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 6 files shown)


mp911de:
That's merged, polished and forward-ported to master now.

PR #317, opened 2018-03-06
DATAREDIS-779 ReactiveValueOperations.set[ifPresent|ifAbsent](…) does not return a value if value was not set

SystemOutPrint:


mp911de:
Thanks for a pull request. We should discuss the actual issue and the solution within the related [ticket](https://jira.spring.io/browse/DATAREDIS-779) first before applying changes. 

This behavior is caused by Lettuce's suppression of `null` values and the change addresses only a single method without considering the other ones.

SystemOutPrint:
I want to use the setIfAbsent method to implement a distributed lock. The return value of setIfAbsent method is a Mono <Boolean> type data, I think the method will returns Mono.just(true) if the method succeeds, Otherwise Mono.just(false). But it's not so, Not quite sure why only emit the onComplete event when the setIfAbsent method failed.

SystemOutPrint:
```java
Retry<Boolean> retry = Retry
                .<Boolean>onlyIf(rc -> rc.exception() instanceof LockHoldedException)
                .randomBackoff(Duration.ofSeconds(1), Duration.ofSeconds(2))
                .retryMax(MAX_NUM_OF_RETRY);

        return template.opsForValue()
                    .setIfAbsent(key, "1")
                    .switchIfEmpty(Mono.error(new LockHoldedException()))
                    .retryWhen(retry);
```

mp911de:
This needs to be fixed in `LettuceReactiveStringCommands.set(…)`. Do you want to apply the fix yourself or should I close this pull request and we do the fix?

SystemOutPrint:
I'll fixed it.

mp911de:
Thanks a lot. I'll take the pull request from here. We will apply the change only to Lovelace since that's a breaking change. Users of Spring Data Redis 2.0 currently don't expect an element to be emitted and after the change, we will emit elements.

mp911de:
That's now merged and polished.

PR #323, opened 2018-03-15
DATAREDIS-788 Check DefaultRedisScript resultType

SystemOutPrint:


SystemOutPrint:
thanks for your reply, I will close the pull request.

PR #324, opened 2018-03-15
DATAREDIS-789 Fixed local is commented out in redis-scripting.adoc

SystemOutPrint:


mp911de:
Thanks a lot. That's merged, polished, and backported now.

PR #326, opened 2018-03-30
DATAREDIS-803 - Work around Redis parameter limitation

cbuehler:
Redis has a [limitation of 1024 * 1024 parameters](https://github.com/antirez/redis/blob/4.0.9/src/networking.c#L1200) for bulk operations.

To insert more than 1024 * 1024 / 2 - 1 entries with putAll(), they need to be split up in multiple HMSET commands.

To reveive more than 1024 * 1024 - 1 entries with entrySet(), we can directly use the HGETALL command instead of first fetching the keys with HKEYS and then fetching the values with HMGET.

### DIFF
--- a/src/main/java/org/springframework/data/redis/support/collections/DefaultRedisMap.java
+++ b/src/main/java/org/springframework/data/redis/support/collections/DefaultRedisMap.java
@@ -38,6 +38,7 @@ import org.springframework.lang.Nullable;
  *
  * @author Costin Leau
  * @author Christoph Strobl
+ * @author Christian Bühler
  */
 public class DefaultRedisMap<K, V> implements RedisMap<K, V> {
 
@@ -156,19 +157,9 @@ public class DefaultRedisMap<K, V> implements RedisMap<K, V> {
 	@Override
 	public Set<java.util.Map.Entry<K, V>> entrySet() {
 
-		Set<K> keySet = keySet();
-		checkResult(keySet);
-		Collection<V> multiGet = hashOps.multiGet(keySet);
-
-		Iterator<K> keys = keySet.iterator();
-		Iterator<V> values = multiGet.iterator();
-
-		Set<Map.Entry<K, V>> entries = new LinkedHashSet<>();
-		while (keys.hasNext()) {
-			entries.add(new DefaultRedisMapEntry(keys.next(), values.next()));
-		}
-
-		return entries;
+		Map<K, V> entries = hashOps.entries();
+		checkResult(entries);
+		return entries.entrySet();
 	}
 
 	/*
--- a/src/test/java/org/springframework/data/redis/support/collections/AbstractRedisMapTests.java
+++ b/src/test/java/org/springframework/data/redis/support/collections/AbstractRedisMapTests.java
@@ -25,6 +25,7 @@ import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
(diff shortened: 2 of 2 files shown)


mp911de:
Please create a ticket in our [issue tracker](https://jira.spring.io/browse/DATAREDIS/) so we can understand, what exactly you want to achieve.

cbuehler:
Ticket created: https://jira.spring.io/browse/DATAREDIS-803

src/main/java/org/springframework/data/redis/core/DefaultHashOperations.java
```diff
@@ -141,16 +143,22 @@ public void putAll(K key, Map<? extends HK, ? extends HV> m) {
 
 		byte[] rawKey = rawKey(key);
 
-		Map<byte[], byte[]> hashes = new LinkedHashMap<>(m.size());
+		int size = Math.min(RedisTemplate.REDIS_MAX_ARGS / 2 - 1, m.size());
```
mp911de: We can't change a single anticipated `HMSET` to multiple ones as we're losing atomicity. That's rather something to be handled in the application code.

src/main/java/org/springframework/data/redis/core/RedisTemplate.java
```diff
@@ -79,12 +79,15 @@
...
  * @author Anqing Shao
  * @author Mark Paluch
+ * @author Christian Bühler
  * @param <K> the Redis key type against which the template works (usually a String)
  * @param <V> the Redis value type against which the template works
  * @see StringRedisTemplate
  */
 public class RedisTemplate<K, V> extends RedisAccessor implements RedisOperations<K, V>, BeanClassLoaderAware {
 
+	public static final int REDIS_MAX_ARGS = 1024 * 1024;
```
mp911de: We should remove this one as there's no need for the constant (see comment above about atomicity).

src/main/java/org/springframework/data/redis/support/collections/DefaultRedisMap.java
```diff
@@ -155,20 +156,7 @@ public boolean containsValue(Object value) {
...
-		Iterator<K> keys = keySet.iterator();
-		Iterator<V> values = multiGet.iterator();
-
-		Set<Map.Entry<K, V>> entries = new LinkedHashSet<>();
-		while (keys.hasNext()) {
-			entries.add(new DefaultRedisMapEntry(keys.next(), values.next()));
-		}
-
-		return entries;
+		return hashOps.entries().entrySet();
```
mp911de: Nice optimization.

src/main/java/org/springframework/data/redis/core/DefaultHashOperations.java
cbuehler: Good point. I didn't think about the atomicity. Removed it again and adapted the test accordingly.

src/main/java/org/springframework/data/redis/core/RedisTemplate.java
cbuehler: Yes, removed.

src/main/java/org/springframework/data/redis/support/collections/DefaultRedisMap.java
cbuehler: Thanks. In the first place, I didn't intend this as an optimization, but as a fix to be able to fetch more than 1024*1024 -1 entries. But I agree, the optimization is a nice side effect ;)
I re-added the `checkResult(...)`.

mp911de:
Thanks a lot. That's merged, polished, and backported to 2.0.x now.

PR #327, opened 2018-04-04
Fix wrong line break

quaff:


mp911de:
Thanks a lot. That's merged and backported now.

PR #335, opened 2018-04-27
DATAREDIS-697 - Add support for BITPOS.

christophstrobl:
We now support the `BITPOS` command throughout the sync and reactive api.


mp911de:

PR #337, opened 2018-04-30
DATAREDIS-716 - Add support for OBJECT REFCOUNT, ENCODING and IDLETIME.

christophstrobl:


mp911de:

PR #338, opened 2018-04-30
DATAREDIS-778 - Align return types of BoundZsetOperations to ZsetOperations.

christophstrobl:
⚠️ **Breaking change** ⚠️


mp911de:
That's merged now.

PR #339, opened 2018-05-03
DATAREDIS-832 - Added a cover page for epub output

Buzzardo:
I added a cover page for the epub output, now that the Spring data reference guides will
have epub output (through a commit against spring-data-build).


mp911de:

PR #344, opened 2018-05-25
DATAREDIS-696 - Add support for RESTORE with REPLACE option.

christophstrobl:
We bypass the lack of driver support by directly executing the command.


mp911de:
That's merged now.

PR #363, opened 2018-10-16
DATAREDIS-881 - Add config parameter shutdownQuietPeriod beside of shutdownTimeout

quaff:
Refer to https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/EventExecutorGroup.java#L52 , shutdownQuietPeriod should be extra option separated from shutdownTimeout.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceClientConfiguration.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceClientConfiguration.java
@@ -104,6 +104,13 @@ public interface LettuceClientConfiguration {
 	Duration getShutdownTimeout();
 
 	/**
+	 * @return the shutdown quiet period used to close the client.
+	 * @since 2.2
+	 * @see io.lettuce.core.AbstractRedisClient#shutdown(long, long, TimeUnit)
+	 */
+	Duration getShutdownQuietPeriod();
+
+	/**
 	 * Creates a new {@link LettuceClientConfigurationBuilder} to build {@link LettuceClientConfiguration} to be used with
 	 * the Lettuce client.
 	 *
@@ -157,6 +164,7 @@ public interface LettuceClientConfiguration {
 		@Nullable ReadFrom readFrom;
 		Duration timeout = Duration.ofSeconds(RedisURI.DEFAULT_TIMEOUT);
 		Duration shutdownTimeout = Duration.ofMillis(100);
+		@Nullable Duration shutdownQuietPeriod;
 
 		LettuceClientConfigurationBuilder() {}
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/data/redis/connection/lettuce/LettuceClientConfiguration.java
```diff
@@ -157,6 +163,7 @@ static LettuceClientConfiguration defaultConfiguration() {
 		@Nullable ReadFrom readFrom;
 		Duration timeout = Duration.ofSeconds(RedisURI.DEFAULT_TIMEOUT);
 		Duration shutdownTimeout = Duration.ofMillis(100);
+		Duration shutdownQuietPeriod = Duration.ofMillis(100);
```
mp911de: Let's pre-initialize this configuration parameter with `null` so we can retain the current behavior. `LettuceClientConfiguration.build()` can then check, whether `shutdownQuietPeriod` was configured. If it was configured, we take the value. If it was left unconfigured, we can apply the `shutdownTimeout` as the quiet period.

The shutdown timeout is right now in use for both parameters. If we change the behavior, then all users that expect a near-zero shutdown are forced to change their code.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceClientConfiguration.java
```diff
@@ -258,12 +265,27 @@ public LettuceClientConfigurationBuilder commandTimeout(Duration timeout) {
...
 			this.shutdownTimeout = shutdownTimeout;
 			return this;
 		}
 
+		/**
+		 * Configure a shutdown timeout.
+		 *
+		 * @param shutdownQuietPeriod must not be {@literal null}.
+		 * @return {@literal this} builder.
+		 * @throws IllegalArgumentException if shutdownQuietPeriod is {@literal null}.
```
mp911de: Nit: `@since 2.2`

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
```diff
@@ -735,6 +736,29 @@ public void setShutdownTimeout(long shutdownTimeout) {
 		getMutableConfiguration().setShutdownTimeout(Duration.ofMillis(shutdownTimeout));
 	}
 
+	/**
+	 * Returns the shutdown quiet period for shutting down the RedisClient (in milliseconds).
+	 *
+	 * @return shutdown timeout.
+	 * @since 2.2
+	 */
+	public long getShutdownQuietPeriod() {
```
mp911de: Let's not introduce property accessors for shutdown quiet period as this pattern is deprecated and we don't require these methods.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
```diff
@@ -1163,6 +1187,7 @@ void resetConnection() {
 		private @Nullable String clientName;
 		private Duration timeout = Duration.ofSeconds(RedisURI.DEFAULT_TIMEOUT);
 		private Duration shutdownTimeout = Duration.ofMillis(100);
+		private Duration shutdownQuietPeriod = Duration.ofMillis(100);
```
mp911de: `MutableLettuceClientConfiguration` is a legacy bridge that provides `LettuceClientConfiguration` for older clients that do not want to upgrade to provide a dedicated `LettuceClientConfiguration`. Let's keep the behavior as-is since we don't want to add features here (i.e. `getShutdownQuietPeriod()` would return `shutdownTimeout `).

mp911de:
Thanks a lot. That's merged and polished now.

PR #373, opened 2018-11-20
DATAREDIS-890 - Randomly send node command to redis cluster.

ChenGuanqun:
Randomly send 'node' command to redis cluster instead of selecting the first one always.

Jira ticket : [ refrence](https://jira.spring.io/projects/DATAREDIS/issues/DATAREDIS-890?filter=allopenissues)


mp911de:
Thanks a lot. We'll take this pull request from here.

PR #376, opened 2018-12-06
DATAREDIS-865 - Expose GenericJackson2JsonRedisSerializer.registerNullValueSerializer() for ObjectMapper customization.

mp911de:
We now expose `registerNullValueSerializer(…)` on `GenericJackson2JsonRedisSerializer` to allow customization of an externally provided `ObjectMapper`. `NullValueSerializer` was previously registered only within a constructor that takes `classPropertyTypeName` and provided no further possibilities to customize `ObjectMapper`.

Code wishing to provide a custom `ObjectMapper` can now call `GenericJackson2JsonRedisSerializer.registerNullValueSerializer(objectMapper, …)` and perform additional customizations before using `ObjectMapper` with `GenericJackson2JsonRedisSerializer`.

---

Related ticket: [DATAREDIS-865](https://jira.spring.io/browse/DATAREDIS-865).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>2.2.0.BUILD-SNAPSHOT</version>
+	<version>2.2.0.DATAREDIS-865-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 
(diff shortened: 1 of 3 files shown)


ekeric13:
hey so `registerNullValueSerializer` is public but it is also static so when i extend GenericJackson2JsonRedisSerializer i cannot overwrite it. do we want to make it a non-static method? is there any other solutions?

i want custom null value serialization than what is provided.

PR #378, opened 2018-12-13
DATAREDIS-909 - Fix constructor creation of entities with collection-typed properties

mp911de:
We now consider the property type in value conversion for constructor arguments to provide the appropriate value type. Previously, we considered the component type which converted a `Collection<T>` to a single `T` that caused downstream `ClassCastException`.

---

Related ticket: [DATAREDIS-909](https://jira.spring.io/browse/DATAREDIS-909).


wdroste:
+1

PR #387, opened 2019-02-26
DATAREDIS-935 - Add single cache configuration option to builder for cache manager.

christophstrobl:


mp911de:
That's merged now.

PR #444, opened 2019-03-26
DATAREDIS-925 - Add and register binary read/write converter for UUID.

christophstrobl:


mp911de:

PR #452, opened 2019-05-21
DATAREDIS-955 - Fix collection initialization when reading nested structures with same name.

christophstrobl:
We now make sure to not falsely populate instances with null values from nested structures.


mp911de:

PR #453, opened 2019-06-03
DATAREDIS-988 - Consider transaction participation when releasing connections.

mp911de:
We now check on connection release whether a connection is in `MULTI` state. If so, then we discard (rollback) the transaction to reset the connection to a fresh state.

We now consider RedisTemplate's `enableTransactionSupport` configuration when releasing connections. Previously, we released bound connections when being in a read-only transaction regardless of the configuration in `RedisTemplate`. This broke session callbacks that expected to remain on the same connection all but the first command were executed on a different connection.

Releasing a connection now no longer unbinds a connection if transaction support is disabled.

---

Related ticket: [DATAREDIS-988](https://jira.spring.io/browse/DATAREDIS-988).
Should be forward-ported.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>2.1.9.BUILD-SNAPSHOT</version>
+	<version>2.1.9.DATAREDIS-988-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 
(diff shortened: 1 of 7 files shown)


quaff:
```
Failed tests: 
  LettuceReactiveRedisConnectionUnitTests.multipleCallsInProgressShouldConnectOnlyOnce:171 
Wanted but not invoked:
connectionProvider.getConnection(
    interface io.lettuce.core.api.StatefulConnection
);
-> at org.springframework.data.redis.connection.lettuce.LettuceReactiveRedisConnectionUnitTests.multipleCallsInProgressShouldConnectOnlyOnce(LettuceReactiveRedisConnectionUnitTests.java:171)
Actually, there were zero interactions with this mock.
```

quaff:
ping

PR #456, opened 2019-06-18
DATAREDIS-963: secondary indexes using parent TTL

antosha4e:
- Secondary indexes using parent TTL


mp911de:
Thanks for your suggestion. Applying a TTL to an index sets the TTL to the last touched object. Since TTL is set on a per-object basis, the last touched object controls the index duration which causes a moment of surprise to users.

While this is an interesting approach that could be useful to some arrangements, we rather favor a index maintenance utility. Closing this pull request without merging it.

PR #461, opened 2019-06-28
DATAREDIS-999 - Use UsingWhen to apply functions to reactive Redis connections

mp911de:
We now use `Flux.usingWhen(…)` to apply functions on reactive Redis connections and to release connections calling non-blocking methods.

---

Related ticket: [DATAREDIS-999](https://jira.spring.io/browse/DATAREDIS-999).

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/ReactiveRedisTemplate.java
+++ b/src/main/java/org/springframework/data/redis/core/ReactiveRedisTemplate.java
@@ -15,9 +15,6 @@
  */
 package org.springframework.data.redis.core;
 
-import org.springframework.data.redis.hash.HashMapper;
-import org.springframework.data.redis.hash.ObjectHashMapper;
-import org.springframework.lang.Nullable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/redis/core/ReactiveRedisTemplate.java
```diff
@@ -190,15 +176,19 @@ public ReactiveRedisConnectionFactory getConnectionFactory() {
...
+			return (exposeConnection ? connToUse : createRedisConnectionProxy(connToUse));
+		}), conn -> {
+			Publisher<T> result = action.doInRedis(conn);
 
-		ReactiveRedisConnection connToExpose = (exposeConnection ? connToUse : createRedisConnectionProxy(connToUse));
-		Publisher<T> result = action.doInRedis(connToExpose);
+			return postProcessResult(result, conn, false);
 
-		return Flux.from(postProcessResult(result, connToUse, false)).doFinally(signal -> conn.close());
+		}, ReactiveRedisConnection::closeLater, ReactiveRedisConnection::closeLater);
```
robotmrv: Is it really nonblocking call?
I see that ReactiveRedisConnection::closeLater returns Mono, but inside LettuceReactiveRedisConnection it uses call marked as blocking
https://github.com/spring-projects/spring-data-redis/blob/2eb7067e8c7e859168a281145cc46ccddb42049f/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveRedisConnection.java#L405-L408 
There are situations when Subscription could be canceled in non blocking thread and this code will block nonblocking thread.

mp911de: Good catch. That's fixed now.

mp911de:
Connections are used either for data or pub/sub. Using both aspects does not happen in Spring Data Redis's support. However, someone could still invoke pub/sub and data operations on the same `LettuceReactiveRedisConnection` so it makes sense to switch to `mergeDelayError(…)`.

PR #463, opened 2019-06-28
DATAREDIS-588 - Use JedisCluster.psetex(…) instead dispatching to a node

mp911de:
We now use `JedisCluster`'s `psetex` command instead of looking up the topology and dispatching the command ourselves.

Simplify `SET` command usage with Jedis using various expiry and presence/absence flags to use consistently `SetParams`.

---

Related ticket: [DATAREDIS-588](https://jira.spring.io/browse/DATAREDIS-588).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>2.2.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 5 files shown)


hebelala:
@mp911de I want to know, why are you doing that "looking up the topology and dispatching..." for `JedisCluster`'s `psetex` before ? Thanks.

mp911de:
@hebelala Prior to Jedis 3, there was no corresponding `psetex` API available on `JedisCluster`.

hebelala:
@mp911de I got it, thank you!

PR #467, opened 2019-07-04
DATAREDIS-1001 - Fix java.util.Date hash mapping when flattening out entries.

christophstrobl:
The `AsArrayTypeDeserializer` causes `Date` values to be rendered as an array like `[java.util.Date, 1561543964015`].  Therefore they are skipped when flattening and if not skipped cannot be read back.

We now register an error handler along with a custom deserializer aware of the used date format to write and read date values without having to store an array.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,13 @@
 		</dependency>
 
 		<dependency>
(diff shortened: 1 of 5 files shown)


mp911de:

PR #472, opened 2019-08-28
DATAREDIS-974 - Apply SSL configuration when creating Jedis Cluster.

christophstrobl:


mp911de:

PR #474, opened 2019-09-02
DATAREDIS-1031 - Fix reactive pExpire/pExpireAt to invoke correct command

mp911de:
We now invoke the correct `PEXPIRE(AT)` command. Previously, we invoked `EXPIRE`/`EXPIREAT` commands resulting in the wrong command being called.

---

Related ticket: [DATAREDIS-1031](https://jira.spring.io/browse/DATAREDIS-1031).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>2.2.0.BUILD-SNAPSHOT</version>
+	<version>2.2.0.DATAREDIS-1031-SNAPSHOT</version>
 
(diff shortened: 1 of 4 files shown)


awso:
I have the same travis-ci/pr issue in my pull request (#471).
Error msg "The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during ."
Please share lights as how to fix this.

mp911de:
TravisCI has removed the ability to install Java 8. We need to enable our builds so they can be built on newer JDK versions. Until then, the TravisCI build is broken.

PR #486, opened 2019-10-26
DATAREDIS-1052 - Improve atomic operation in DefaultRedisCacheWriter.

joongsoo:
issue : https://jira.spring.io/browse/DATAREDIS-1052

Hi. i'm joongsoo.

`lockingCacheWriter` using spin wait like spin lock.
It is send very many request to redis.
And existing code is not atomic. It is check exists lock and do acquire lock, but not check is acquired lock.

If using lua script. it is guarantees to atomic. and it's removing lock/unlock code so possible decrement to sending request count. and reduce complexit in code.

## Changed
- Remove spin lock. (It is send too many request to redis)
- Apply lua script instead of lock operation. (It guarantees to atomic operation)
- Remove tests related to locks

## Check list

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -16,8 +16,6 @@
 package org.springframework.data.redis.cache;
 
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


joongsoo:
Umm.. I will join the commits. close

PR #491, opened 2019-10-30
DATAREDIS-1049 - Fix retrieval of empty keyspaceNotificationsConfigParameter used to keep server defaults.

christophstrobl:
Empty Strings changed to be considered `null` throwing an exception on `getRequiredAttribute` so we now set the value explicitly for this attribute when considered empty.


mp911de:

PR #493, opened 2019-11-09
DATAREDIS-1052 - Improve atomic operation in DefaultRedisCacheWriter.

joongsoo:
issue : https://jira.spring.io/browse/DATAREDIS-1052

Hi. I'm joongsoo.

`lockingCacheWriter` using spin wait for acquire lock. It is send too many request to redis.
And existing wait & acquire code is not atomic. 
1. It is check exists lock and do acquire lock, but not atomic between wait and acquire. 
2. It is not check is aquire lock.

In this case, the atomicity provided by Redis is effective.
If using lua script. it is guarantees to atomic. and it's possible removing lock/unlock code so decrement to sending request count. and reduce complexit in code.

## Added
- Remove spin lock. (It is send too many request to redis)
- Applied lua script instead of lock operation using `evalSha`. `SHA` is pre initialized in class constructor. If it is lost in redis server, try reload.
- Rename `nonLockingRedisCacheWriter` to `newDefaultRedisCacheWriter`. because lock is gone, so the name `lock` is meaningless.

## Removed
- Tests related to locks
- `lockingRedisCacheWriter` in `RedisCacheWriter`, because lock is not needed anymore.

I have already sent a PR on this issue. But sent back for further improvement.
(Before PR : https://github.com/spring-projects/spring-data-redis/pull/489)

## p.s
I removed spin lock. but not remove `name` parameter.
Usually `name` parameter is cache name. so i think if which user want implement `RedisCacheWriter`, they may need a cache name.

## Check list

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -15,18 +15,14 @@
  */
 package org.springframework.data.redis.cache;
 
-import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.Collections;
-import java.util.Optional;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Consumer;
 import java.util.function.Function;
 
-import org.springframework.dao.PessimisticLockingFailureException;
 import org.springframework.data.redis.connection.RedisConnection;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.connection.RedisStringCommands.SetOption;
+import org.springframework.data.redis.connection.ReturnType;
 import org.springframework.data.redis.core.types.Expiration;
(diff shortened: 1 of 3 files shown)


joongsoo:
@mp911de @christophstrobl Hi. I wonder when I can get feedback. Is my explain reasonable?

mp911de:
Thank you for your contribution. We're quite undecided about how to proceed. The current mechanism works for the majority of users and locking can be disabled if someone wishes to do so. Switching to Lua is a valid alternative but there are two issues with that:

1. We haven't seen much demand for a Lua-based implementation
2. Switching to Lua has a huge impact on Redis performance

Therefore, we would not want to include that change. `RedisCacheWriter` comes with a pluggable API where customizations can be made by applications that require such changes.

joongsoo:
@mp911de Thanks for fast reply. 
But the same is true that the lock is not valid. (Does not guarantee synchronized processing)
So if not want using lua script, What about trying  to improve spin lock instead of change to lua script?

PR #494, opened 2019-11-11
DATAREDIS-1066 - Add explanation for Jedis usePooling flag

chang-chao:
Jedis Sentinel and Cluster will always use connection-pooling. so added one line in javadoc for `JedisPoolingClientConfigurationBuilder.usePooling()` method.


mp911de:
Thanks a lot, that's merged, polished, and backported now.

PR #495, opened 2019-11-12
DATAREDIS-1060 - Redis password should not automatically be applied to Sentinel.

christophstrobl:
Instead we now offer a dedicated property for authenticating with Sentinel.

However providing authentication for Sentinel only works using Lettuce until xetorthio/jedis#1943 is resolved.


mp911de:

PR #499, opened 2019-11-15
DATAREDIS-1064 - Modify inappropriate comments

wqlm:
DATAREDIS-1064 - Modify inappropriate comments in /src/main/java/org/springframework/data/redis/core/ListOperations.java


mp911de:
Thanks a lot. That's merged, polished, and backported now.

PR #502, opened 2019-11-19
Fix comment on ZSetOperations.reverseRangeByScore()

wqlm:
DATAREDIS-1070 - Fix incorrect comments.

ZSetOperations.reverseRangeByScore corresponds to redis command zrevrangebyscore instead of zrevrange.


mp911de:

PR #503, opened 2019-12-12
DATAREDIS-1075 - Add EnableKeyspaceNotifications for more configuration options of keyspace event listening.

christophstrobl:


mp911de:
How about meta-annotating `@EnableRedisRepositories` with `@EnableKeyspaceNotifications` and using `@AliasFor` to forward the attributes we have? In the same go, we should deprecate `EnableRedisRepositories.enableKeyspaceEvents` and `EnableRedisRepositories.keyspaceNotificationsConfigParameter` so we can recommend `@EnableKeyspaceNotifications`.

serac:
I'm interested to see this PR merged and I'm willing to implement suggestions @mp911de suggested above and work it to completion. Please let me know how to proceed.

PR #507, opened 2020-01-10
DATAREDIS-1041 - Introduce RedisCache configuration option prefixCacheNameWith.

christophstrobl:
`prefixCacheNameWith` allows to prefix the cache name, still sicking to the default `CacheKeyPrefix.simple()` format.


mp911de:

PR #512, opened 2020-02-21
DATAREDIS-1084 - Add support for XPENDING & XCLAIM. 

christophstrobl:
There are a couple of issues that need to be discussed before we continue with this PR.

* The `JUSTID` flag is currently not suppoprted by drivers.
* `Mono<PendingMessages> xPending(ByteBuffer key, Consumer consumer)` vs. `Flux<PendingMessage> xPending(ByteBuffer key, Consumer consumer)`.
* Coupling between `RedisStreamCommands` and `ReactiveStreamCommands` via `RedisStreamCommands.XClaimOptions`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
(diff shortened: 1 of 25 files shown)


mp911de:

PR #515, opened 2020-03-04
DATAREDIS-1110 - fix type parameter on StreamMessageListenerContainer…

Bakkenrak:
…OptionsBuilder.hashValueSerializer()


christophstrobl:
Thank you very much @Bakkenrak! Merged to _master_ and back ported to _2.2.x_

PR #517, opened 2020-03-19
DATAREDIS-1091 - Add configuration to disable storage of shadow copies for entities using TimeToLive.

christophstrobl:


mp911de:
That's merged now.

PR #519, opened 2020-03-27
DATAREDIS-1119 - Add support XINFO.

christophstrobl:


mp911de:

PR #526, opened 2020-04-26
support microseconds

hsoftxl:
support get microseconds

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
+++ b/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
@@ -309,9 +309,9 @@ abstract public class Converters {
 	 * @param microseconds elapsed microseconds in current second
 	 * @return
 	 */
-	public static Long toTimeMillis(String seconds, String microseconds) {
-		return NumberUtils.parseNumber(seconds, Long.class) * 1000L
-				+ NumberUtils.parseNumber(microseconds, Long.class) / 1000L;
+	public static Long toTimeMicros(String seconds, String microseconds) {
+		return NumberUtils.parseNumber(seconds, Long.class) * 1000 * 1000L
+				+ NumberUtils.parseNumber(microseconds, Long.class);
 	}
 
 	/**
--- a/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
+++ b/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
@@ -315,6 +315,17 @@ abstract public class Converters {
 	}
 
 	/**
+	 * Returns the timestamp constructed from the given {@code seconds} and {@code microseconds}.
+	 *
+	 * @param seconds server time in seconds
+	 * @param microseconds elapsed microseconds in current second
+	 * @return
+	 */
+	public static Long toTimeMillis(String seconds, String microseconds) {
+		return NumberUtils.parseNumber(seconds, Long.class) * 1000L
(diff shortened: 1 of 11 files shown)


src/main/java/org/springframework/data/redis/connection/convert/Converters.java
```diff
@@ -309,9 +309,9 @@ protected static RedisClusterNode toClusterNode(String clusterNodesLine) {
 	 * @param microseconds elapsed microseconds in current second
 	 * @return
 	 */
-	public static Long toTimeMillis(String seconds, String microseconds) {
-		return NumberUtils.parseNumber(seconds, Long.class) * 1000L
-				+ NumberUtils.parseNumber(microseconds, Long.class) / 1000L;
+	public static Long toTimeMicros(String seconds, String microseconds) {
```
mp911de, line 312: Let's keep the old method and introduce a new one to not break backward compatibility.

Also, we should use `TimeUnit.MILLIS`/`MICROS` for conversion for more expressive code. It makes sense to update `toTimeMillis` as well.

src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterServerCommands.java
```diff
@@ -518,11 +529,15 @@ public void migrate(byte[] key, RedisNode target, int dbIndex, @Nullable Migrate
 
 	private Long convertListOfStringToTime(List<String> serverTimeInformation) {
 
+		return convertListOfStringToMicros(serverTimeInformation) / 1000;
```
mp911de, line 532: Let's use `TimeUnit` constants here as well.

src/main/java/org/springframework/data/redis/connection/jedis/JedisServerCommands.java
```diff
@@ -353,7 +353,16 @@ public void resetConfigStats() {
 	 */
 	@Override
 	public Long time() {
+		Long time = microseconds();
+		return  time == null ? null : time / 1000;
```
mp911de, line 357: Let's use `TimeUnit` constants for conversion.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceClusterServerCommands.java
```diff
@@ -312,7 +312,16 @@ public void resetConfigStats(RedisClusterNode node) {
 	 */
 	@Override
 	public Long time() {
+		Long time = microseconds();
+		return  time == null ? null : time / 1000;
```
mp911de, line 316: Let's use `TimeUnit` constants for conversion.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceServerCommands.java
```diff
@@ -361,7 +361,16 @@ public void resetConfigStats() {
 	 */
 	@Override
 	public Long time() {
+		Long time = microseconds();
+		return  time == null ? null : time / 1000;
```
mp911de, line 365: Let's use TimeUnit constants for conversion.

mp911de (changes requested):
First, sorry for the long radio silence. I reviewed this PR and left a few comments to be addressed.

mp911de:
Thanks for the pull request. During the merge, we found that the implementation could be streamlined towards overloaded `time` methods accepting `TimeUnit`. Therefore we went for a slightly different approach.

PR #530, opened 2020-05-11
DATAREDIS-1143 - Delombok source files.

christophstrobl:


mp911de:

quaff:
@mp911de Would you explain why?

mp911de:
Lombok introduces some side-effects when starting out to use it, such as the class files get changed with the annotation processor while the source code itself (which goes into `-sources.jar`) remains unchanged. Similar goes for Javadoc: While lombok generates methods and such, those don't show up in the Javadoc. One needs to heavily customize a project setup to bring in back those things. So Lombok caused more issues in production code than good.

quaff:
Does those issues can be fixed in lombok or it's impossible by annotation processor?

quaff:
Class files get changed even source files remains changed is trivial since we are not add those generated files to VCS.
does this occurs even lombok self not upgraded?

PR #531, opened 2020-05-18
DATAREDIS-1225 - Fix copy-paste error

quaff:
RedisStandaloneConfiguration should be RedisSentinelConfiguration


mp911de:

PR #532, opened 2020-05-21
DATAREDIS-1151 - change clean method  `keys` command to `scan`

eyison:
`clean` method use `keys` command to get all keys which is blocked , may cause other commands time out.

https://jira.spring.io/browse/DATAREDIS-1151

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -17,18 +17,20 @@ package org.springframework.data.redis.cache;
 
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.Collections;
-import java.util.Optional;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
+import com.sun.istack.internal.Nullable;
 import org.springframework.dao.PessimisticLockingFailureException;
 import org.springframework.data.redis.connection.RedisConnection;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.connection.RedisStringCommands.SetOption;
+import org.springframework.data.redis.core.Cursor;
+import org.springframework.data.redis.core.ScanOptions;
 import org.springframework.data.redis.core.types.Expiration;
-import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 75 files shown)


pivotal-issuemaster:
@eyison Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-redis&pullRequestId=532)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-redis&pullRequestId=532) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@eyison Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-redis&pullRequestId=532)!

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -179,11 +179,14 @@ public void clean(String name, byte[] pattern) {
...
-				if (keys.length > 0) {
-					connection.del(keys);
+				List<byte[]> keys = new ArrayList<byte[]>();
+				Cursor<byte[]> cursor = connection.scan(new ScanOptions.ScanOptionsBuilder().match(new String(pattern)).build());
+				while (cursor.hasNext()) {
+					byte[] bytes = cursor.next();
+					keys.add(bytes);
+				}
+				if (!keys.isEmpty()) {
+					connection.del(keys.toArray(new byte[][]{}));
```
chengchen, line 189: As the complexity of DEL is O(N), if the number of keys is huge, DEL command could time out as well, right?
Maybe we could do it batch by batch as well?

I was thinking of using [UNLINK](https://redis.io/commands/unlink), but it's only supported in higher Redis version and it's a change of behavior (from sync to async)...

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -179,11 +179,14 @@ public void clean(String name, byte[] pattern) {
 					wasLocked = true;
 				}
 
-				byte[][] keys = Optional.ofNullable(connection.keys(pattern)).orElse(Collections.emptySet())
-						.toArray(new byte[0][]);
-
-				if (keys.length > 0) {
-					connection.del(keys);
+				List<byte[]> keys = new ArrayList<byte[]>();
```
cnhongwei: Use LinkedList replace ArrayList because the list may be huge.

mp911de:
The change performs a `DEL` call per cache key which isn't efficient. We have received a newer PR with #547 that supersedes this one.

PR #540, opened 2020-06-24
DATAREDIS-1173 - Close connection unsubscribes from all channels incorrectly.

sabbeyPivotal:
The LettuceSubscription doClose method calls doUnsubscribe with an empty byte array, which results in a call to Redis of:

`UNSUBSCRIBE ""`
This command will unsubscribe from a channel with the name ""  However, we believe the intended behavior is to unsubscribe from all channels, which can be done by not sending any argument to doUnsubscribe, resulting in a call to Redis of:

`UNSUBSCRIBE`
We wrote a failing and passing test showing the problem.  We could not find a way to expose the issue by calling doClose on the subscription (because doClose also releases the connection), so we had to call unsubscribe directly.  We are not sure where this belongs in your testing pyramid.

Co-authored-by: Sarah Abbey <sabbey@pivotal.io>
Co-authored-by: Murtuza Boxwala <mboxwala@pivotal.io>
Co-authored-by: Jens Deppe <jdeppe@vmware.com>

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceSubscription.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceSubscription.java
@@ -27,6 +27,9 @@ import org.springframework.data.redis.connection.util.AbstractSubscription;
  * @author Costin Leau
  * @author Mark Paluch
  * @author Christoph Strobl
+ * @author Sarah Abbey
+ * @author Murtuza Boxwala
+ * @author Jens Deppe
  */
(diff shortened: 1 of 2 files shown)


mp911de:
Sorry, for some reason this pull request went unnoticed. We created #549 as the ticket didn't had a pull request URL with it associated. Both PR's have similar changes however #549 has a simpler test setup.

christophstrobl:
Thanks for the fix. Back ported to 2.1.x, 2.2.x and 2.3.x

PR #546, opened 2020-07-07
DATAREDIS-1175 - Allow direct mapping of Collection like structures via MappingRedisConverter.

christophstrobl:
It is now possible to hand a plain `List` to the converter that is then converted into a hash structure using index numbers as keys.


mp911de:

akuma8:
Hi @mp911de, 
How can I test this? 
Thanks

PR #547, opened 2020-07-10
DATAREDIS-1151 - Provide an example of using scan as an option in RedisCache

chengchen:
I commented out the original implementation and added an example of using `SCAN`.
After a second thought, it might be reasonable to keep both and allow users to toggle the choice?

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.data.redis.cache;
 
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.Collections;
@@ -27,9 +28,12 @@ import org.springframework.dao.PessimisticLockingFailureException;
 import org.springframework.data.redis.connection.RedisConnection;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.connection.RedisStringCommands.SetOption;
+import org.springframework.data.redis.core.Cursor;
+import org.springframework.data.redis.core.ScanOptions;
 import org.springframework.data.redis.core.types.Expiration;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import reactor.core.publisher.Flux;
 
 /**
  * {@link RedisCacheWriter} implementation capable of reading/writing binary data from/to Redis in {@literal standalone}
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -179,12 +183,31 @@ public void clean(String name, byte[] pattern) {
...
 				}
 
-				byte[][] keys = Optional.ofNullable(connection.keys(pattern)).orElse(Collections.emptySet())
-						.toArray(new byte[0][]);
+//				byte[][] keys = Optional.ofNullable(connection.keys(pattern)).orElse(Collections.emptySet())
+//					.toArray(new byte[0][]);
+//
+//				if (keys.length > 0) {
+//					connection.del(keys);
+//				}
```
chengchen, line 191: If this is a behavior change for you, maybe we could keep both and use a feature flag to toggle the choice?

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -179,12 +183,31 @@ public void clean(String name, byte[] pattern) {
...
 
-				if (keys.length > 0) {
-					connection.del(keys);
+				int batchSize = 1000; // maybe make it configurable
```
chengchen, line 193: If the user choose to use `SCAN` approach, we could let them configure this batch size as well, so that the Redis `timeout` option and `batch delete size` option are both in the hands of users.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -179,12 +183,31 @@ public void clean(String name, byte[] pattern) {
...
+				ScanOptions options = ScanOptions.scanOptions().match(new String(pattern)).build();
```
rickgong, line 194: batchSize here seems not applied to ScanOptions

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -179,12 +183,31 @@ public void clean(String name, byte[] pattern) {
...
+
+				try (Cursor<byte[]> cursor = connection.scan(options)) {
+					while(cursor.hasNext()) {
+						byte[][] keys = new byte[batchSize][];
+
+						for(int i = 0; i < batchSize && cursor.hasNext(); i++) {
+							keys[i] = cursor.next();
+						}
+
+						connection.del(keys);
```
rickgong, line 204: keys array may contains null element, which may fail DEL or UNLINK command

christophstrobl:
Usage of `SCAN` command is now possible via  a`BatchStrategy`. Resolved via #2051.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
bert82503: nice work
`int batchSize = 1000; // maybe make it configurable`

PR #548, opened 2020-07-14
DATAREDIS-1179 - Allow reuse of RedisConverter in ObjectHashMapper.

christophstrobl:


mp911de:

PR #549, opened 2020-07-15
DATAREDIS-1173 - Correctly unsubscribe from patterns/channels through LettuceSubscription

mp911de:
`doPUnsubscribe(…)` and `doUnsubscribe(…)` now consider the all flag to unsubscribe from all subscribed patterns/channels. Previously, both methods didn't consider all and were invoked with an empty byte array which unsubscribed from an empty pattern/channel name and left subscriptions active.

---

Related ticket: DATAREDIS-1173
Should be forward/backported to all maintained versions.


mp911de:
We aren't able to test each feature through integration tests as the already huge number of tests would massively grow. I think an integration test should be possibly by using `CLIENT LIST` as client details enumerate to how many channels a client is subscribed.

PR #551, opened 2020-07-27
fix rename method when targetKey exists

majianming21:


src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterKeyCommands.java
```diff
@@ -269,7 +269,7 @@ public void rename(byte[] sourceKey, byte[] targetKey) {
 		byte[] value = dump(sourceKey);
 
 		if (value != null && value.length > 0) {
-
+			del(targetKey);
 			restore(targetKey, 0, value);
```
mp911de, line 273: We should rather use the `restore` command with the `REPLACE` option.

mp911de:
Closing due to lack of requested changes

PR #556, opened 2020-08-22
DATAREDIS-1207 - Fix LettuceStreamCommands.xRevRange Range conversion.

jankovd:
LettuceStreamCommands.xRevRange converts Ranges with a StringCodec that produces Boundary<ByteBuffer>, while Lettuce expects the range's Boundary to include strings. The conversion worked well only in cases where the Range was unbounded (- or +), and failed with ClassCastException on the consuming site for other cases.


christophstrobl:
Thank you @jankovd! Merged to _main_ line and back ported to _2.3.x_ and _2.2.x_.

PR #557, opened 2020-08-28
DATAREDIS-1210 - fix StreamReadOptions toString method

zkz2004:
original StreamReadOptions toString method build String with another StreamReadOptions  (formed by autoAcknowledge() ) inside, which cause StringBuilder append be called infinitely, and cause StackOverflowError.
fix it by change toString return string to

"StreamReadOptions{" + "block=" + block + ", count=" + count + ", noack=" + noack  + ", blocking=" + isBlocking() + '}'


christophstrobl:
Thank you @zkz2004! Merged to _main_ line.

PR #562, opened 2020-09-14
DATAREDIS-1103 - Support new KEEPTTL option of the SET operation

dengliming:


mp911de:
Thanks for your pull request. Instead of adding a new method with a different name, we should introduce overloads accepting `keepTtl` as `boolean` option. `KEEPTTL` can be combined with `XX` and `NX` but not `PX`/`EX` options.

Care to update your pull request?

dengliming:
@mp911de Hi~ Do you mean like this?
`Boolean set(byte[] key, byte[] value, Expiration expiration, SetOption option, boolean keepttl);`

mp911de:
Yes, exactly.

dengliming:
@mp911de Got it. Thanks. That was updated.

christophstrobl:
Thank you @dengliming. We changed the original PR a bit and merged it to the main line.
To ensure existing behaviour is not broken or changed, please make sure to run all tests (with the `-DrunLongTests=true` flag) - thanks!

PR #563, opened 2020-09-15
DATAREDIS-1196 - Add support for LPOS.

christophstrobl:
We now support `LPOS` via the imperative and reactive list commands. 
`(Reactive)ListOperations` support a minimal set of `LPOS` variants (`indexOf` & `lastIndexOf`) to back the corresponding `DefaultRedisList` methods.

Some of the tests (those using the `RANK` option) are still ignored because they require lettuce-io/lettuce-core#1410.


mp911de:

PR #565, opened 2020-10-01
DATAREDIS-1214 - Fix null return value of GEODIST command.

christophstrobl:


mp911de:

PR #574, opened 2020-11-22
DATAREDIS-1253 - Fixed a bug that the key is temporarily deleted.

taka2:
The put method of the RedisKeyValueAdapter class temporarily removes the key (del & hMSet), so the existing hash could sometimes disappear.
To avoid this issue, I changed the hash existence check from connection.del to connection.exists.


mp911de:
Closing this ticket as per comment in the related ticket since this change introduces inconsistency bugs.

PR #1961, opened 2021-02-08
DATAREDIS-1955 - Phantom copy is not persisted when entity ttl is changed from positive to negative, causing phantom copy not updated to no expire and counting down, then RedisKeyExpiredEvent publishing

muchnik:
DATAREDIS-1955 - Phantom copy is not persisted when entity ttl is changed from positive to negative, causing phantom copy not updated to no expire and counting down, then RedisKeyExpiredEvent publishing.

When updating an object that has a positive time to live to a negative time to live you must persist the phantom key.
Otherwise, after expiring time to live of the phantom, which was created when object with a positive time to live was saved - the indices are dropped and the keys are deleted from the sets as if entity is expired.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>2.5.0-SNAPSHOT</version>
+	<version>2.5.0-DATAREDIS-1955-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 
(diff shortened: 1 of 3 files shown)


christophstrobl:
@muchnik thanks for the PR! 
We can take it from here - I tend to switch from persisting the _phantom key_ to removing it (also for the update scenario). This would reduce the footprint on server side.  Objections?

muchnik:
> @muchnik thanks for the PR!
> We can take it from here - I tend to switch from persisting the _phantom key_ to removing it (also for the update scenario). This would reduce the footprint on server side. Objections?

Nice idea, updated PR!

christophstrobl:
Thanks @muchnik! Merged to _main_ branch and back ported to _2.3.x_ and _2.4.x_.

PR #1962, opened 2021-02-10
Support LPOS command via Jedis. 

christophstrobl:
This PR introduces support for the `LPOS` command when using the Jedis Redis driver and ensures an empty `List` is returned in the case the member does not exist in the target key.

Closes: #1957


mp911de:

Issue #1711, opened 2020-05-06
Add support to Redis Streams using Jedis Client [DATAREDIS-1140]


spring-projects-issues:
**[Deep Shiv](https://jira.spring.io/secure/ViewProfile.jspa?name=deepshiv126)** opened **[DATAREDIS-1140](https://jira.spring.io/browse/DATAREDIS-1140?redirect=false)** and commented

Spring Data Redis supports Redis Streams using Lettuce Driver but not with Jedis Driver.

At this moment Jedis Support - Redis Streams.
https://github.com/xetorthio/jedis/releases/tag/jedis-3.1.0

Also please update the ticket with a tentative time.

 

 

```java
[...]

PR #1977, opened 2021-02-18
DATAREDIS-1140 - Add support to Redis Streams using Jedis Client

dengliming:
resolves [https://github.com/spring-projects/spring-data-redis/issues/1711](https://github.com/spring-projects/spring-data-redis/issues/1711)

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
@@ -15,6 +15,14 @@
  */
 package org.springframework.data.redis.connection.jedis;
 
+import static org.springframework.data.redis.connection.jedis.StreamConverters.convertToByteRecord;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
 import org.springframework.data.domain.Range;
 import org.springframework.data.redis.connection.RedisStreamCommands;
 import org.springframework.data.redis.connection.RedisZSetCommands;
@@ -29,18 +37,12 @@ import org.springframework.data.redis.connection.stream.StreamInfo;
 import org.springframework.data.redis.connection.stream.StreamOffset;
 import org.springframework.data.redis.connection.stream.StreamReadOptions;
 import org.springframework.util.Assert;
+
 import redis.clients.jedis.BinaryJedis;
+import redis.clients.jedis.BuilderFactory;
 import redis.clients.jedis.MultiKeyPipelineBase;
 import redis.clients.jedis.StreamConsumersInfo;
 import redis.clients.jedis.StreamGroupInfo;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
-
-import static org.springframework.data.redis.connection.jedis.StreamConverters.convertToByteRecord;
 
 /**
  * @author Dengliming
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
```diff
@@ -0,0 +1,496 @@
...
+		Assert.notNull(record, "Record must not be null!");
+		Assert.notNull(record.getStream(), "Stream must not be null!");
+
+		byte[] id = JedisConverters.toBytes(record.getId().getValue());
+		Long maxLength = Long.MAX_VALUE;
+		if (options.hasMaxlen()) {
+			maxLength = options.getMaxlen();
+		}
+
+		return connection.invoke().from(BinaryJedis::xadd, MultiKeyPipelineBase::xadd, record.getStream(), id,
```
mp911de, line 85: Good work, we introduced the invoker just recently.

src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
```diff
@@ -0,0 +1,496 @@
...
+		if (options.getRetryCount() != null) {
+			retryCount = options.getRetryCount().intValue();
+		}
+		long unixTime = -1L;
+		if (options.getUnixTime() != null) {
+			unixTime = options.getUnixTime().toEpochMilli();
+		}
+		try {
+			if (isPipelined()) {
+				pipeline(connection.newJedisResult(connection.getRequiredPipeline().xclaim(key, JedisConverters.toBytes(group),
```
mp911de, line 122: Any reason to not use `JedisInvoker` here?

src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
```diff
@@ -0,0 +1,496 @@
...
+	 * (non-Javadoc)
+	 * @see org.springframework.data.redis.connection.RedisStreamCommands#xInfoConsumers(byte[], java.lang.String)
+	 */
+	@Override
+	public StreamInfo.XInfoConsumers xInfoConsumers(byte[] key, String groupName) {
+		Assert.notNull(key, "Key must not be null!");
+		Assert.hasText(groupName, "Group name must not be null or empty!");
+
+		return connection.invoke().from(BinaryJedis::xinfoConsumers, (k, v, r) -> {
+			throw new UnsupportedOperationException("'XINFO CONSUMERS' cannot be called in pipeline / transaction mode.");
```
mp911de, line 244: We typically add that guard beforehand, see https://github.com/spring-projects/spring-data-redis/blob/53ce84851cbb0350f5df1daf40d5623d5c7e756f/src/main/java/org/springframework/data/redis/connection/jedis/JedisServerCommands.java#L210-L212

src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
```diff
@@ -0,0 +1,496 @@
...
+	 */
+	@Override
+	public PendingMessages xPending(byte[] key, String groupName, XPendingOptions options) {
+		Assert.notNull(key, "Key must not be null!");
+		Assert.notNull(groupName, "GroupName must not be null!");
+
+		Range<String> range = (Range<String>) options.getRange();
+		byte[] group = LettuceConverters.toBytes(groupName);
+		try {
+			if (isPipelined()) {
```
mp911de, line 285: Let's use `JedisInvoker` here.

src/main/java/org/springframework/data/redis/connection/jedis/StreamConverters.java
```diff
@@ -0,0 +1,126 @@
...
+ * Converters typically convert between value objects/argument objects retaining the actual types of values (i.e. no
+ * serialization/deserialization happens here).
+ *
+ * @author dengliming
+ * @since 2.3
+ */
+@SuppressWarnings({ "rawtypes" })
+class StreamConverters {
+
+	private static final BiFunction<Object, String, org.springframework.data.redis.connection.stream.PendingMessages> PENDING_MESSAGES_CONVERTER = (
```
mp911de, line 48: We refactored a few converters recently to do the conversion directly in the actual conversion method (e.g. `toPendingMessages`) as the indirection over `BiFunction` or `Converter` always requires navigating to the actual implementation. Having the conversion code in the actual method makes the code easier to read.

mp911de (commented):
Thank you for your contribution, great work. I left a few comments, mostly high-level aspects. Care to have a look?

src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
dengliming: This is a real headache because the parameter types of the `BinaryJedis::xclaim` `MultiKeyPipelineBase::xclaim` two methods are not the same.

src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
dengliming: Got it. Thanks.

src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
dengliming: Do you have any good suggestions? I'm doing this because the return types of the BinaryJedis::xpending MultiKeyPipelineBase::xpending two methods are not the same.

src/main/java/org/springframework/data/redis/connection/jedis/StreamConverters.java
dengliming: Thanks. see the lastest commit.

src/main/java/org/springframework/data/redis/connection/jedis/JedisStreamCommands.java
mp911de: Yikes. Instead, you could unfold the method reference into a lambda and adopt the parameters there. Also, it would make sense to file a ticket so Jedis folks can fix the method signature.

dengliming: Thanks. fixed at lastest commit. will file  a ticket on Jedis later.

mp911de:

Issue #1794, opened 2020-10-21
ifAbsent support with ZsetOperations [DATAREDIS-1235]


spring-projects-issues:
**[Kuldeep Singh](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER50584)** opened **[DATAREDIS-1235](https://jira.spring.io/browse/DATAREDIS-1235?redirect=false)** and commented

**Expected Behavior**
ZsetOperations Should support ifAbsent method
**Current Behavior**

ZsetOperations does not support ifAbsent method currently

**Context**

I am trying to use Redis SortedSets for a project where I Need to insert a set only if its already not present. But with current implementation of RedisTemplate I am unable to achieve that as ifAbsent is not supported for ZsetOperations yet

 

Also please suggest if there is any alternative way or workaround to achieve the same thing.
PS: I want to achieve the functionality of ifAbsent only in per operation basis. I do not want ifAbsent to be enabled for all operations

---
No further details from [DATAREDIS-1235](https://jira.spring.io/browse/DATAREDIS-1235?redirect=false)

PR #1988, opened 2021-03-01
Support ZADD Flags NX, XX, GT & LT

christophstrobl:
Closes: #1794


mp911de:

PR #2012, opened 2021-03-22
Add support for CONFIG REWRITE

### LINKED ISSUE #1992: Add support for CONFIG REWRITE
Hi, I called the setConfig method in the RedisConnection class to modify some configurations, but I did not find the config rewrite command. It seems that this command is not supported. Will this command be supported in the future? How do I write the modified config into the configuration file now?
Thanks.

manousos:
Closes #1992


mp911de:

PR #2056, opened 2021-04-27
Fix for database index select when creating client.

s-sathish:
Closes #2053
Related tickets redis/jedis#2523

Issue #2057, opened 2021-04-27
Guard non-initialized `RedisConnectionFactory.getConnection()` with `IllegalStateException`


mp911de:
From #2053 we learned that it is pretty simple to run into a scenario where a connection factory remains unconfigured while obtaining connections works to a certain degree. We should fail early with an `IllegalStateException` if the connection factory is not fully initialized via `afterPropertiesSet` to avoid using partially configured clients.

PR #2058, opened 2021-04-27
Guard `RedisConnectionFactory.getConnection()` against usage without initialization

mp911de:
Implementations of `getConnection()` and their cluster, sentinel and reactive variants on `JedisConnectionFactory` and `LettuceConnectionFactory` now throw `IllegalStateException` if the connection factory was not initialized with `afterPropertiesSet()`. This change prevents the usage of partially configured clients.

Closes #2056 
Should be backported to 2.5.x.


mp911de:
Superseded by #2057.

PR #2077, opened 2021-06-05
Remove unused comment characters

weixsun:
This pr only remove unused comment characters in `JedisConnectionFactory`


mp911de:
There's actually nothing broken here. We intentionally have trailing comment markers on long lines to specify line breaks so the auto-formatter doesn't re-join these lines.

Issue #2079, opened 2021-06-10
RedisCache synchronises all get(key, valueLoader) calls


miensol:
I was examining the performance in our Spring Boot app. Based on profiling data, I noticed that we have a lot of wait times on `RedisCache#get(key, valueLoader)` method. 

The [method signature](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/cache/RedisCache.java#L121) makes it obvious why this is happening:

`public synchronized <T> T get(Object key, Callable<T> valueLoader) {`

In case of our app, it causes **all** cache lookups to be executed sequentially. (Admittedly a great protection for Redis server 😅).

In the past this method wasn't synchronized. It changed in [this commit](https://github.com/spring-projects/spring-data-redis/commit/a9c2b1cdda6a9d185854c404d6b3889837be56ea).

Documentation of the [Cache interface states](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/cache/Cache.html#get-java.lang.Object-java.util.concurrent.Callable-):

> Return the value to which this cache maps the specified key, obtaining that value from valueLoader if necessary. This method provides a simple substitute for the conventional "if cached, return; otherwise create, cache and return" pattern.
>
> If possible, implementations should ensure that the loading operation is synchronized so that the specified valueLoader is only called once in case of concurrent access on the same key.

Technically speaking, the current synchronized implementation adheres to documentation. However, so would version that is not synchronized. 

For comparison, here's how the [`EhCacheCache`](https://github.com/spring-projects/spring-framework/blob/20d9a1e84430830edca6cc58680ba74e40a96ca2/spring-context-support/src/main/java/org/springframework/cache/ehcache/EhCacheCache.java#L92) is implemented:

```
    public <T> T get(Object key, Callable<T> valueLoader) {
		Element element = lookup(key);
		if (element != null) {
			return (T) element.getObjectValue();
		}
		else {
			this.cache.acquireWriteLockOnKey(key);
			try {
				element = lookup(key);  // one more attempt with the write lock
				if (element != null) {
					return (T) element.getObjectValue();
				}
				else {
					return loadValue(key, valueLoader);
				}
			}
			finally {
				this.cache.releaseWriteLockOnKey(key);
			}
		}
	}
```

I'm happy to submit a PR that changes the implementation. However, before doing so, I'd love to ask if the synchronized method was implemented in the current way deliberately and there's and edge case which makes changing it harder?
[...]

PR #2082, opened 2021-06-11
Synchronize all get value loaders only if value is absent for given key

miensol:
See #2079

After this PR, calls to `get(key, valueLoader)` will be synchronized only in case an initial lookup for value fails. 

The synchronization ignores cache key. That is, if 2 threads call the method e.g. `get(1, ()-> loadValueFor(1))` and `get(2, () -> loadValueFor(2))`, _at the same time_, the `loadValueFor` calls are going to be executed sequentially. 


mp911de:

PR #2083, opened 2021-06-11
Synchronize get value loaders taking cache key into account

### LINKED ISSUE #2079: RedisCache synchronises all get(key, valueLoader) calls

miensol:
The previous version synchronize all calls to `get(key, valueLoader)`. After this PR, the calls to value loader will only be synchronized if we do not have a value for the given key.

This is an extension of #2082.  

This version synchronizes loaders taking `key` into account. 

Fixes #2079

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -34,6 +34,7 @@ import org.springframework.data.redis.serializer.RedisSerializer;
 import org.springframework.data.redis.util.ByteUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.ConcurrentReferenceHashMap;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.ReflectionUtils;
 
@@ -44,6 +45,7 @@ import org.springframework.util.ReflectionUtils;
  *
  * @author Christoph Strobl
  * @author Mark Paluch
+ * @author Piotr Mionskowski
  * @see RedisCacheConfiguration
  * @see RedisCacheWriter
  * @since 2.0
@@ -118,7 +120,7 @@ public class RedisCache extends AbstractValueAdaptingCache {
 	 */
 	@Override
 	@SuppressWarnings("unchecked")
-	public synchronized <T> T get(Object key, Callable<T> valueLoader) {
+	public <T> T get(Object key, Callable<T> valueLoader) {
 
 		ValueWrapper result = get(key);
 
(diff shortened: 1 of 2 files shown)


christophstrobl:
I share the concerns, about keeping a mutex within `RedisCache`, @mp911de outlined in #2079, . 
At this point this is something that we don’t want to add.
Thanks for the time invested, both, on discussing the issue, as well as creating this PR.

PR #2087, opened 2021-06-15
Fix result sorting using Redis repository query methods

### LINKED ISSUE #2080: Support findAll with sorting in derived query methods
Consider this POJO code:

```java
@RedisHash("somehash")
public class Example {

    @Id
    private String id;
    @Indexed
    private String some;
    @Indexed
    private String thing;
    @Indexed
    private Integer rank;
```

The following code would be expected to return elements sorted:

```java
@Repository
public interface ExampleRepository extends CrudRepository<Example, string> {
    List<Example> findAllOrderByRank(Integer rank);
}
```

As described in [this StackOverflow post](https://stackoverflow.com/q/54339026)) and corroborated by several users in the comments (and my own project), this does return all elements, but they are not sorted as expected. Indeed, [the table of supported keywords inside method names](https://docs.spring.io/spring-data-redis/docs/current/reference/html/#redis.repositories.queries) does not list `OrderBy` as a valid keyword.

The [section for Query By Example](https://docs.spring.io/spring-data-redis/docs/current/reference/html/#query-by-example.running) also specifically says that `findAll` with sorting is not supported.

This is important functionality that I would like to see supported in future releases.

mp911de:
We now correctly apply sorting when calling a repository query method with a Sort parameter or when defining the sort order as part of the method name.

Closes #2080

### DIFF
--- a/src/main/asciidoc/reference/redis-repositories.adoc
+++ b/src/main/asciidoc/reference/redis-repositories.adoc
@@ -673,7 +673,6 @@ public interface PersonRepository extends CrudRepository<Person, String> {
 ----
 ====
 
-
 NOTE: Please make sure properties used in finder methods are set up for indexing.
 
 NOTE: Query methods for Redis repositories support only queries for entities and collections of entities with paging.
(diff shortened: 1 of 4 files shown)


src/main/asciidoc/reference/redis-repositories.adoc
```diff
@@ -711,6 +710,26 @@ The following table provides an overview of the keywords supported for Redis and
...
+
+Redis repositories allow various approaches to define sorting order. Redis itself does not support in-flight sorting when retrieving hashes or sets. Therefore, Redis repository query methods construct a `Comparator` that is applied to the result before returning results as `List`. Let's take a look at the following example:
+
+.Sorting Query Results
+====
+[source,java]
+----
+interface PersonRepository extends RedisRepository<Person, String> {
+
+  List<Person> findByFirstnameSortByAgeDesc(String firstname); <1>
```
Thunderforge, line 724: Shouldn't this be `OrderBy` instead of `SortBy`? The former is what the unit test uses.

mp911de: Good catch. This snippet originates from the MongoDB documentation that needs to be updated as well. Also, we should document the `OrderBy` keyword in the table of supported keywords.

christophstrobl:
merged to 2.5.x and forward ported to master

Issue #2096, opened 2021-06-23
Secondary indexes not removed upon keyspace removal event if colon in name


gpizarro-gfm:
This may be related to/a duplicate of #1326, but that issue was closed back in 2018. 

For this Kotlin example, say I have:
```kotlin
@RedishHash(value = "keysubname", timeToLive = 10L)
data class KeySubname(
    @Id
    val id: Long
    @Indexed
    val ext_id: Long
)
```

And that keyspace events are enabled in the Spring application:
```kotlin
@EnableRedisRepositories(enableKeyspaceEvents = RedisKeyValueAdapter.EnableKeyspaceEvents.ON_STARTUP)
```

After saving the hash, it and its index, secondary index, and lookup hashmap are deleted after 10 seconds. This is expected behavior. 

However, if I add a colon to the hash's name (and make no other changes):
```kotlin
@RedishHash(value = "key:subname", timeToLive = 10L)
data class KeySubname(
    @Id
    val id: Long
    @Indexed
    val ext_id: Long
)
```

Then the secondary index and lookup hashmap are not deleted after 10 seconds, leaving a permanent record in the redis db of a hash that no longer exists.

Hopefully this issue can be fixed soon, as it's affecting repositories that are currently in production. Please let me know how else I can help.

mp911de:
Double-colon is considered to be a separator character and as per https://github.com/spring-projects/spring-data-redis/pull/2100#issuecomment-881256388, allowing double-colon usage in the keyspace name requires a wider revision of the repository support.

GFriedrich:
@mp911de: Could this ticket be updated to a more up-to-date milestone? Having this feature would be nice, because some Redis UIs also use the colon to group keys. So using colons allow for organizing keys in a more structured way.

petenattress:
Can I ask if this behaviour is documented anywhere beyond this GH issue? We recently experienced a memory leak in our Redis instance due to the cleanup behaviour failing because of this. Given that colons are canonically used as separator characters in Redis key names it feels quite easy to run afoul of this, and the impact is not obvious beyond the Redis instance slowly running out of space.

PR #2100, opened 2021-06-27
 Support for key containing colons.

home-body:
#2096 
I think this can be solved, please check.


gpizarro-gfm:
@home-body Thanks for taking a look. Do you have an ETA for which release might include this fix?

mp911de:
Thank you @home-body for looking into this issue. Our entire code assumes double-colon (`:`) as a separator between the keyspace and any identifier so the entire codebase contains usage of double-colon. We use multiple levels of information when it comes to keys/identifiers (the actual hash, secondary indexes) and we cannot simply assume that the last part of any given identifier is going to be the thing we're looking for.

In fact, to enable double-colon to be used in the keyspace attribute, we need to revisit and update all places that work with identifiers to introduce an abstraction to properly create and introspect identifiers.

The submitted changes do not cover the full range of changes that are required to make this work. For the time being, please avoid using `:` in the keyspace name and we're closing this pull request without a merge.

Issue #2116, opened 2021-07-07
Introduce factory methods to configure `LettuceConnectionFactory` from `RedisURI`


mp911de:
Lettuce supports multiple URI schemes (`redis-sentinel://`, `redis-socket://`, …). It would make sense to provide a factory method on `LettuceConnectionFactory` that takes, `RedisURI` or `URI` and returns a properly configured `RedisConfiguration` object for easier configuration of the connection factory.

PR #2117, opened 2021-07-12
Add factory method to `LettuceConnectionFactory` to create `RedisConfiguration` from `RedisURI`

bono007:
Fixes gh-2116

@markpollack I know this still needs formatting and tests written but I wanted to push the proposal that I have so far to make sure this is what you had in mind. I will follow on w/ the formatting and tests once you confirm (or not). 

Thanks

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
@@ -114,6 +114,7 @@ import static org.springframework.data.redis.connection.lettuce.LettuceConnectio
  * @author Ruben Cervilla
  * @author Luis De Bello
  * @author Andrea Como
+ * @author Chris Bono
  */
 public class LettuceConnectionFactory
 		implements InitializingBean, DisposableBean, RedisConnectionFactory, ReactiveRedisConnectionFactory {
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
@@ -15,16 +15,7 @@
  */
 package org.springframework.data.redis.connection.lettuce;
 
-import java.nio.ByteBuffer;
-import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.TimeUnit;
-import java.util.function.Consumer;
-import java.util.stream.Collectors;
+import static org.springframework.data.redis.connection.lettuce.LettuceConnection.*;
 
 import io.lettuce.core.AbstractRedisClient;
 import io.lettuce.core.ClientOptions;
@@ -39,6 +30,18 @@ import io.lettuce.core.cluster.RedisClusterClient;
 import io.lettuce.core.cluster.api.StatefulRedisClusterConnection;
 import io.lettuce.core.codec.RedisCodec;
 import io.lettuce.core.resource.ClientResources;
+
+import java.nio.ByteBuffer;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
```diff
@@ -15,7 +15,16 @@
  */
 package org.springframework.data.redis.connection.lettuce;
 
-import static org.springframework.data.redis.connection.lettuce.LettuceConnection.*;
+import java.nio.ByteBuffer;
```
bono007, line 18: Yep - I will adjust this tomorrow am.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
```diff
@@ -279,6 +296,32 @@ public LettuceConnectionFactory(RedisClusterConfiguration clusterConfiguration,
...
+	 *
+	 * @param redisURI the connection info in the format of a RedisURI
+	 * @return an appropriate {@link RedisConfiguration} instance representing the Redis URI.
+	 */
+	public static RedisConfiguration redisConfigurationFromRedisUri(RedisURI redisURI) {
+
+		Assert.notNull(redisURI, "RedisURI must not be null");
+
+		if (!ObjectUtils.isEmpty(redisURI.getSentinels())) {
+			return LettuceConverters.redisUriToSentinelConfiguration(redisURI);
```
bono007, line 315: I went ahead and followed suit w/ LettuceConverters housing the conversions as it is the reverse direction of the `sentinel config to redis uri` that was already in there. I am happy moving elsewhere if need be.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
```diff
@@ -279,6 +297,32 @@ public LettuceConnectionFactory(RedisClusterConfiguration clusterConfiguration,
...
+	 * <ul>
+	 * <li>If {@code redisURI} has sentinel info a {@link RedisSentinelConfiguration} is returned</li>
+	 * <li>If {@code redisURI} has socket info a {@link RedisSocketConfiguration} is returned</li>
+	 * <li>Otherwise a {@link RedisStandaloneConfiguration} is returned</li>
+	 * </ul>
+	 *
+	 * @param redisURI the connection info in the format of a RedisURI
+	 * @return an appropriate {@link RedisConfiguration} instance representing the Redis URI.
+	 */
+	public static RedisConfiguration redisConfigurationFromRedisUri(RedisURI redisURI) {
```
mp911de, line 311: How about naming this method `createRedisConfiguration`?

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConverters.java
```diff
@@ -535,6 +566,86 @@ public static RedisURI sentinelConfigurationToRedisURI(RedisSentinelConfiguratio
...
 	}
 
+	/**
+	 * Converts a {@link RedisURI} to its corresponding {@link RedisSentinelConfiguration}.
+	 *
+	 * @param redisURI the uri containing the Redis Sentinel connection info
+	 * @return a {@link RedisSentinelConfiguration} representing the Redis Sentinel information in the Redis URI.
+	 * @since 2.6
+	 */
+	public static RedisSentinelConfiguration redisUriToSentinelConfiguration(RedisURI redisURI) {
```
mp911de, line 576: Let's keep these methods package-private. We don't want to expose these any further.

mp911de (changes requested):
Thanks for your pull request. I left a few comments. You can find the formatter settings at https://github.com/spring-projects/spring-data-build/tree/main/etc/ide

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConverters.java
bono007: Yes, please. I was not sure if the pattern was to keep these public as `sentinelConfigurationToRedisURI` is public and should have called that out w/ a comment. I too am a member of the "principle of least privilege" camp.

bono007:
Adding tests now.

bono007:
@mp911de well I just realized that "Mark Pollack" != "Mark Paluch"  🤦

I am leaving WIP as I want to consume these changes locally to verify it is what we want for https://github.com/spring-projects/spring-boot/issues/21920 so as to avoid any churn in case this proposal needs to be tweaked slightly. Of course, if you would rather merge it before I do that - that is fine too.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConverters.java
```diff
@@ -535,6 +527,86 @@ public static RedisURI sentinelConfigurationToRedisURI(RedisSentinelConfiguratio
...
+	/**
+	 * Converts a {@link RedisURI} to its corresponding {@link RedisSentinelConfiguration}.
+	 *
+	 * @param redisURI the uri containing the Redis Sentinel connection info
+	 * @return a {@link RedisSentinelConfiguration} representing the Redis Sentinel information in the Redis URI.
+	 * @since 2.6
+	 */
+	static RedisSentinelConfiguration redisUriToSentinelConfiguration(RedisURI redisURI) {
+
+		Assert.notNull(redisURI, "RedisURI is required");
```
bono007, line 539: These `notNull` asserts are a bit redundant currently as the only place that calls this current is `LettuceConnectionFactory.createRedisConfiguration` which itself does this null check. The only reason I left it in here is in case it gets called elsewhere (package protected).

src/test/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactoryCreateRedisConfigurationUnitTests.java
```diff
@@ -0,0 +1,193 @@
...
+import org.springframework.data.redis.connection.RedisSentinelConfiguration;
+import org.springframework.data.redis.connection.RedisSocketConfiguration;
+import org.springframework.data.redis.connection.RedisStandaloneConfiguration;
+
+/**
+ * Unit tests for the {@link LettuceConnectionFactory#createRedisConfiguration(RedisURI)} factory method.
+ *
+ * @author Chris Bono
+ */
+class LettuceConnectionFactoryCreateRedisConfigurationUnitTests {
```
bono007, line 37: A few things to note about this test:

1. I opted to test the functionality at `LettuceConnectionFactory.createRedisConfiguration` rather than directly on the `LettuceConverters.redisUriTo<T>Configuration` as the tests would have been almost identical. The coverage is the same doing it this way. The assumption is that the LC methods are an internal impl of this API. Also, because the LC methods are static, just unit testing the routing in LCF to LC (that is all that the LCF really is doing) would have been problematic to mock. 

2. I broke it out into its own specific LCF unit test as the other one is getting pretty big. 

3. I leveraged `@Nested` to group the "flavor" of RedisConfiguration under test.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConverters.java
mp911de: Guarding `public` methods only is sufficient. I can take care of this during the merge.

src/test/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactoryCreateRedisConfigurationUnitTests.java
mp911de: I'd suggest introducing proper equals/hashCode methods, that's something we can do during the merge.

mp911de:

src/test/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactoryCreateRedisConfigurationUnitTests.java
bono007: Yeh good call. I was wrapped up in Lettuce specific-land that I did not realize the config objects were not lettuce constructs - they are clearly spring data redis and I could have added EHC rather than the AssertJ field by field config approach.

Issue #2156, opened 2021-09-07
`JedisClusterKeyCommands ` makes extra `KEYSLOT` calls to the server


upthewaterspout:
In `JedisClusterKeyCommands`, most of the commands find the node for a key using `connection.getTopologyProvider().getTopology().getKeyServingMasterNode(key)`. However, there are a few commands like [restore](https://github.com/spring-projects/spring-data-redis/blob/a944aa2d1bd3551cf2146cdf4e03ed10c1f112a3/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterKeyCommands.java#L515) that  call `connection.clusterGetNodeForKey`. 

`clusterGetNodeForKey` actually ends up making a call to the redis server to execute the `CLUSTER KEYSLOT` command to determine the slot for the key. This is inefficient because it requires an extra round trip to the server to compute the slot of the key. 

I can't see any particular reason why these commands are not calling `getKeyServingMasterNode`. That method computes the slot for the key on the client side, which is more efficient.

mp911de:
Good catch. Do you want to submit a pull request to fix that issue?

upthewaterspout:
Sure, I created a PR for this change - #2159.

I ran `make test` - which looks like it is running the unit test that exercises this code. I'm not quite sure what the best way to add new tests for this change is - or if that makes sense. I could add a unit test of JedisClusterKeyCommands that mocks the connection and asserts that clusterGetNodeForKey isn't called, but I'm not sure that has much value.

PR #2159, opened 2021-09-08
Hash keys locally rather than call cluster keyslot

upthewaterspout:
Changing the remaining commands in JedisClusterKeyCommands to use the
topology's getKeyServingMasterNode to find the node of key, rather than calling
connection.clusterGetNodeForKey. clusterGetNodeForKey was making a cluster
keyslot call to the server to hash the key, rather than hashing it locally.

Closes: #2156


mp911de:
I think it's easier to switch `clusterGetNodeForKey ` to calculate the slot locally instead of obtaining the `TopologyProvider`. We have the same behavior already in place for Lettuce, so it makes sense to align the Jedis implementation.

PR #2162, opened 2021-09-10
Extract creating of connection in ReactiveRedisTemplate into a separate method

### LINKED ISSUE #2145: Enable `ReactiveRedisTemplate` for multi-tenancy usage
Hi,

I would like to use `ReactiveRedisTemplate` in multitenant environment. In short, the option to use single `ReactiveRedisTemplate` which can connect to multiple Redis instances based on tenant values passed in reactive context.
Unfortunately, I can not connect the reactive context with the connection factory so I can create different connections (or using different connection pool) based on tenant value in the context.

One option is to override doInConnection, but it has private modifier. Changing it will allow me to provide custom logic about connection creation.

petromir:
A new method, called `getConnection`, is introduced in `ReactiveRedisTemplate` to allow overriding of connection creation.

Closes #2145


mp911de:

petromir:
You are welcome :)

PR #2189, opened 2021-11-02
Inconsistent key names in `RENAME`/`RENAMENX` command documentation

toddmerrill:


toddmerrill:
I also see the name 'oleName' in the classes StringRedisConnection, ReactiveKeyCommands and ReactiveRedisOperations.  At the least I think they should be 'oldName', but is there a reason we don't use the same names for the parameters as Redis does?  https://redis.io/commands/rename  They use 'key' and 'newKey'.

mp911de:
Thanks for the PR, I'm going to merge it. It would make sense to rename these for better precision in the documentation.

mp911de:

PR #2206, opened 2021-12-02
Nullability annotations for indexOf methods in ListOperations

alturkovic:
`indexOf` and `lastIndexOf` methods are missing the `@Nullable` annotation causing issues for nullability checks in Kotlin.


mp911de:
See also #2208

PR #2225, opened 2022-01-05
Update `RedisOperations#convertAndSend` to return number of clients receiving the message.

### LINKED ISSUE #2209: Update `RedisOperations#convertAndSend` to return number of subscribers
Currently, `RedisOperations#convertAndSend` is `void` event though the underlying `RedisPubSubCommands#publish` returns a `Long` that represents the number of subscribers receiving the published message (see https://redis.io/commands/publish).

The information about the number of subscribers can be useful in some scenarios, and is also returned by the reactive counterpart in `ReactiveRedisOperations#convertAndSend`.

Digging through the history of `RedisOperations`, it seems that `#convertAndSend` used to return `Long` for a brief period of time in 2013: it was introduced in b35f6a3c and reverted in d46bee7e.

I understand that adding return value would be a breaking change, but the impact doesn't seem to be too big (?). I'm willing to submit a PR to address this, if the change is welcome.

vpavic:
Currently, `RedisOperations#convertAndSend` is void event though the underlying `RedisPubSubCommands#publish` returns a `Long` that represents the number of clients receiving the published message (see https://redis.io/commands/publish).

This commit updates `RedisOperations#convertAndSend` to return the number of clients that received the message.

Closes #2209


src/test/java/org/springframework/data/redis/listener/PubSubResubscribeTests.java
```diff
@@ -147,22 +149,22 @@ void testContainerPatternResubscribe() throws Exception {
 		container.addMessageListener(anotherListener, new PatternTopic(PATTERN));
 
 		// Wait for async subscription tasks to setup
-		Thread.sleep(400);
-
-		// test no messages are sent just to patterns
-		template.convertAndSend(CHANNEL, payload1);
-		template.convertAndSend(ANOTHER_CHANNEL, payload2);
+		await().atMost(Duration.ofMillis(600)).untilAsserted(() -> {
```
vpavic, line 152: I utilized Awaitility in use cases like this as it's more handy vs having to resort to `Thread#sleep`. Note that for this particular set of assertions I had to increase the grace period from 400 to 600 ms as the test was flaky, on my machine at least.

mp911de:
We're considering this change for the next Spring Data Redis 3.0 milestone.

mp911de:

PR #2228, opened 2022-01-07
Add trimming support for xAdd

Quantum64:
This change adds support for trimming and approximate trimming to reactive stream commands. This is beneficial for the common use case of adding and immediately trimming a stream because it reduces the number of Redis commands needed for the operation from two to one.

Test cases will be added shortly.


mp911de:
Thanks for submitting a pull request. We have now a couple of PR's for the same topic and we're going to merge #2247.

PR #2236, opened 2022-01-13
Allow configure CacheManagerBuilder defaults base on current defaults

quaff:
For example:
```java
@Bean
RedisCacheManagerBuilderCustomizer redisCacheManagerBuilderCustomizer() {
	return builder -> builder.cacheDefaults(builder.cacheDefaults()
		.serializeValuesWith(SerializationPair.fromSerializer(new GenericJackson2JsonRedisSerializer())));
}
```


mp911de:
Care to elaborate what you want to achieve? The default cache config can be obtained from `RedisCacheConfiguration.defaultCacheConfig()` respective `RedisCacheConfiguration.defaultCacheConfig(ClassLoader)`. Builders aren't intended for the retrieval of values. Looking at the builder, there are several complex objects that might be customized and we do not want to expose these to retain encapsulation.

PR #2238, opened 2022-01-13
feature:getSynchronized method performance update.

LL-sanmu-LL:
hi，
  In my company's project,We want to use @Cacheable(...sync = true)annotations as cache.But when we find a serious problems that the method performance is very poor in multi-thread jvm env。In my project，we need to cache large number of cache key，and expect a excellent performance。So we try to read the RedisCache  code，and find the cause is from `private synchronized <T> T getSynchronized(Object key, Callable<T> valueLoader)`。This method is synchronized。If we create much cache in a short time，this method's performance will degenerate quickly.
 So we try to assign a ReentrantLock to every cache key.At the time,we want it will not expend too much stack space to store the lock,because the key's quantity is so large.We find an idea that we use weakreference to hold this lock,when the lock is not use in any thread,it will be release by gc.
  The code is running in our project,and i think it will be a good feature modify,So we report this modify to you hope to help others.
 At last,it's my first time to commit code in such excelent open source project.If having any problem，we hope to receive any reply from you.Thank you for your read.

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -60,12 +60,12 @@ public class RedisCache extends AbstractValueAdaptingCache {
 	private final RedisCacheWriter cacheWriter;
 	private final RedisCacheConfiguration cacheConfig;
 	private final ConversionService conversionService;
-    private final WeakHashLock weakHashLock = new WeakHashLock();
+	private final WeakHashLock weakHashLock = new WeakHashLock();
 
 	/**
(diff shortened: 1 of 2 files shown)


mp911de:
The `synchronized` lock is in place to exclusively lock the cache without a timeout bypass. The goal of the method is to enforce the serialization of calls by blocking these. If you need a customized behavior, then please either subclass `RedisCache` and maintain your changes there or implement your own Redis `Cache`. We do want to keep the code as-is.

LL-sanmu-LL:
您的邮件已经收到。

PR #2256, opened 2022-02-09
Revise implementation of RedisMessageListenerContainer

### LINKED ISSUE #964: RedisMessageListenerContainer has race conditions [DATAREDIS-389]
**[Adrian Riley](https://jira.spring.io/secure/ViewProfile.jspa?name=adrianriley)** opened **[DATAREDIS-389](https://jira.spring.io/browse/DATAREDIS-389?redirect=false)** and commented

The lazyListen() method sets the listening flag to true, then starts a Thread to run the SubscriptionTask. Calls to addMessageListener() then use the same JedisConnection to subscribe to further topics. But there is no guard on the SubscriptionTask startup process, a call to addMessageListener() may arrive before that process is complete, so the subscription may be lost entirely or both threads may write to the same output stream at the same time, so Redis is sent a corrupted command.

There is a similar issue when the task is shutdown after all topics have been unsubscribed. It is possible for a new subscribe command to be sent on the connection before it is returned to the pool. If that connection is then used for non-subscription commands, an error occurs.

Attached is a TestNG test which shows the problem, sometimes. You can modify the two constants to control the multi-threaded execution, but for me it usually fails every few times I run it. You may also hit the issue https://github.com/xetorthio/jedis/issues/933

---

**Affects:** 1.5 GA (Fowler)

**Reference URL:** http://stackoverflow.com/questions/29353615/spring-data-redis-redismessagelistenercontainer-seems-to-have-race-conditions

**Attachments:**
- [SpringDataRedisPubSubIT.java](https://jira.spring.io/secure/attachment/22722/SpringDataRedisPubSubIT.java) (_6.50 kB_)

4 votes, 6 watchers

mp911de:
`RedisMessageListenerContainer` is now reimplemented using non-blocking synchronization guards and state management to simplify its maintenances. Additionally, listener registration and subscription setup through the `start()` method waits until the listener subscription is confirmed by the Redis server. The synchronization removes potential race conditions that could happen by concurrent access to blocking Redis connectors in which the registration state was guessed and not awaited.

Closes #964


christophstrobl:
Merged to _main_ development line and fw ported to _3.0.x_.

PR #2276, opened 2022-03-07
Remove long deprecated API and refine API

### LINKED ISSUE #2273: Remove deprecated `LettucePool` and other long deprecated API

mp911de:
This pull request removes long deprecated API and methods that should have been deprecated. Upon agreeing on which methods and API should be removed, we will deprecate these items in the 2.7 development line.

The defaulted Connection arrangement is now refined and will not be removed as removal will cause a difficult time for existing applications. Instead, using commands directly on the connection is now yielding a deprecation warning to guide users towards using the segregated command interfaces.

Also, this PR introduces a few suggestions towards bound keys API and `RedisTemplate` API refinements that should be split into another pull request once the deprecation removal is merged.

Closes #2273


christophstrobl:
merged to _3.0.x_

PR #2297, opened 2022-04-09
Fix a broken link in the redis-transaction docuement.

dveamer:
The link to "declarative transaction management" was broken.


mp911de:

PR #2299, opened 2022-04-11
Improve thread safety of RedisMessageListenerContainer.

jebeaudet:
- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).

Hi,

As instructed in #964, a PR to fix the issues I've encountered in RedisMessageListenerContainer.

I've commented the PR inline to ease the comprehension. Let me know if you need further details.

Testing those kind of multithreaded behavior is never easy but I believe all my additions have been tested in the unit tests I've added.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisSubscription.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisSubscription.java
@@ -42,11 +42,13 @@ class JedisSubscription extends AbstractSubscription {
 	 */
 	@Override
 	protected void doClose() {
-		if (!getChannels().isEmpty()) {
-			jedisPubSub.unsubscribe();
-		}
-		if (!getPatterns().isEmpty()) {
-			jedisPubSub.punsubscribe();
+		synchronized (this) {
+			if (!getChannels().isEmpty()) {
+				jedisPubSub.unsubscribe();
+			}
+			if (!getPatterns().isEmpty()) {
+				jedisPubSub.punsubscribe();
+			}
 		}
 	}
 
@@ -56,7 +58,9 @@ class JedisSubscription extends AbstractSubscription {
 	 */
 	@Override
 	protected void doPsubscribe(byte[]... patterns) {
-		jedisPubSub.psubscribe(patterns);
+		synchronized (this) {
+			jedisPubSub.psubscribe(patterns);
+		}
 	}
 
 	/*
@@ -65,10 +69,12 @@ class JedisSubscription extends AbstractSubscription {
 	 */
 	@Override
 	protected void doPUnsubscribe(boolean all, byte[]... patterns) {
-		if (all) {
-			jedisPubSub.punsubscribe();
-		} else {
-			jedisPubSub.punsubscribe(patterns);
+		synchronized (this) {
+			if (all) {
+				jedisPubSub.punsubscribe();
+			} else {
+				jedisPubSub.punsubscribe(patterns);
+			}
 		}
 	}
 
@@ -78,7 +84,9 @@ class JedisSubscription extends AbstractSubscription {
 	 */
 	@Override
 	protected void doSubscribe(byte[]... channels) {
-		jedisPubSub.subscribe(channels);
+		synchronized (this) {
+			jedisPubSub.subscribe(channels);
+		}
 	}
 
 	/*
@@ -87,10 +95,12 @@ class JedisSubscription extends AbstractSubscription {
 	 */
 	@Override
 	protected void doUnsubscribe(boolean all, byte[]... channels) {
-		if (all) {
-			jedisPubSub.unsubscribe();
-		} else {
-			jedisPubSub.unsubscribe(channels);
+		synchronized (this) {
+			if (all) {
+				jedisPubSub.unsubscribe();
+			} else {
+				jedisPubSub.unsubscribe(channels);
+			}
 		}
 	}
 }
--- a/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
+++ b/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
@@ -25,6 +25,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -69,6 +70,7 @@ import org.springframework.util.ErrorHandler;
  * @author Way Joke
  * @author Thomas Darimont
  * @author Mark Paluch
+ * @author Jacques-Etienne Beaudet
  */
 public class RedisMessageListenerContainer implements InitializingBean, DisposableBean, BeanNameAware, SmartLifecycle {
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/springframework/data/redis/connection/jedis/JedisSubscription.java
```diff
@@ -42,11 +42,13 @@ class JedisSubscription extends AbstractSubscription {
 	 */
 	@Override
 	protected void doClose() {
-		if (!getChannels().isEmpty()) {
-			jedisPubSub.unsubscribe();
-		}
-		if (!getPatterns().isEmpty()) {
-			jedisPubSub.punsubscribe();
+		synchronized (this) {
```
jebeaudet, line 45: For this class, synchronization would technically not be necessary alongside my other changes. But I figure it's not a bad idea to add it in case there's a race condition somewhere else in Spring code that I've missed. 

This class is not thread safe since commands like `jedisPubSub::unsubscribe` write to the non synchronized connection `OutputStream` then flush it after. Therefore, if it's accessed by many threads, you can pollute the `OutputStream`.

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -69,6 +70,7 @@
  * @author Way Joke
  * @author Thomas Darimont
  * @author Mark Paluch
+ * @author Jacques-Etienne Beaudet
```
jebeaudet, line 73: Let me know if my changes aren't worthy enough to be here, I don't really mind I was mainly following the contributions guideline!

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -689,36 +690,55 @@ private class SubscriptionTask implements SchedulingAwareRunnable {
 		private class PatternSubscriptionTask implements SchedulingAwareRunnable {
 
 			private long WAIT = 500;
-			private long ROUNDS = 3;
```
jebeaudet, line 692: This pattern subscription thread will now run until it's canceled or if it's done.

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -689,36 +690,55 @@ private class SubscriptionTask implements SchedulingAwareRunnable {
...
+			private AtomicBoolean isThreadRunning = new AtomicBoolean(true);
+
 
 			public boolean isLongLived() {
 				return false;
 			}
 
+			void cancel() {
+				isThreadRunning.set(false);
```
jebeaudet, line 701: this is called exclusively by the `SubscriptionTask` when there is a `RedisConnectionFailureException` or when the container is stopped.

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -689,36 +690,55 @@ private class SubscriptionTask implements SchedulingAwareRunnable {
...
-								done = true;
-								connection.getSubscription().pSubscribe(unwrap(patternMapping.keySet()));
-							} else {
-								try {
-									Thread.sleep(WAIT);
-								} catch (InterruptedException ex) {
-									Thread.currentThread().interrupt();
+							try {
+								if (connection.isSubscribed()) {
+									synchronized (localMonitor) {
```
jebeaudet, line 712: this is synchronized by `localMonitor` to make sure the thread exit cleanly when canceled by the `SubscriptionTask`. This should makes the shutdown more graceful and ordered if the container is shutdown while booting and subscription to patterns

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -689,36 +690,55 @@ private class SubscriptionTask implements SchedulingAwareRunnable {
...
+									try {
+										Thread.sleep(WAIT);
+									} catch (InterruptedException ex) {
+										logger.info("PatternSubscriptionTask was interrupted, exiting.");
+										Thread.currentThread().interrupt();
+										return;
+									}
+								}
+							} catch(Throwable e) {
+								if (e instanceof RedisConnectionFailureException) {
```
jebeaudet, line 726: connection errors will now be retried, like it is on the `SubscriptionTask`. This could sneakily exit before and without proper exception handling, the stack would be written to the application `stderr` and could go unnoticed.

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -689,36 +690,55 @@ private class SubscriptionTask implements SchedulingAwareRunnable {
...
+									if (isRunning() && isThreadRunning.get()) {
+										logger.error("Connection failure occurred on pattern subscription task. Restarting subscription task after " + recoveryInterval + " ms");
+										sleepBeforeRecoveryAttempt();
+									}
+								} else {
+									logger.error("PatternSubscriptionTask aborted with exception:", e);
```
jebeaudet, line 732: non redis connection errors will exit the thread like `SubscriptionTask`

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -689,36 +690,55 @@ private class SubscriptionTask implements SchedulingAwareRunnable {
...
+								if (connection.isSubscribed()) {
+									synchronized (localMonitor) {
+										connection.getSubscription().pSubscribe(unwrap(patternMapping.keySet()));
+										done = true;
+									}
+								} else {
+									try {
+										Thread.sleep(WAIT);
+									} catch (InterruptedException ex) {
+										logger.info("PatternSubscriptionTask was interrupted, exiting.");
```
jebeaudet, line 720: Added info log to make sure debugging is not too hard if you're chasing what's up, let me know if you want this in debug/trace but I figure it shouldn't hit much in a real life scenario.

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -759,6 +779,7 @@ public void run() {
 					}
 				}
 			} catch (Throwable t) {
+				cancelPatternSubscriptionTask();
```
jebeaudet, line 782: since this whole code block is rerun on error, I cancel the task here so that everything is clean for the next run. Since it's synchronized by `localMonitor`, everything should be ordered.

jebeaudet:
Thanks for you insight.

While I agree on the GA part, to me, it's still a bug and bugs should be fixed. Especially this one since you could have a pattern subscription fail completely silently or a corrupted jedis stream. Also, version 2.5.x is the one managed from Spring Boot 2.5.x which is still in support for another month and EoL in August 2023 if my math is correct. Even Spring Boot 2.6.x, which is the current stable version, does not pull on the 2.7.x branch. This means that everyone using Spring Boot with a released version is currently affected. 

I'd also argue that since Spring Session is using the [problematic code path](https://github.com/spring-projects/spring-session/blob/main/spring-session-data-redis/src/main/java/org/springframework/session/data/redis/config/annotation/web/http/RedisHttpSessionConfiguration.java#L157) (having a topic + pattern subscription), it affects even more users and could potentially affect users that rely on `SessionCreatedEvent`.

I guess we could shuffle things a bit to reduce the footprint of the changes. Basically this PR fixes : 
1. The non thread safe usage of `JedisSubscription`
2. The non resilient behavior of the `PatternSubscriptionTask` (exceptions are uncaught and it only retries 3 times)

Just synchronizing `JedisSubscription` would avoid corrupting the jedis stream and would not add much overhead. With this, we could bump the `ROUNDS` variable along with a larger try-catch to make sure the `PatternSubscriptionTask` does not exit silently and keep trying. This would reduce the changes a bit and we could target this PR against the 2.7.x branch and the proposed one agains 2.5.x and 2.6.x.

Unless you have another idea?

Thanks again

jebeaudet:
WDYT @mp911de ?

Thanks

mp911de:
Version 2.5 is no longer maintained.

Issue #2309, opened 2022-04-22
HashOperations MultiGet Probably shouldn't return null elements in the collection


ShaneLee:
Hi, 

There was an[ issue raised a while ago](https://github.com/spring-projects/spring-data-redis/pull/340) which supported null values in the collection that is returned from the `ReactiveHashOperations multiGet` method. Whilst this fixes the original NPE, it creates other NPEs for clients using this code. For example, if the client code translates the returned collection to a Flux using Flux.fromIterable this will throw an NPE. When this happens the NPE that is throw is quite tricky to isolate as the stack trace that comes out often has no reference to the client code for example:

```
[...]

PR #2321, opened 2022-05-12
Prevent HashOperations MultiGet from returning null values in the collection

ShaneLee:
This PR addresses issue #2309. It prevents the multi-get operation from returning null values inside the collection. This is an issue that can be very difficult to diagnose as it leads to stack traces without reference to the client code. 

I've raised this PR against `main` for now. I expect it would be better going into `2.7.x`, but happy to put it into another branch.

- [x ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [x ] You submit test cases (unit or integration tests) that back your changes.
- [ x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/DefaultReactiveHashOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultReactiveHashOperations.java
@@ -19,11 +19,12 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


mp911de:
Duplicates #2402

PR #2328, opened 2022-05-18
Fix Jedis 4.2 version typo

sazzad16:


christophstrobl:
Thank you @sazzad16 .

Issue #2374, opened 2022-07-28
Add support for deserializing objects with generics


JosRoseboom:
Currently there is no RedisSerializer that can deserialize an object with generics like Set\<Person>. This is for example a problem when using redis as cache in spring boot:

```
 @Cacheable("personCache)
 public Set<Person> getPersons()  {
   // expensive code
 }
```

Using for example Jackson2JsonRedisSerializer, we could specify Set as type, but it won't handle the Person type. It deserializes to Set<LinkedHashMap> with the LinkedHashMap having the properties of the person. A cache hit results in a ClassCastException because a LinkedHashMap cannot be cast to a Person (similarly, a Set of enums is deserialized as a Set of Strings).

Classes using a generic (Person in this case) should not be forced to implement Serializable, so Jackson based serialization in itself is nice.

Note: Created this issue to refer to in pull request: pull request on the way

JosRoseboom:
Hi Mark,

thanks for your fast response. I created a pull request: #2375 

One of the tests I added shows a serializer that could make the Set\<Person> case in my original post work:

    @Test // GH-2374
	void testSetOfPersons() {
		PersonObjectFactory personFactory = new PersonObjectFactory();
		Set<Person> personSet = Set.of(personFactory.instance(), personFactory.instance());
		JacksonTypeReference2JsonRedisSerializer<Set<Person>> personSetSerializer = 
                      new JacksonTypeReference2JsonRedisSerializer<>(new TypeReference<>() {});

		assertThat(personSetSerializer.deserialize(personSetSerializer.serialize(personSet))).isEqualTo(personSet);
	}

In a Spring project that uses Redis as cache for @Cacheable as, in a configuration Bean for redis cache, I could do something like:

```
@Configuration
public class RedisCacheConfig {

    @Bean
    public RedisCacheManagerBuilderCustomizer redisCacheManagerBuilderCustomizer() {
        JacksonTypeReference2JsonRedisSerializer<Set<Person>> serializer = new JacksonTypeReference2JsonRedisSerializer<>(new TypeReference<>() {
        });
        RedisCacheConfiguration conf = RedisCacheConfiguration.defaultCacheConfig()
                .serializeValuesWith(RedisSerializationContext.SerializationPair.fromSerializer(serializer));
        return builder -> builder.withCacheConfiguration("personCache", conf);
    }

}


```

[...]

PR #2375, opened 2022-07-28
Add support for deserializing objects with generics.

JosRoseboom:
Currently there is no RedisSerializer that can deserialize an object with generics like Set\<Person>. This is for example a problem when using redis as cache in spring boot. See #2374 

### DIFF
--- a/src/main/java/org/springframework/data/redis/serializer/JacksonTypeReference2JsonRedisSerializer.java
+++ b/src/main/java/org/springframework/data/redis/serializer/JacksonTypeReference2JsonRedisSerializer.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2022-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


mp911de:
The Jackson serializers are rather lightweight and aim to solve the general case of serializing individual objects. Collections do not work well because we do not have a good way to store typing information. We recommend using JDK serialization if your input can be trusted. Other than that, have you seen `Jackson2JsonRedisSerializer` accepting a `JavaType`?

Beyond that, we do not want to add another Jackson serializer. It makes sense to reach out to the team before proposing a code change.

JosRoseboom:
Thanks for your reply. A pity it got rejected. 

> We recommend using JDK serialization if your input can be trusted

Preferably, we do not want to change our classes (adding implements Serializable) to have them cached. On top of that, we'd like to follow the recommendation to stay away from Java serialization.

> Other than that, have you seen Jackson2JsonRedisSerializer accepting a JavaType

Yes, but unfortunately that does not make caching of generics like Set\<Person> possible like the pull request does.

> It makes sense to reach out to the team before proposing a code change.

I tried to reach out to the team by describing the problem I faced in https://github.com/spring-projects/spring-data-redis/issues/2374 in combination with sharing the way that I solved it in the project where I faced it in this pull request. What would be the proper way of reaching out to the team in such case?

Am I right to conclude that spring-data-redis is not interested in adding the possibility of serializing generics like Set\<Person> for objects that are not Serializable? Or is it just that the way that this functionality is implemented in this pull request is not how the team likes it?

Thanks for all the good work in this project,
Jos

mp911de:
> I tried to reach out to the team by describing the problem I faced in https://github.com/spring-projects/spring-data-redis/issues/2374

Sorry, this was my bad. I didn't connect the dots regarding your issue. I suggest that you keep the desired serializer within your code base.

PR #2378, opened 2022-08-01
Fix the JDK version required to build from the source code #2377

lanicc:
Fix the JDK version description required to build the source code in the README doc


mp911de:
Thank you for your contribution. That's merged now.

PR #2397, opened 2022-09-02
Add missing hints for key bound operations.

christophstrobl:


src/main/java/org/springframework/data/redis/aot/RedisRuntimeHints.java
```diff
@@ -112,24 +141,31 @@ public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader)
...
+		boundOperationsProxy(BoundKeyOperations.class, classLoader, hints);
+		boundOperationsProxy(BoundListOperations.class, classLoader, hints);
+		boundOperationsProxy(BoundSetOperations.class, classLoader, hints);
+		boundOperationsProxy(BoundStreamOperations.class, classLoader, hints);
+		boundOperationsProxy(BoundValueOperations.class, classLoader, hints);
+		boundOperationsProxy(BoundZSetOperations.class, classLoader, hints);
 		boundOperationsProxy(
-				TypeReference.of("org.springframework.data.redis.core.BoundOperationsProxyFactory$DefaultBoundKeyOperations"),
-				hints);
+				TypeReference.of("org.springframework.data.redis.core.BoundOperationsProxyFactory$BoundKeyOperations"),
```
artembilan, line 153: Isn't this a typo?
Doesn't look that there is an inner class like this, but that `DefaultBoundKeyOperations` is really there.
I believe @wilkinsona 's point (as well as mine) to not have this proxy hint at all. The one for `BoundKeyOperations` should be enough. No? 
Maybe reflection for those methods in the  `BoundKeyOperations` has to be exposed?
Might not be related, but I used `@Reflective` on my interfaces' methods to give them access from SpEL...

src/main/java/org/springframework/data/redis/aot/RedisRuntimeHints.java
```diff
@@ -47,62 +48,90 @@
...
+						TypeReference.of(RedisKeyValueAdapter.class), TypeReference.of(RedisKeyValueTemplate.class),
+
+						// Key-Value
+						TypeReference.of(KeySpace.class), TypeReference.of(AbstractKeyValueAdapter.class),
+						TypeReference.of(KeyValueAdapter.class), TypeReference.of(KeyValueOperations.class),
+						TypeReference.of(KeyValueTemplate.class), TypeReference.of(KeyValueMappingContext.class),
+						TypeReference.of(KeyValueRepository.class), TypeReference.of(KeyValueRepositoryFactoryBean.class),
+						TypeReference.of(QueryCreatorType.class), TypeReference.of(KeyValuePartTreeQuery.class)),
+
+				hint -> hint.withMembers(MemberCategory.INVOKE_DECLARED_CONSTRUCTORS, MemberCategory.INVOKE_DECLARED_METHODS));
```
artembilan, line 134: Why have you changed from the `INVOKE_PUBLIC_METHODS` to this `INVOKE_DECLARED_METHODS`, please?

artembilan (changes requested):

src/main/java/org/springframework/data/redis/aot/RedisRuntimeHints.java
christophstrobl: capturing invocations with the `RuntimeHintsRecorder` indicated we need to have a broader reflection scope.

src/main/java/org/springframework/data/redis/aot/RedisRuntimeHints.java
christophstrobl: thanks @artembilan - refined hints.

PR #2446, opened 2022-11-02
Fix method usage in MappingRedisConverter.java

wzhix:
Replace `ObjectUtils.isEmpty` with `StringUtils.hasText` since `ObjectUtils.isEmpty(Boolean)` is always false.


src/main/java/org/springframework/data/redis/core/convert/MappingRedisConverter.java
```diff
@@ -317,7 +317,7 @@ protected Object readProperty(String path, RedisData source, RedisPersistentProp
 			return null;
 		}
 
-		if (persistentProperty.isIdProperty() && ObjectUtils.isEmpty(path.isEmpty())) {
+		if (persistentProperty.isIdProperty() && !StringUtils.hasText(path.isEmpty())) {
```
mp911de, line 320: This is still wrong as we want to check whether `path` is empty, not whether `isEmpty` returns an empty result.

PR #2467, opened 2022-12-01
Track channel and pattern subscriptions to avoid unintended unsubscribe in `ReactiveRedisMessageListenerContainer`

mp911de:


src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactivePubSubCommands.java
```diff
@@ -65,20 +83,157 @@ public Mono<Void> subscribe(ByteBuffer... channels) {
...
+
+			long l = SUBSCRIBERS.get(this);
+
+			if (l > 0) {
+				if (SUBSCRIBERS.compareAndSet(this, l, l - 1)) {
+					return l == 1; // return true if this was the last subscriber
+				}
+			}
+
+			return false;
```
christophstrobl, line 197: maybe just `return SUBSCRIBERS.getAndUpdate(this, cur -> cur > 0 ? cur - 1 : cur) == 1;`

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveRedisConnection.java
```diff
@@ -51,6 +51,8 @@ class LettuceReactiveRedisConnection implements ReactiveRedisConnection {
 	private final AsyncConnect<StatefulConnection<ByteBuffer, ByteBuffer>> dedicatedConnection;
 	private final AsyncConnect<StatefulRedisPubSubConnection<ByteBuffer, ByteBuffer>> pubSubConnection;
 
+	private final LettuceReactivePubSubCommands pubSub = new LettuceReactivePubSubCommands(this);
```
christophstrobl, line 54: Does it make sense to have this one as `Lazy<LettuceReactivePubSubCommands>`?

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveSubscription.java
```diff
@@ -182,8 +184,12 @@ static class State {
 		 */
 		Mono<Void> subscribe(ByteBuffer[] targets, Function<ByteBuffer[], Mono<Void>> subscribeFunction) {
 
-			return subscribeFunction.apply(targets).doOnSuccess((discard) -> this.targets.addAll(Arrays.asList(targets)))
-					.onErrorMap(exceptionTranslator);
+			return subscribeFunction.apply(targets).doOnSuccess((discard) -> {
+
+				for (ByteBuffer target : targets) {
```
christophstrobl, line 189: Maybe we could introduce a wrapper object that can do the wrapping of the `ByteBuffer` instead of adding to the `Set` directly.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveRedisConnection.java
mp911de: `Lazy` isn't synchronized. On the other hand, creating concurrent hash maps isn't ideal either. Maybe a `synchronized` access with lazy creation in the getter is a better way out.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveSubscription.java
mp911de: The entire buffer handling isn't ideal. I currently don't have a better idea, Buffers and byte arrays aren't great as map or set keys.

alexroode:
I am pretty new to the ecosystem so please forgive me if this is a dumb question, but would it be possible to bring these changes to a `2.x` release? We are affected by this issue and are unable to move to Spring Boot 3 anytime soon.

mp911de:
We're rather cautious when it comes to backporting enhancements. This change has the potential to break existing applications and as a measure to reduce any risk, we decided to not backport the change to the `2.7.x` development line. Let me take your request to our team to see what the rest of the team thinks

/cc @jxblum @christophstrobl

PR #2482, opened 2023-01-10
Use correct output for `ZCARD` command using `LettuceConnection.execute(…)`

### LINKED ISSUE #2473: `LettuceConnection.execute(ZCARD)` fails with unsupported output exception
when i invoke zcard command by this method, it print an error
image

image

i use Lettuce, my default codec is bytearraycodes

image
because the inteface dosn't defined that method, it define the flowing
image

 commandOutputTypeHint is null 
image

, when i run into that method, i found it auto locate the prototype and adjustment the result resolve, but i can't find the zcard map to the commondoutput class, so it use the default, and resolve the result cause an error

mp911de:
We now provide a type hint for the `ZCARD` command.

Closes #2473

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>3.1.0-SNAPSHOT</version>
+	<version>3.1.0-GH-2473-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 	<description>Spring Data module for Redis</description>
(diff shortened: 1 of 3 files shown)


jxblum:
Is this required for SD Redis `2.7.7` given the `TypeHints` are used for native??

mp911de:
Jedis uses a different mechanism to parse command outputs. Type hints are in the context of redis output types, not native image ones.

jxblum:
This has been merged to `main` (`3.1.0`), `3.0.x` and `2.7.x`.

PR #2491, opened 2023-01-18
Cyclic references resolve

m1c1b:

---

Hello! First of all I want to describe the problem that I have fixed.

Imagine a couple of `@RedisHash` entities that references one to anoter like A <->B. In that case StackOverflow will be thrown because of infinite calls stack for resolving `@Reference` fields on any kind of reading.

So I have made a clone of `MappingRedisConverter` with name `ReferenceMappingRedisConverter` but with one difference. `ReferenceMappingRedisConverter` has context of already resolved field branches, `ReferenceRedisAdapter` control the context of converter on reading operations.

I'm not really confident should I edit already existing classes or create new, this is why I have created new ones. I didn't know what version should I write in `@since` javadoc annotation, so I just didn't write it.

And of course I am ready to write documentation for my code and a sample configuration in spring boot, but I need a little bit explanation of how to do it.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
+++ b/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
@@ -86,6 +86,7 @@ import org.springframework.util.CollectionUtils;
  * @author Denis Zavedeev
  * @author ihaohong
  * @author Chen Li
+ * @author Ilya Viaznin
  * @author Vedran Pavic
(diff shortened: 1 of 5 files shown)


mp911de:
Thanks for your pull request and sorry for the long silence. It makes indeed sense to have cyclic guards in place. With the current state, it is impossible to review the changes because the converter is fully duplicated. Instead, it would make sense to carry around a contextual state object that holds references to resolved objects from hashes and have that as starting point.

PR #2525, opened 2023-03-08
Use BitSet instead of LinkedHashSet to improve the performance of SlotRange construction

daihuabin:
Here are the JMH test results for BitSet and LinkedHashSet
```
Benchmark              (slotSize)  (startFrom)   Mode  Cnt     Score     Error   Units
Scratch.bitSet                100          100  thrpt   10  6584.879 ± 178.558  ops/ms
Scratch.bitSet                100         3000  thrpt   10  5546.809 ± 216.817  ops/ms
Scratch.bitSet                100         6000  thrpt   10  4024.224 ± 111.127  ops/ms
Scratch.bitSet               3000          100  thrpt   10   197.832 ±  27.294  ops/ms
Scratch.bitSet               3000         3000  thrpt   10   201.136 ±  32.989  ops/ms
Scratch.bitSet               3000         6000  thrpt   10   203.528 ±  23.295  ops/ms
Scratch.bitSet               6000          100  thrpt   10   101.227 ±  14.600  ops/ms
Scratch.bitSet               6000         3000  thrpt   10   101.928 ±  19.558  ops/ms
Scratch.bitSet               6000         6000  thrpt   10   100.366 ±  17.370  ops/ms
Scratch.linkedHashSet         100          100  thrpt   10  1556.557 ±  95.599  ops/ms
Scratch.linkedHashSet         100         3000  thrpt   10  1553.883 ±  40.297  ops/ms
Scratch.linkedHashSet         100         6000  thrpt   10  1563.915 ±  29.968  ops/ms
Scratch.linkedHashSet        3000          100  thrpt   10    62.365 ±   2.632  ops/ms
Scratch.linkedHashSet        3000         3000  thrpt   10    61.362 ±   3.366  ops/ms
Scratch.linkedHashSet        3000         6000  thrpt   10    61.484 ±   4.668  ops/ms
Scratch.linkedHashSet        6000          100  thrpt   10    29.789 ±   1.172  ops/ms
Scratch.linkedHashSet        6000         3000  thrpt   10    29.868 ±   1.360  ops/ms
Scratch.linkedHashSet        6000         6000  thrpt   10    29.746 ±   1.843  ops/ms
```

Here is the JMH test code
```java
import org.openjdk.jmh.annotations.*;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;

import java.util.BitSet;
import java.util.LinkedHashSet;
import java.util.Set;
import java.util.concurrent.TimeUnit;

@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Thread)
@Fork(value = 1)
@Warmup(iterations = 1)
@Measurement(iterations = 10, time = 1)
@Threads(Threads.MAX)
public class Scratch {

    Set<Integer> slots;

    @Param({"100", "3000", "6000"})
    int slotSize;

    @Param({"100", "3000", "6000"})
    int startFrom;

    @Setup(Level.Iteration)
    public void init() {
        slots = new LinkedHashSet<>();
        for (int i = startFrom; i < startFrom + slotSize; i++) {
            slots.add(i);
        }
    }

    @Benchmark
    public void linkedHashSet() {
        LinkedHashSet<Integer> integers = new LinkedHashSet<>();
        for (Integer slot : slots) {
            integers.add(slot);
        }
    }

    @Benchmark
    public void bitSet() {
        BitSet set = new BitSet();
        for (Integer slot : slots) {
            set.set(slot);
        }
    }

    public static void main(String[] args) throws RunnerException {
        Options options = new OptionsBuilder().include(Scratch.class.getSimpleName()).build();
        new Runner(options).run();
    }
}
```

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/ClusterSlotHashUtil.java
+++ b/src/main/java/org/springframework/data/redis/connection/ClusterSlotHashUtil.java
@@ -28,7 +28,7 @@ import org.springframework.util.Assert;
  */
 public final class ClusterSlotHashUtil {
 
-	private static final int SLOT_COUNT = 16384;
+	public static final int SLOT_COUNT = 16384;
 
 	private static final byte SUBKEY_START = '{';
 	private static final byte SUBKEY_END = '}';
--- a/src/main/java/org/springframework/data/redis/connection/RedisClusterNode.java
+++ b/src/main/java/org/springframework/data/redis/connection/RedisClusterNode.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.redis.connection;
 
+import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
@@ -166,11 +168,12 @@ public class RedisClusterNode extends RedisNode {
 
 	/**
 	 * @author Christoph Strobl
+	 * @author daihuabin
 	 * @since 1.7
 	 */
 	public static class SlotRange {
 
-		private final Set<Integer> range;
+		private final BitSet range;
 
 		/**
 		 * @param lowerBound must not be {@literal null}.
@@ -181,19 +184,30 @@ public class RedisClusterNode extends RedisNode {
 			Assert.notNull(lowerBound, "LowerBound must not be null");
 			Assert.notNull(upperBound, "UpperBound must not be null");
 
-			this.range = new LinkedHashSet<>();
+			this.range = new BitSet(upperBound + 1);
 			for (int i = lowerBound; i <= upperBound; i++) {
-				this.range.add(i);
+				this.range.set(i);
 			}
 		}
 
 		public SlotRange(Collection<Integer> range) {
-			this.range = CollectionUtils.isEmpty(range) ? Collections.emptySet() : new LinkedHashSet<>(range);
+			if (CollectionUtils.isEmpty(range)) {
+				this.range = new BitSet(0);
+			} else {
+				this.range = new BitSet(ClusterSlotHashUtil.SLOT_COUNT);
+				for (Integer pos : range) {
+					this.range.set(pos);
+				}
+			}
+		}
+
+		public SlotRange(BitSet range) {
+			this.range = (BitSet) range.clone();
 		}
 
 		@Override
 		public String toString() {
-			return range.toString();
+			return Arrays.toString(this.getSlotsArray());
 		}
 
 		/**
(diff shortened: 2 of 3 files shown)


jxblum:
First, I will start by saying that there is little question a `BitSet` in this UC will be more performant than a `LinkedHashSet`, not only in terms of throughput, but also memory consumption, particularly since the purpose of `SlotRange` is simply  to track which "slots" a particular Redis cluster node is responsible for managing when the data set stored in Redis, across a "cluster", is sharded (partitioned).

Therefore, a "throughput" measurement for "adding slots" to the `Set` used to track which slots a Redis cluster node is managing (performed only once, on startup) is less interesting than measuring the "average (access) time" when determining whether a particular Redis cluster node is managing specific (hashed) slots, which would be a factor of the size of the Redis cluster, since not all nodes manage all slots, and would be necessary in certain key operations.

Therefore, I created another [Benchmark](https://github.com/jxblum/redis-experiments/blob/main/spring-data-redis-experiments/src/main/java/io/vmware/spring/data/redis/connection/RedisClusterNodeSlotRangeBenchmarks.java) to measure the "average time" of calling the [`RedisClusterNode.SlotRange.contains(slot)`](https://github.com/spring-projects/spring-data-redis/blob/3.0.1/src/main/java/org/springframework/data/redis/connection/RedisClusterNode.java#L199-L205) method, which is called by the [`RedisCluterNode.servesSlot(slot)`](https://github.com/spring-projects/spring-data-redis/blob/3.0.1/src/main/java/org/springframework/data/redis/connection/RedisClusterNode.java#L112-L118) method.

The `ReidsClusterNodes.servesSlot(..)` method is used when assessing the topology of the cluster, in SD Redis's `ClusterTopology` class. And, in particular, this is important in determining the master node managing the slot containing the key ([this](https://github.com/spring-projects/spring-data-redis/blob/3.0.1/src/main/java/org/springframework/data/redis/connection/RedisClusterNode.java#L112-L118)), which is essential during write operations to ensure a degree of consistency when the value of the key changes, most likely in a concurrent environment.

It is also used to determine all the nodes in the cluster that serve a particular slot ([this](https://github.com/spring-projects/spring-data-redis/blob/3.0.1/src/main/java/org/springframework/data/redis/connection/ClusterTopology.java#L107-L122)). Of course, this is useful during reads in a replicated, clustered environment in order to improve read performance (throughput, latency).

jxblum:
My finding are consistent with the expected outcome, that there is sizable improvement in performance when assessing whether a Redis cluster node is managing a slot:

Using `LinkedHashSet`:

```
Benchmark                                                         (iterations)  Mode  Cnt   Score    Error  Units
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot          2000  avgt    3   6.716 ± 23.909  us/op
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot          3000  avgt    3  10.528 ± 11.847  us/op
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot          5000  avgt    3  21.953 ± 64.360  us/op
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot         10000  avgt    3  58.908 ± 13.313  us/op
```

Using `BitSet`:

```
Benchmark                                                         (iterations)  Mode  Cnt  Score   Error  Units
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot          2000  avgt    3  0.899 ± 0.266  us/op
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot          3000  avgt    3  1.345 ± 0.041  us/op
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot          5000  avgt    3  2.327 ± 0.896  us/op
RedisClusterNodeSlotRangeBenchmarks.measureSlotRangeContainsSlot         10000  avgt    3  4.597 ± 4.279  us/op
```

jxblum:
Of course, using a `BitSet` should improve startup time of the Spring Data Redis application on average, and to a degree, dependent on environment factors, but again this is a function of the cluster size and number of configured slots.

Issue #2503, opened 2023-02-13
Migrate `RedisConnectionFactory` to `Lifecycle` beans


mp911de:
Based on https://github.com/spring-projects/spring-framework/issues/29921, we could move our connection factories towards beans that implement the lifecycle. This would be especially useful to decouple `afterPropertiesSet` from the connection pool startup and `dispose()` where we dispose resources. With `start()` and `stop()` connection factories could start up pools and tear down as necessary without destroying the entire bean.

jxblum:
I'd need to think on this a bit more, but my first reaction when I read your comment (@mp911de) is that this might also be beneficial when using Spring Data Redis with Project Loom (Threading model).

PR #2554, opened 2023-04-14
Reset JedisConnectionFactory's destroyed flag when initializing bean

### LINKED ISSUE #2553: JedisConnectionFactory failed to get connection after configuration auto update
Error Trace:
> java.lang.IllegalStateException: JedisConnectionFactory was destroyed and cannot be used anymore##  at org.springframework.util.Assert.state(Assert.java:76) ~[spring-core-5.3.23.jar!/:5.3.23]##  at org.springframework.data.redis.connection.jedis.JedisConnectionFactory.assertInitialized(JedisConnectionFactory.java:971) ~[spring-data-redis-2.7.3.jar!/:2.7.3]##  at org.springframework.data.redis.connection.jedis.JedisConnectionFactory.getConnection(JedisConnectionFactory.java:509) ~[spring-data-redis-2.7.3.jar!/:2.7.3]##  at org.springframework.data.redis.core.RedisConnectionUtils.fetchConnection(RedisConnectionUtils.java:193) ~[spring-data-redis-2.7.3.jar!/:2.7.3]##  at org.springframework.data.redis.core.RedisConnectionUtils.doGetConnection(RedisConnectionUtils.java:144)

We receive `EnvironmentChangeEvent` in the configuration hot update scenario. 
The `ConfigurationPropertiesRebinder` will destroy and reinitialize `JedisConnectionFactory`.

image

In destroy method we set the destroyed flag as true.
image

But in initial method (afterPropertiesSet) we don't reset destroyed as false.
image

This will cause us to be unable to get a connection from the `JedisConnectionFactory` any more.
image

Jiabao-Sun:
Reset JedisConnectionFactory's destroyed flag when initializing bean to prevent configuration hot upgrade issue.

Closes #2553

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
@@ -305,6 +305,7 @@ public class JedisConnectionFactory implements InitializingBean, DisposableBean,
 		}
 
 		this.initialized = true;
+		this.destroyed = false;
 	}
 
 	JedisClientConfig createSentinelClientConfig(SentinelConfiguration sentinelConfiguration) {


mp911de:
We should revisit initialization in the scope of #2503

jxblum:
@Jiabao-Sun  - I agree with @mp911de here. This needs to be considered carefully, and will likely and necessarily be tied to the Spring containers [lifecycle management capabilities](https://docs.spring.io/spring-framework/reference/core/beans/factory-nature.html#beans-factory-lifecycle) in support of our [CRaC](https://openjdk.org/projects/crac/) efforts.

I also don't necessarily think or agree that allowing the `JedisConnectionFactory` bean to be reused after it has already been destroyed is necessarily the correct thing to do, in any case. After all, this [assertion](https://github.com/spring-projects/spring-data-redis/blob/3.1.0/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java#L870) has been in place for sometime and while there is not a documented reason behind it, it is not seemingly arbitrary either.

Additionally, whatever we do with the Jedis based connection support should be consistently applied to the Lettuce connection support as well.

While this may look safe on the surface that is Spring Data (Redis), you must also consider what occurs, or state that is held in the drivers.

Unfortunately, I would need to think on and test this with several arrangements as well as different contexts (e.g. Spring Boot application), to determine whether this would even be safe to do in a majority of the cases.  There is a good/strong reason why the core Spring Framework team rejected this [ticket](https://github.com/spring-projects/spring-framework/issues/28001) of mine many moons ago, which is similar in concept.

Issue #2261, opened 2022-02-15
StreamMessageListenerContainer should wait until listeners finish processing within timeout during shutdown


lrozek:
when spring boot application is being shutdown gracefully, DefaultStreamMessageListenerContainer should wait until listeners have finished their processing within given timeout. Currently redis stream listeners are terminated instantly in the middle of the listener handler without giving them a chance to finish

Similar functionality is offered for RabbitMQ listeners (https://github.com/spring-projects/spring-amqp/blob/main/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.java#L634) and JMS listeners (https://www.amitph.com/spring-boot-graceful-shutdown/#JMS_Listener)

repro project can be found here https://github.com/lrozek/redis-listener-graceful-shutdown

mp911de:
Thanks for raising the topic. It makes sense to await completion in the `stop` method to avoid any races or connection errors because the connection is closed while still in use.

lrozek:
@mp911de the point was rather to allow graceful shutdown of redis listener to finish its processing instead of terminating abruptly, which could lead to data being lost or corrupted.

Races and connections errors that you mentioned I also encountered during shutdown, however I intended to report it as a separate issue.

MrNocTV:
Hi, I have the same problem with @lrozek that also posted here on stackoverflow. https://stackoverflow.com/questions/74003013/spring-boot-web-socket-message-broker-shutdown-before-redis-stream

It's kinda similar. The application needs to process all remaining messages and only then terminated.
Cant find any workaround at the moment :( 
With this bug I don't think spring-data-redis is production ready.

aooohan:
@mp911de Hi, I have tried to fix this issue. Please review #2563, let me if any improvement need do.

tubbynl:
not really a race condition because it always happens on shutdown, the racing part perhaps is that the connections are closed before the `StreamMessageListenerContainer` 

only "solution" currently is to disable the logger to prevent the exceptions spamming the logs / alerting.
```
logging.level.org.springframework.data.redis.stream=off
```
~perhaps a half-way solution would be to let `Subscription` implement `Lifecycle` and close it if he's the last listener.~
scratch that, he is not aware if it's container

localstorageserver:
@mp911de

Following up on #2261 - I see PR #3040 moved away from busy-spin to CountDownLatch, but it's still marked "pending-design-work."

[...]

PR #2563, opened 2023-04-20
`StreamMessageListenerContainer` should wait until listeners finish processing within timeout during shutdown

aooohan:
Fix issue: #2261 

Set a new option `shutdownTimeout` for container option.   The container will wait listener with `shutdownTimeout` time to finish processing.

If any improvement can do, feel free to tell me.

### DIFF
--- a/src/main/java/org/springframework/data/redis/stream/DefaultStreamMessageListenerContainer.java
+++ b/src/main/java/org/springframework/data/redis/stream/DefaultStreamMessageListenerContainer.java
@@ -19,6 +19,10 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Executor;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
 
 import org.apache.commons.logging.Log;
@@ -51,6 +55,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Mark Paluch
  * @author Christoph Strobl
+ * @author Han Li
  * @since 2.2
  */
 class DefaultStreamMessageListenerContainer<K, V extends Record<K, ?>> implements StreamMessageListenerContainer<K, V> {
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/redis/stream/StreamMessageListenerContainer.java
```diff
@@ -627,6 +639,21 @@ public StreamMessageListenerContainerOptionsBuilder<K, V> pollTimeout(Duration p
...
+		/**
+		 * Configure a timeout for shutdown container.
+		 *
+		 * @param shutdownTimeout must not be {@literal null} or negative.
+		 * @return {@code this} {@link StreamMessageListenerContainerOptionsBuilder}.
+		 */
+		public StreamMessageListenerContainerOptionsBuilder<K, V> shutdownTimeout(Duration shutdownTimeout) {
+
+			Assert.notNull(shutdownTimeout, "Shutdown timeout must not be null");
+			Assert.isTrue(!shutdownTimeout.isNegative(), "Shutdown timeout must not be negative");
```
vuhp, line 651: The assertion condition should be `shutdownTimeout. isNegative()` instead of `!shutdownTimeout. isNegative()`.

aooohan: Thank you for the reviewing. But here is correct, if you replace it with ` shutdownTimeout. isNegative() ` current assertion, that means you're asking for a negative `shutdownTime` parameter.

vuhp: Sorry, I was confused with the negative of negative

src/main/java/org/springframework/data/redis/stream/DefaultStreamMessageListenerContainer.java
```diff
@@ -160,9 +165,22 @@ public void stop() {
 		synchronized (lifecycleMonitor) {
 
 			if (this.running) {
-
-				subscriptions.forEach(Cancelable::cancel);
-
+				subscriptions.stream()
```
mp911de, line 168: Running a busy-spin cannot be interrupted. Also, keeping common ForkJoin pool threads busy asks for trouble.

Instead, our Tasks should report a completion future that we can listen to.

Issue #1433, opened 2018-08-03
Add `TtlFunction` to `RedisCacheConfiguration` for dynamic Time to Live


spring-projects-issues:
**[Kyle Cronin](https://jira.spring.io/secure/ViewProfile.jspa?name=kylecronin@adp)** opened **[DATAREDIS-858](https://jira.spring.io/browse/DATAREDIS-858?redirect=false)** and commented

`RedisCache` should allow sub classes to define the TTL duration based on the cache key or value for use cases where the TTL is not know at configuration time or could change depending on the value being stored. For example access tokens where the token provider specifies how long the token is valid for.

 

```java
public class RedisCache {
   // default impementation delegates to cache config
   @NonNull protected Duration getTtl(Object key, Object value) { 
      return cacheConfig.getTtl(); 
   }
}

public class AccessTokenRedisCache extends RedisCache {
   @Override
   @NonNull
   protected Duration getTtl(Object key, Object value) {
      if (value instanceof AccessToken) {
         return ((AccessToken) value).getTtl();
      } else {
         return super.getTtl(key, value);
      }
   }
}
```

If there was a way to specify a TTL policy in the RedisCacheConfiguration that would be bonus

---
No further details from [DATAREDIS-858](https://jira.spring.io/browse/DATAREDIS-858?redirect=false)

nicdard:
Hi, is this issue still active? I have multiple use cases were I would like to randomise the TTL at each put operation, and I would like to rise a PR for this :)

PR #2587, opened 2023-05-24
Allow to dynamically provide TTL configuration

Koooooo-7:
Fix issue: #1433 .

Give users a chance to change the `ttl` dynamically and give them the flexibility to do so.


mp911de (changes requested):
Using a context-less supplier doesn't add much value here. I could well think of a case, where the TTL could be computed from the actual object. 

How about a design of a TTL function (`Function<Object, Duration>`) where the object argument is provided by the actual cache value?

Koooooo-7:
Hi @mp911de , I agree with that we should provide some context for user to get the `ttl` and I did the updates.
But I have something should mention about it.
1. For now I provide the original `key` and `value` to users, is it okay? or we should provide the key after `createAndConvertCacheKey`, so does the `value`.
2. About the `public Duration getTtl() {}` for backward compatibility， I user the `Object.class` as the default key instead.

mp911de:

Issue #2602, opened 2023-06-13
Introduce `ReversedRedisListView` for Java 21 compatibility


jxblum:
Builds Spring Data Redis on JDK 21 (EA).

jxblum:
Seems the `List` interface and `Deque` interface in Java 21 have incompatible APIs in Java 21. Both interfaces defined `reversed()` method.

However, in the `List` interface, [`reversed()`](https://download.java.net/java/early_access/jdk21/docs/api/java.base/java/util/List.html#reversed()) returns a `List<E>` and in `Deque`, [`reversed()`](https://download.java.net/java/early_access/jdk21/docs/api/java.base/java/util/Deque.html#reversed()) returns a `Deque<E>`.

This means types like [`RedisList`](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/support/collections/RedisList.java#L40) cannot implement both `List` and `Deque`.

jxblum:
In addition, `RedisList` must also override and provide default implementations for the following `List` / `Deque` interface methods:

* `addFirst(:E)`
* `addLast(:E)`
* `getFirst():E`
* `getLast():E`
* `removeFirst():E`
* `removeLast():E`

PR #2608, opened 2023-06-14
Introduce `ReversedRedisListView` for Java 21 compatibility

jxblum:
See #2602 

Do not merge yet.


src/main/java/org/springframework/data/redis/support/collections/DefaultRedisList.java
```diff
@@ -592,4 +593,15 @@ private void cap() {
 			listOps.trim(0, maxSize - 1);
 		}
 	}
+
+	@Override
+	public RedisList<E> reversed() {
+
+		DefaultRedisList<E> redisList = new DefaultRedisList<>(getKey(), getOperations(), this.maxSize);
+
+		StreamSupport.stream(this.spliterator(), false)
```
mp911de, line 602: We should honor that `reversed` is a view of the underlying list. Here, we duplicate the contents of the list by inserting everything that is stored at `getKey()`.

I suggest introducing a package-private `ReversedRedisList` type where we delegate to `this`. As our baseline remains Java 17, I think we will have to set up our own implementation without being able to reuse much from the Java libraries.

Looking at the JDK libs, their classes are delegations.

src/main/java/org/springframework/data/redis/support/collections/RedisList.java
```diff
@@ -229,4 +229,39 @@ default E moveLastTo(RedisList<E> destination, Direction destinationPosition, Du
...
+	}
+
+	@Override
+	default E removeLast() {
+		return List.super.removeLast();
+	}
+
+	@Override
+	default RedisList<E> reversed() {
+		throw new UnsupportedOperationException("Not Implemented");
```
mp911de, line 265: Let's turn that into a proper implementation using a package-private delegation type. Users looking at the code and finding `UOE` will have a hard time understanding why this is and whether `reversed` is supported at all.

src/main/java/org/springframework/data/redis/support/collections/DefaultRedisList.java
jxblum: Yeah, that was intentional and temporary hack. I will change this into a proper, view-based abstraction.

src/main/java/org/springframework/data/redis/support/collections/RedisList.java
jxblum: Well, methods  [ `addFirst(:E)`, 'addLast(:E)`, `getFirst()`, `getLast()`, `removeFirst(:E)`, `removeLast(:E)` ] must be redefined in `RedisList<E>` since `RedisList`, unlike Java's LinkedList (a "class"), is an "interface.

And, since the declarations/definitions of these methods differ between Java's `List` interface and the `Deque` interface, where `List` provides default implementations, and `Deque` does not, leads to the following compilation error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (java-compile) on project spring-data-redis: Compilation failure
[ERROR] /Users/jblum/pivdev/spring-data-redis/src/main/java/org/springframework/data/redis/support/collections/RedisList.java:[40,7] error: types List<E#1> and BlockingDeque<E#2> are incompatible;
[ERROR]   interface RedisList<E#3> inherits abstract and default for addFirst(E#3) from types List and BlockingDeque
[ERROR]   where E#1,E#2,E#3 are type-variables:
[ERROR]     E#1 extends Object declared in interface List
[ERROR]     E#2 extends Object declared in interface BlockingDeque
[ERROR]     E#3 extends Object declared in interface RedisList
```

jxblum: `reversed()`on the other hand, can indeed be encapsulated in a proper type with a supporting implementation.

src/main/java/org/springframework/data/redis/support/collections/DefaultRedisList.java
```diff
@@ -592,4 +582,122 @@ private void cap() {
...
+
+		@Override
+
+		public void set(E element) {
+
+			Assert.state(this.lastReturnedElement != null,
+				"Next or previous must be called before set");
+
+			try {
+				DefaultRedisList.this.set(indexOf(this.lastReturnedElement), element);
```
mp911de, line 697: Let's resort to using the cursor index. Using the index of the last returned element can hit an equal object in the list (e.g. if the list contains `[a a a b c c c]` and the last returned element is `c`).

src/main/java/org/springframework/data/redis/support/collections/ReversedRedisListView.java
```diff
@@ -0,0 +1,738 @@
...
+	public boolean add(E element) {
+		this.base.add(0, element);
+		return true;
+	}
+
+	@Override
+	public boolean addAll(Collection<? extends E> collection) {
+
+		@SuppressWarnings("unchecked")
+		E[] adds = (E[]) collection.toArray();
```
mp911de, line 178: We could have the `toArray` conversion even after the `isEmpty()` check

mp911de (approved):
This looks pretty decent. I left a few thoughts, other than that, I will go ahead and merge this one for Friday.

jxblum:
> This looks pretty decent. I left a few thoughts, other than that, I will go ahead and merge this one for Friday.

FYI, I am still in favor of removing all this (copy/pasted) code after we move to a Java baseline >= 21, with preference for the provided Java library & API facilities. I had this working reliably on Java 21 in this [commit](https://github.com/spring-projects/spring-data-redis/pull/2608/commits/4c9e5912d409d96e1ca62d45e3e8c4a6b0b8109e), and I am not interested in maintaining their code.

jxblum:
> > This looks pretty decent. I left a few thoughts, other than that, I will go ahead and merge this one for Friday.
> 
> FYI, I am still in favor of removing all this (copy/pasted) code after we move to a Java baseline >= 21, with preference for the provided Java library & API facilities. I had this working reliably on Java 21 in this [commit](https://github.com/spring-projects/spring-data-redis/pull/2608/commits/4c9e5912d409d96e1ca62d45e3e8c4a6b0b8109e), and I am not interested in maintaining their code. As such, I see this only as a temporary solution.
> 
> Additionally, though I did not try it, _Google Guava_ would provide what we need in this case (for example: [here](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Lists.html#reverse(java.util.List))), even though it would add a library dependency.

Once more, I have made some modifications based on your feedback. SD Redis still builds reliably on Java 17 and Java 21.

mp911de:

Issue #2351, opened 2022-06-28
Add support for GETEX in Spring Redis Cache


danibs:
Hi,
in version > 2.6:
https://docs.spring.io/spring-data/data-redis/docs/current/reference/html/#new-in-2.6.0

you support [GETEX](https://redis.io/commands/getex/) command that permits (I think) to have a kind of "time to idle" like in EhCache (`ExpiryPolicyBuilder.timeToIdleExpiration`) where the key doesn't expire if someone get its value.

I updated my `pom.xml` with:
```xml
<dependency>
    <groupId>org.springframework.data</groupId>
    <artifactId>spring-data-redis</artifactId>
    <version>2.7.1</version>
</dependency>
```

but in `RedisCacheConfiguration` class there is only `Duration ttl` attribute.
I found this https://dzone.com/articles/how-to-boost-redis-with-local-caching-in-java but it talks about _Redisson_ and I found in library that is possible to have **timeToLive** and **Idle**:
```java
LocalCachedMapOptions.defaults()
   .evictionPolicy(EvictionPolicy.LFU)
   .timeToLive(48, TimeUnit.MINUTES)
   .maxIdle(24, TimeUnit.MINUTES);
```

Do I miss something?
Thanks

jxblum:
@bytestreme - Before discussing and reviewing any [changes](https://github.com/spring-projects/spring-data-redis/pull/2643) to Spring Data Redis, I want to clarify a few things.

Your (implied) recommendation is to support cache entry, idle-timeout expiration (TTI) using the Redis [GETEX](https://redis.io/commands/getex/) command (as evident in your [PR](https://github.com/spring-projects/spring-data-redis/pull/2643)) in addition to cache entry, time-to-live expiration (TTL) currently used by Spring Data Redis's [implementation](https://docs.spring.io/spring-data/redis/docs/current/reference/html/#redis:support:cache-abstraction) of Spring Framework's [Cache Abstraction](https://docs.spring.io/spring-framework/reference/integration/cache/strategies.html).

[...]

PR #2643, opened 2023-07-16
Add support for GETEX in Spring Redis Cache

bytestreme:

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -44,6 +44,7 @@ import org.springframework.util.Assert;
  * @author Christoph Strobl
  * @author Mark Paluch
  * @author André Prata
+ * @author Shyngys Sapraliyev
  * @since 2.0
  */
 class DefaultRedisCacheWriter implements RedisCacheWriter {
@@ -93,7 +94,7 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 	}
 
 	@Override
-	public void put(String name, byte[] key, byte[] value, @Nullable Duration ttl) {
+	public void put(String name, byte[] key, byte[] value, @Nullable Duration ttl, @Nullable Duration maxIdle) {
 
 		Assert.notNull(name, "Name must not be null!");
 		Assert.notNull(key, "Key must not be null!");
@@ -104,7 +105,11 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 			if (shouldExpireWithin(ttl)) {
 				connection.set(key, value, Expiration.from(ttl.toMillis(), TimeUnit.MILLISECONDS), SetOption.upsert());
 			} else {
-				connection.set(key, value);
+				if (shouldExpireMaxIdleWithin(maxIdle)) {
+					connection.set(key, value, Expiration.from(maxIdle), SetOption.upsert());
+				} else {
+					connection.set(key, value);
+				}
 			}
 
 			return "OK";
@@ -114,12 +119,18 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 	}
 
 	@Override
-	public byte[] get(String name, byte[] key) {
+	public byte[] get(String name, byte[] key, @Nullable Duration maxIdle) {
 
 		Assert.notNull(name, "Name must not be null!");
 		Assert.notNull(key, "Key must not be null!");
 
-		byte[] result = execute(name, connection -> connection.get(key));
+		byte[] result;
+
+		if (shouldExpireMaxIdleWithin(maxIdle)) {
+			result = execute(name, connection -> connection.getEx(key, Expiration.from(maxIdle)));
+		} else {
+			result = execute(name, connection -> connection.get(key));
+		}
 
 		statistics.incGets(name);
 
@@ -133,7 +144,7 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 	}
 
 	@Override
-	public byte[] putIfAbsent(String name, byte[] key, byte[] value, @Nullable Duration ttl) {
+	public byte[] putIfAbsent(String name, byte[] key, byte[] value, @Nullable Duration ttl, @Nullable Duration maxIdle) {
 
 		Assert.notNull(name, "Name must not be null!");
 		Assert.notNull(key, "Key must not be null!");
@@ -152,7 +163,11 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 				if (shouldExpireWithin(ttl)) {
 					put = connection.set(key, value, Expiration.from(ttl), SetOption.ifAbsent());
 				} else {
-					put = connection.setNX(key, value);
+					if (shouldExpireMaxIdleWithin(maxIdle)) {
+						put = connection.set(key, value, Expiration.from(maxIdle), SetOption.ifAbsent());
+					} else {
+						put = connection.setNX(key, value);
+					}
 				}
 
 				if (put) {
@@ -316,6 +331,10 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 
 	private static boolean shouldExpireWithin(@Nullable Duration ttl) {
 		return ttl != null && !ttl.isZero() && !ttl.isNegative();
+	}
+
+	private static boolean shouldExpireMaxIdleWithin(@Nullable Duration maxIdle) {
+		return maxIdle != null && !maxIdle.isZero() && !maxIdle.isNegative();
 	}
 
 	private static byte[] createCacheLockKey(String name) {
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -45,6 +45,7 @@ import org.springframework.util.ReflectionUtils;
  * @author Christoph Strobl
  * @author Mark Paluch
  * @author Piotr Mionskowski
+ * @author Shyngys Sapraliyev
  * @see RedisCacheConfiguration
  * @see RedisCacheWriter
  * @since 2.0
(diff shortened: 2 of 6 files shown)


bytestreme:
resolves https://github.com/spring-projects/spring-data-redis/issues/2351

bytestreme:
@mp911de

mp911de:
@jxblum can you have a look as well? Introducing an expire after idle (Setting the TTL on GET) makes sense as we have already support for `GETEX`. I just wonder about the configuration side, whether we want to enable our TTL function for idle or whether we want to introduce another function to determine the Idle time for a key when getting the value. WDYT?

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -93,7 +94,7 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 	}
 
 	@Override
-	public void put(String name, byte[] key, byte[] value, @Nullable Duration ttl) {
+	public void put(String name, byte[] key, byte[] value, @Nullable Duration ttl, @Nullable Duration maxIdle) {
```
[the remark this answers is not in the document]
jxblum: Additionally, "`maxIdle`" is a **misnomer** and does not accurately reflect the behavior of "idle" data access and _idle-timeout expiration_ (TTL) policies in general.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -104,7 +105,11 @@ public void put(String name, byte[] key, byte[] value, @Nullable Duration ttl) {
 			if (shouldExpireWithin(ttl)) {
 				connection.set(key, value, Expiration.from(ttl.toMillis(), TimeUnit.MILLISECONDS), SetOption.upsert());
 			} else {
-				connection.set(key, value);
+				if (shouldExpireMaxIdleWithin(maxIdle)) {
+					connection.set(key, value, Expiration.from(maxIdle), SetOption.upsert());
+				} else {
+					connection.set(key, value);
+				}
```
jxblum, line 112: Expiration (and specifically, TTL) is something that is already [handled](https://github.com/spring-projects/spring-data-redis/blob/3.0.8/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java#L104-L105) as of Spring Data Redis `3.0.x` (and beyond: `3.1.x`, and currently, `3.2.x`).

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -152,7 +163,11 @@ public byte[] putIfAbsent(String name, byte[] key, byte[] value, @Nullable Durat
 				if (shouldExpireWithin(ttl)) {
 					put = connection.set(key, value, Expiration.from(ttl), SetOption.ifAbsent());
 				} else {
-					put = connection.setNX(key, value);
+					if (shouldExpireMaxIdleWithin(maxIdle)) {
+						put = connection.set(key, value, Expiration.from(maxIdle), SetOption.ifAbsent());
+					} else {
+						put = connection.setNX(key, value);
+					}
```
jxblum, line 170: Again, handled in Spring Data Redis `3.0.x`, `3.1.x` (most recent GA version) and `3.2.x` (current development version).

jxblum, line 170: What happens when TTL is less than TTI? The lesser of TTL and TTI should take precedence.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum, line 112: Expiration (and specifically, TTL) is something that is already [handled](https://github.com/spring-projects/spring-data-redis/blob/3.0.8/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java#L104-L105) as of Spring Data Redis `2.7.x` (and beyond: `3.0.x`, `3.1.x`, and currently, `3.2.x`), as is apparent from line 105-107, above.

src/main/java/org/springframework/data/redis/cache/RedisCache.java
```diff
@@ -82,7 +83,7 @@ protected RedisCache(String name, RedisCacheWriter cacheWriter, RedisCacheConfig
 	@Override
 	protected Object lookup(Object key) {
 
-		byte[] value = cacheWriter.get(name, createAndConvertCacheKey(key));
+		byte[] value = cacheWriter.get(name, createAndConvertCacheKey(key), cacheConfig.getMaxIdle());
```
jxblum, line 86: Again, the lesser of TTL and TTI should apply.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -133,7 +144,7 @@ public byte[] get(String name, byte[] key) {
 	}
 
 	@Override
-	public byte[] putIfAbsent(String name, byte[] key, byte[] value, @Nullable Duration ttl) {
+	public byte[] putIfAbsent(String name, byte[] key, byte[] value, @Nullable Duration ttl, @Nullable Duration maxIdle) {
```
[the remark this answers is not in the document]
jxblum: I don't think this signature change makes any sense in any of the changed `RedisCacheWriter` methods: `get(..)`, `put(..)` or `putIfAbsent(..)`; see [below](src/main/java/org/springframework/data/redis/cache/RedisCacheWriter.java).

I'd prefer, if we decide to add any option to configure TTI (perhaps overriding TTL when TTI < TTL) per cache entry by key at all, then the configuration would be included in the `RedisCacheConfiguration` per cache (not unlike TTL) and the `RedisCacheConfiguration` object is passed to the components of the cache implementation as needed to alter the caching behavior. Otherwise with each new option, the signature will only grow.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum, line 112: As @mp911de alluded to above, there is technically only 1 expiration policy in Redis for a given key: TTL.

Expiration (and specifically, TTL) is something that is already [handled](https://github.com/spring-projects/spring-data-redis/blob/3.0.8/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java#L104-L105) in Spring Data Redis as of `2.7.x` (and beyond: `3.0.x`, `3.1.x`, and currently, `3.2.x`), which is apparent in lines 105-107 shown above.

**Plus, what happens when TTL is less than TTI?** The lesser of TTL and TTI should take precedence.

Also note, as of Spring Data Redis `3.2.x` (currently in development with the latest version being `3.2.0-M1`) TTL expiration can even be computed dynamically, per cache entry by key (and value) using the new [TtlFunction](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M1/src/main/java/org/springframework/data/redis/cache/RedisCacheWriter.java#L174-L223) interface.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -114,12 +119,18 @@ public void put(String name, byte[] key, byte[] value, @Nullable Duration ttl) {
...
+	public byte[] get(String name, byte[] key, @Nullable Duration maxIdle) {
 
 		Assert.notNull(name, "Name must not be null!");
 		Assert.notNull(key, "Key must not be null!");
 
-		byte[] result = execute(name, connection -> connection.get(key));
+		byte[] result;
+
+		if (shouldExpireMaxIdleWithin(maxIdle)) {
+			result = execute(name, connection -> connection.getEx(key, Expiration.from(maxIdle)));
```
jxblum, line 130: Again, the lesser of TTL and TTI needs to be taken into consideration.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -152,7 +163,11 @@ public byte[] putIfAbsent(String name, byte[] key, byte[] value, @Nullable Durat
 				if (shouldExpireWithin(ttl)) {
 					put = connection.set(key, value, Expiration.from(ttl), SetOption.ifAbsent());
 				} else {
-					put = connection.setNX(key, value);
+					if (shouldExpireMaxIdleWithin(maxIdle)) {
+						put = connection.set(key, value, Expiration.from(maxIdle), SetOption.ifAbsent());
```
jxblum, line 167: The lesser of TTL and TTI should take precedence.

jxblum:
@bytestreme & @mp911de -

First, please see my comments in Issue #2351, [here](https://github.com/spring-projects/spring-data-redis/issues/2351#issuecomment-1642697384), specifically [here](https://github.com/spring-projects/spring-data-redis/issues/2351#issuecomment-1642740961), and in conclusion, [here](https://github.com/spring-projects/spring-data-redis/issues/2351#issuecomment-1642782243).

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum: Technically, TTI expiration policies (e.g. timeouts) have no bearing on cache entry creates or updates. TTL expiration is only updated when another create or update occurs after the initial create.

This would apply to `RedisCacheWriter.putIfAbsent(..)` as well.

jxblum:
@mp911de - I agree. Adding an expiration on `Cache.get(key)` in the context of (Spring Data) Redis's cache implementation makes sense.

However, I would caution that Redis really does not technically have a proper notion of cache entry, _idle-timeout expiration_ (TTI). It is essentially only TTL, or effectively TTI when `GETEX` along with `SET` including expiration options are used "consistently" across the application system architecture, and specifically, when caching.

It seems to me, frameworks like Redisson offer TTI as an additional feature implemented and supported by the framework itself outside of Redis. Without the use of `OBJECT IDLETIME` and an external expiration monitor, there is really no TTI, and juggling TTL with TTI could get messy and confusing for users.

jxblum:
See PR #2649.

Issue #2494, opened 2023-01-24
Make RedisAccessor abstract


jxblum:
`RedisAccessor` was intended to be an abstract base class for implementations requiring a connection to a Redis server. It was not intended to be used directly.

PR #2645, opened 2023-07-17
Make RedisAccessor abstract

jxblum:
See #2494


mp911de:

PR #2649, opened 2023-07-21
Support TTI expiration in Redis Cache implementation

jxblum:
See issue #2351.

See original pull request #2643.

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -124,12 +124,14 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 	}
 
 	@Override
-	public byte[] get(String name, byte[] key) {
+	public byte[] get(String name, byte[] key, @Nullable Duration ttl) {
 
 		Assert.notNull(name, "Name must not be null");
 		Assert.notNull(key, "Key must not be null");
 
-		byte[] result = execute(name, connection -> connection.get(key));
+		byte[] result = shouldExpireWithin(ttl)
+			? execute(name, connection -> connection.getEx(key, Expiration.from(ttl)))
+			: execute(name, connection -> connection.get(key));
 
 		statistics.incGets(name);
 
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -124,6 +124,11 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 	}
 
 	@Override
+	public byte[] get(String name, byte[] key) {
+		return get(name, key, null);
+	}
+
+	@Override
 	public byte[] get(String name, byte[] key, @Nullable Duration ttl) {
 
 		Assert.notNull(name, "Name must not be null");
--- a/src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
@@ -47,11 +47,11 @@ import org.springframework.util.Assert;
 public class RedisCacheConfiguration {
 
 	protected static final boolean DEFAULT_CACHE_NULL_VALUES = true;
-	protected static final boolean DEFAULT_ENABLE_TTI_EXPIRATION = false;
+	protected static final boolean DEFAULT_ENABLE_TIME_TO_IDLE_EXPIRATION = false;
 	protected static final boolean DEFAULT_USE_PREFIX = true;
 	protected static final boolean DO_NOT_CACHE_NULL_VALUES = false;
 	protected static final boolean DO_NOT_USE_PREFIX = false;
-	protected static final boolean ENABLE_IDLE_TIME_EXPIRATION = true;
+	protected static final boolean USE_TIME_TO_IDLE_EXPIRATION = true;
 
 	/**
 	 * Default {@link RedisCacheConfiguration} using the following:
@@ -112,7 +112,7 @@ public class RedisCacheConfiguration {
 
 		return new RedisCacheConfiguration(TtlFunction.persistent(),
 				DEFAULT_CACHE_NULL_VALUES,
-				DEFAULT_ENABLE_TTI_EXPIRATION,
+				DEFAULT_ENABLE_TIME_TO_IDLE_EXPIRATION,
 				DEFAULT_USE_PREFIX,
 				CacheKeyPrefix.simple(),
 				SerializationPair.fromSerializer(RedisSerializer.string()),
@@ -121,7 +121,7 @@ public class RedisCacheConfiguration {
 	}
 
 	private final boolean cacheNullValues;
-	private final boolean enableTtiExpiration;
+	private final boolean enableTimeToIdle;
 	private final boolean usePrefix;
 
 	private final CacheKeyPrefix keyPrefix;
@@ -134,13 +134,13 @@ public class RedisCacheConfiguration {
 	private final TtlFunction ttlFunction;
 
 	@SuppressWarnings("unchecked")
-	private RedisCacheConfiguration(TtlFunction ttlFunction, Boolean cacheNullValues, Boolean enableTtiExpiration,
+	private RedisCacheConfiguration(TtlFunction ttlFunction, Boolean cacheNullValues, Boolean enableTimeToIdle,
 			Boolean usePrefix, CacheKeyPrefix keyPrefix, SerializationPair<String> keySerializationPair,
 			SerializationPair<?> valueSerializationPair, ConversionService conversionService) {
 
 		this.ttlFunction = ttlFunction;
 		this.cacheNullValues = cacheNullValues;
-		this.enableTtiExpiration = enableTtiExpiration;
+		this.enableTimeToIdle = enableTimeToIdle;
 		this.usePrefix = usePrefix;
 		this.keyPrefix = keyPrefix;
 		this.keySerializationPair = keySerializationPair;
@@ -175,7 +175,7 @@ public class RedisCacheConfiguration {
 
 		Assert.notNull(cacheKeyPrefix, "Function used to compute prefix must not be null");
 
-		return new RedisCacheConfiguration(getTtlFunction(), getAllowCacheNullValues(), isTtiExpirationEnabled(),
+		return new RedisCacheConfiguration(getTtlFunction(), getAllowCacheNullValues(), isTimeToIdleEnabled(),
 				DEFAULT_USE_PREFIX, cacheKeyPrefix, getKeySerializationPair(), getValueSerializationPair(),
 				getConversionService());
 	}
@@ -189,7 +189,7 @@ public class RedisCacheConfiguration {
 	 * @return new {@link RedisCacheConfiguration}.
 	 */
 	public RedisCacheConfiguration disableCachingNullValues() {
-		return new RedisCacheConfiguration(getTtlFunction(), DO_NOT_CACHE_NULL_VALUES, isTtiExpirationEnabled(),
+		return new RedisCacheConfiguration(getTtlFunction(), DO_NOT_CACHE_NULL_VALUES, isTimeToIdleEnabled(),
 				usePrefix(), getKeyPrefix(), getKeySerializationPair(), getValueSerializationPair(),
 				getConversionService());
 	}
@@ -202,7 +202,7 @@ public class RedisCacheConfiguration {
 	 * @return new {@link RedisCacheConfiguration}.
 	 */
 	public RedisCacheConfiguration disableKeyPrefix() {
-		return new RedisCacheConfiguration(getTtlFunction(), getAllowCacheNullValues(), isTtiExpirationEnabled(),
+		return new RedisCacheConfiguration(getTtlFunction(), getAllowCacheNullValues(), isTimeToIdleEnabled(),
 				DO_NOT_USE_PREFIX, getKeyPrefix(), getKeySerializationPair(), getValueSerializationPair(), getConversionService());
 	}
 
@@ -221,9 +221,10 @@ public class RedisCacheConfiguration {
 	 *
 	 * @return this {@link RedisCacheConfiguration}.
 	 * @see <a href="https://redis.io/commands/getex/">GETEX</a>
+	 * @since 3.2.0
 	 */
-	public RedisCacheConfiguration enableTtiExpiration() {
-		return new RedisCacheConfiguration(getTtlFunction(), getAllowCacheNullValues(), ENABLE_IDLE_TIME_EXPIRATION,
+	public RedisCacheConfiguration enableTimeToIdle() {
+		return new RedisCacheConfiguration(getTtlFunction(), getAllowCacheNullValues(), USE_TIME_TO_IDLE_EXPIRATION,
 				usePrefix(), getKeyPrefix(), getKeySerializationPair(), getValueSerializationPair(),
 				getConversionService());
 	}
@@ -253,7 +254,7 @@ public class RedisCacheConfiguration {
 
 		Assert.notNull(ttlFunction, "TtlFunction must not be null");
 
-		return new RedisCacheConfiguration(ttlFunction, getAllowCacheNullValues(), isTtiExpirationEnabled(),
+		return new RedisCacheConfiguration(ttlFunction, getAllowCacheNullValues(), isTimeToIdleEnabled(),
 				usePrefix(), getKeyPrefix(), getKeySerializationPair(), getValueSerializationPair(),
 				getConversionService());
 	}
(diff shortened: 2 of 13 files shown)


src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -124,12 +124,14 @@ public void put(String name, byte[] key, byte[] value, @Nullable Duration ttl) {
...
 	@Override
-	public byte[] get(String name, byte[] key) {
+	public byte[] get(String name, byte[] key, @Nullable Duration ttl) {
 
 		Assert.notNull(name, "Name must not be null");
 		Assert.notNull(key, "Key must not be null");
 
-		byte[] result = execute(name, connection -> connection.get(key));
+		byte[] result = shouldExpireWithin(ttl)
+			? execute(name, connection -> connection.getEx(key, Expiration.from(ttl)))
```
quaff, line 133: Should we fallback to `get` and `expire` if server doesn't support `getex`?

jxblum: Thank you for the feedback, @quaff.

I spoke with @mp911de about this earlier today and we decided this feature will only be included from Spring Data Redis `3.2` and later, which is based on the latest Jedis and Lettuce drivers supporting most of the features in the newest Redis server releases.

In addition, it is an opt-in feature. See [here](https://github.com/spring-projects/spring-data-redis/pull/2649/files#diff-9b091dbe458d10f65774fb0a47e282e73b8535d4d9ae178342041badf6fb3e46R50) and [here](https://github.com/spring-projects/spring-data-redis/pull/2649/files#diff-9b091dbe458d10f65774fb0a47e282e73b8535d4d9ae178342041badf6fb3e46R225-R228), defaulting to the previous behavior.

quaff: > which is based on the latest Jedis and Lettuce drivers supporting most of the features in the newest Redis server releases.

I mean server not client doesn't support `getex`, it's a new command since redis 6.2.0, upgrading server is not convenient as client, many server still using 4.x and 5.x, especially service provided by private cloud.

mp911de: If the server does not support `getex` then you cannot use this feature.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
```diff
@@ -47,9 +47,11 @@
 public class RedisCacheConfiguration {
 
 	protected static final boolean DEFAULT_CACHE_NULL_VALUES = true;
+	protected static final boolean DEFAULT_ENABLE_TTI_EXPIRATION = false;
```
mp911de, line 50: Let's use `TIME_TO_IDLE` instead of `TTI` for more expressivity.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
```diff
@@ -47,9 +47,11 @@
...
 	protected static final boolean DEFAULT_USE_PREFIX = true;
 	protected static final boolean DO_NOT_CACHE_NULL_VALUES = false;
 	protected static final boolean DO_NOT_USE_PREFIX = false;
+	protected static final boolean ENABLE_IDLE_TIME_EXPIRATION = true;
```
mp911de, line 54: @quaff has a good point regarding server versions. We cannot assume the application runs on the latest Redis version. If we'd enabled this for everyone, then all users with older Redis versions suddenly see failures on their target environments unless they reconfigure the cache.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
```diff
@@ -108,14 +110,18 @@ public static RedisCacheConfiguration defaultCacheConfig(@Nullable ClassLoader c
...
+				DEFAULT_ENABLE_TTI_EXPIRATION,
+				DEFAULT_USE_PREFIX,
 				CacheKeyPrefix.simple(),
 				SerializationPair.fromSerializer(RedisSerializer.string()),
 				SerializationPair.fromSerializer(RedisSerializer.java(classLoader)),
 				conversionService);
 	}
 
 	private final boolean cacheNullValues;
+	private final boolean enableTtiExpiration;
```
mp911de, line 124: Let's expand on the name to `enableTimeToIdle` instead of `Tti`.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
```diff
@@ -193,8 +202,30 @@ public RedisCacheConfiguration disableCachingNullValues() {
...
+	 * Enabling this option applies the same {@link #getTtlFunction() TTL expiration policy} to {@link Cache} read
+	 * operations as it does for {@link Cache} write operations. In effect, this will invoke the Redis {@literal GETEX}
+	 * command in place of {@literal GET}.
+	 * <p>
+	 * Redis does not support the concept of {@literal TTI}, only {@literal TTL}. However, if {@literal TTL} expiration
+	 * is applied to all {@link Cache} operations, both read and write alike, and {@link Cache} operations passed with
+	 * expiration are used consistently across the application, then in effect, an application can achieve
+	 * {@literal TTI} expiration-like behavior.
+	 *
+	 * @return this {@link RedisCacheConfiguration}.
```
mp911de, line 222: Nit: `@since 3.2`

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
```diff
@@ -275,6 +307,19 @@ public boolean getAllowCacheNullValues() {
...
+	/**
+	 * Determines whether {@literal time-to-idle (TTI) expiration} has been enabled for caching.
+	 * <p>
+	 * Use {@link #enableTtiExpiration()} to opt-in and enable {@literal time-to-idle (TTI) expiration} for caching.
+	 *
+	 * @return {@literal true} if {@literal time-to-idle (TTI) expiration} was configured and enabled for caching.
+	 * Defaults to {@literal false}.
+	 * @see <a href="https://redis.io/commands/getex/">GETEX</a>
+	 */
+	public boolean isTtiExpirationEnabled() {
```
mp911de, line 319: Nit: `@since 3.2`

src/main/java/org/springframework/data/redis/cache/RedisCacheWriter.java
```diff
@@ -104,29 +106,43 @@ static RedisCacheWriter lockingRedisCacheWriter(RedisConnectionFactory connectio
...
-	 * @param value The value stored for the key. Must not be {@literal null}.
-	 * @param ttl Optional expiration time. Can be {@literal null}.
+	 * @param name must not be {@literal null}.
+	 * @param key must not be {@literal null}.
+	 * @return {@literal null} if key does not exist.
+	 * @see #get(String, byte[], Duration)
 	 */
-	void put(String name, byte[] key, byte[] value, @Nullable Duration ttl);
+	@Nullable
+	default byte[] get(String name, byte[] key) {
```
mp911de, line 122: Let's do the defaulting the other way round, that we introduce `get(String name, byte[] key, @Nullable Duration ttl)` as default method to not break existing implementations.

src/main/java/org/springframework/data/redis/core/types/Expiration.java
```diff
@@ -219,6 +220,25 @@ public boolean isUnixTimestamp() {
...
+		if (!(obj instanceof Expiration that)) {
+			return false;
+		}
+
+		return this.getTimeUnit().toMillis(getExpirationTime()) == that.getTimeUnit().toMillis(that.getExpirationTime());
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(getExpirationTime(), getTimeUnit());
```
mp911de, line 239: We use Spring's `ObjectUtils.nullSafeHashCode(…)` to compute the hash code in other places.

src/test/java/org/springframework/data/redis/cache/DefaultRedisCacheWriterUnitTests.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2017-2023 the original author or authors.
```
mp911de, line 2: Nit: Copyright starts at 2023 (inception of the file).

src/test/java/org/springframework/data/redis/cache/RedisCacheWriterUnitTests.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2017-2023 the original author or authors.
```
mp911de, line 2: Nit: Copyright starts at 2023.

src/test/java/org/springframework/data/redis/cache/RedisCacheWriterUnitTests.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for {@link RedisCacheWriter}.
+ *
+ * @author John Blum
+ */
+public class RedisCacheWriterUnitTests {
```
mp911de, line 37: Nit: JUnit 5 test classes can be package-private.

src/test/java/org/springframework/data/redis/core/types/ExpirationUnitTests.java
```diff
@@ -53,4 +56,38 @@ void fromMinutes() {
 		assertThat(expiration.getExpirationTime()).isEqualTo(5L * 60);
 		assertThat(expiration.getTimeUnit()).isEqualTo(TimeUnit.SECONDS);
 	}
+
+	@Test
+	public void equalValuedExpirationsAreEqual() {
```
mp911de, line 61: Nit: Test methods can be package-private.

src/test/java/org/springframework/data/redis/core/types/ExpirationUnitTests.java
```diff
@@ -13,17 +13,20 @@
...
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
 
+import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 
 import org.junit.jupiter.api.Test;
 
 /**
+ * Unit tests for {@link Expiration}.
```
mp911de, line 26: 🥳

mp911de (commented):
Great work, I left a few comments we should consider regarding server versions and naming.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
quaff: I did similar work before, like this:

```java

	private volatile boolean getexSupported = true;

	@Override
	public Object getWithTti(String key, int timeToIdle, TimeUnit timeUnit) {
		BoundValueOperations operations = redisTemplate.boundValueOps(key);
		try {
			if (getexSupported)
				return timeToIdle > 0 ? operations.getAndExpire(timeToIdle, timeUnit) : operations.getAndPersist();
			// fallback to get and expire
			Object result = operations.get();
			if (result != null && timeToIdle > 0)
				operations.expire(timeToIdle, timeUnit);
			return result;
		} catch (RuntimeException e) {
			log.error(e.getMessage(), e);
			if (e.getCause() instanceof RedisCommandExecutionException) {
				getexSupported = false; // server.version < 6.2
				return getWithTti(key,  timeToIdle, timeUnit);
			}
			throw e;
		}
	}
```

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
quaff: We could fallback to `get` and `expire`.

src/test/java/org/springframework/data/redis/cache/RedisCacheTests.java
```diff
@@ -15,10 +15,10 @@
  */
 package org.springframework.data.redis.cache;
 
-import static org.assertj.core.api.Assertions.*;
```
christophstrobl, line 18: The import changes do not seem in line with the import order defined in `spring-data-build`.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
jxblum: Will do.

Ironically, I initially spelled it out in favor of being explicit, but ultimately opted to match our use of the "TTL" abbreviation.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
jxblum: Please note, the `ENABLE_IDLE_TIME_EXPIRATION` constant is **not** a flag defaulting _time-to-idle_ (TTI) expiration to `true`.

I simply prefer not to use constant values like  `true` for boolean flag method arguments, especially when or where constructors and/or methods accept multiple boolean parameters.  For [instance](https://github.com/spring-projects/spring-data-redis/blob/issue/2351/src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java#L226), or [here](https://github.com/spring-projects/spring-data-redis/blob/issue/2351/src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java#L192) and [here](https://github.com/spring-projects/spring-data-redis/blob/issue/2351/src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java#L206).

However, I recognize the constant name (`ENABLE_IDLE_TIME_EXPIRATION `) is a really poor choice of words and have changed this to `USE_TIME_TO_IDLE_EXPIRATION`.

Keep in mind, this constant (now `USE_TIME_TO_IDLE_EXPIRATION`) is only ever used when TTI is enabled by calling the `enableTtiExpiration()` method on the `RedisCacheConfiguration` instance, [here](https://github.com/spring-projects/spring-data-redis/blob/issue/2351/src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java#L209-L229).

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
jxblum: Good catch!

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
jxblum: Done.

I also renamed `isTtiExpirationEnabled()` to `isTimeToIdleEnabled()` and `enableTtiExpiration()` to `enableTimeToIdle()`.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
jxblum: Done

src/test/java/org/springframework/data/redis/cache/DefaultRedisCacheWriterUnitTests.java
jxblum: Done

src/test/java/org/springframework/data/redis/cache/RedisCacheWriterUnitTests.java
jxblum: Done

src/test/java/org/springframework/data/redis/cache/RedisCacheWriterUnitTests.java
jxblum: Indeed

src/test/java/org/springframework/data/redis/core/types/ExpirationUnitTests.java
jxblum: Yes

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum: @quaff - I understood what you meant between client and server.

@christophstrobl and I are meeting this week on the topic of caching and using TTI-like expiration in SD Redis. Let me touch base with **Christoph** on this whether or not we will take into consideration Redis server versions.

src/main/java/org/springframework/data/redis/cache/RedisCacheWriter.java
jxblum: Good point.

src/main/java/org/springframework/data/redis/core/types/Expiration.java
jxblum: Right! Old habit.

jxblum: Spring Framework's `ObjectUtils.nullSafeHashCode(..)` is a bit more verbose and sloppy, but no matter.

src/main/java/org/springframework/data/redis/cache/RedisCacheConfiguration.java
jxblum: If @christophstrobl and I decide it makes sense to handle Redis server version for a (Redis) `Cache.get(key)`, using either `GETEX` or a `GET` with `EXPIRE` (although, I am not certain SD Redis takes Redis server version into consideration anywhere else, though), then I would envision something a bit more robust, such as by (proactively) inspecting the Redis server version on construction of the `DefaultRedisCacheWriter` using the `INFO` command from `RedisServerCommands` given a `RedisConnection` (retrieved from the factory).

jxblum:
I am not sure why, but the GitHub PR did not pick up my latest set of changes, a [polishing commit](https://github.com/spring-projects/spring-data-redis/compare/main...issue/2351) based on the review feedback above.

Issue #2651, opened 2023-07-26
`DefaultRedisSerializationContextBuilder.string()` should assign param to `stringTuple` instead of `hashValueTuple`


imzhoukunqiang:
Hi Team,
   
Below there is a code snippet from `DefaultRedisSerializationContext` ,line 130. The `string()` method assign param `tuple` to `hashValueTuple`, it should be assigned  `stringTuple` I think.

```java
		@Override
		public RedisSerializationContextBuilder<K, V> hashValue(SerializationPair<?> tuple) {

			Assert.notNull(tuple, "SerializationPair must not be null");

			this.hashValueTuple = tuple;
			return this;
		}

		@Override
		public RedisSerializationContextBuilder<K, V> string(SerializationPair<String> tuple) {

			Assert.notNull(tuple, "SerializationPair must not be null");
                        
			this.hashValueTuple = tuple;
			return this;
		}
```

jxblum:
Hi @imzhoukunqiang - Yes, upon review, you assessment here seems correct to me. Thank you for catching this problem and [contributing](https://github.com/spring-projects/spring-data-redis/pull/2652) a fix, as well as the nice set of test cases. Much appreciated.

It appears, this problem [stems back](https://github.com/spring-projects/spring-data-redis/blob/2.7.14/src/main/java/org/springframework/data/redis/serializer/DefaultRedisSerializationContext.java#L160) to the `2.7.x` branch as well.

I will finish the review, polish this a bit and merge it!

PR #2652, opened 2023-07-26
Fix bug issues/2651 ,wrong assignment in `DefaultRedisSerializationContextBuilder.string`

imzhoukunqiang:
`DefaultRedisSerializationContextBuilder.string()` should assign param to `stringTuple` instead of `hashValueTuple`
see #2651 


jxblum:
This has been reviewed, polished and merged back to `2.7.x` (onward). Thank you for the contribution!

PR #2659, opened 2023-08-03
Rename Function Lambda parameter to match command interface in (Default) ReactiveXxxOperations

### LINKED ISSUE #2658: Rename Lambda Function parameter to match command interface in ReactiveXxxOperations
In all (Default) `ReactiveXxxOperation` interfaces used by `ReactiveRedisTemplate`, the (Redis) operation methods refer to their corresponding command interface as  "`connection`" when passing the `Function` _Lambda_ to the `createMono(:Function<..>, Publisher<T>)` or `createFlux(:Function<..>, Publisher<T>)` methods, which is not technically accurate, and confusing.

The `Function` _Lambda_ expects an instance of reactive commands (e.g. `ReactiveSetCommands` in the (Default) `ReactiveSetOperations` interface), which is obtained from the `ReactiveRedisConnection`.

jxblum:
Resolves #2658

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
+	<version>3.2.0-GH-2658-SNAPSHOT</version>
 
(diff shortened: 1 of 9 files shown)


mp911de:

Issue #2662, opened 2023-08-04
Simplify logic in RedisMessageListenerContainer and supporting classes


jxblum:
Code across and within `org.springframework.data.redis.listener` classes is redundant.

PR #2663, opened 2023-08-05
Simplify logic in RedisMessageListenerContainer and supporting classes

jxblum:
See #2662


mp911de:
That was a good excercise, our code looks much clean now. That's merged, polished, and backported now.

Issue #2653, opened 2023-07-27
Pipeline mode has Thread-safety bug


zhaozhiguang:
#### Current Behavior

Use parallelStream for concurrent operations. The amount of data is about 300k. After execution, all results will be added to the pipe. However, this pipe is of the ArrayList type when initializing, so some results are not added to the pipe. Then, NullPointerException will be thrown when traversing the pipe when closePipeline()

<details>
<summary>Stack trace</summary>

```java
java.lang.NullPointerException:null
 at org.springframework.data.redis.connection.lettuce.LettuceConnection.closePipeline(LettuceConnection.java:515)
...
```

</details>

#### Input Code

<details>
<summary>Input Code</summary>

```java
redisTemplate.executePipeLined(connection -> {
	objectives.parallelStream().forEach(obj -> {
		connection.setCommands().sAdd(rawKey,rawValue);
	});
})
```

</details>

#### Environment

- Spring-Data-Redis version(s): [e.g. 2.5.6]
- Redis version: [e.g. 6.2.5]

jxblum:
@zhaozhiguang -

Perhaps you can explain what `objectives` is from your "_Input Code_":

```java
redisTemplate.executePipelined(connection -> {
	objectives.parallelStream().forEach(obj -> {
		connection.setCommands().sAdd(rawKey, rawValue);
	});
})
```

For my explanation, I assume (and I am reasonably certain) that:

1) `objectives` is some type of `Collection`
2) And, you are calling `parallelStream()` ([Javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Collection.html#parallelStream())) on this `Collection`
3) And, the `Collection` implementation supports parallel processing of the `Stream` returned by the `Collection` that is also the source for the elements contained in the `Stream`.

First of all, the Spring Data Redis `RedisConnection` interface is **not** documented as being Thread-safe, not in the [reference documentation](https://docs.spring.io/spring-data/redis/docs/current/reference/html/#redis:connectors) and not in [Javadoc](https://docs.spring.io/spring-data/redis/docs/current/api/org/springframework/data/redis/connection/RedisConnection.html).

[...]

PR #2667, opened 2023-08-08
Edit ref doc and Javadoc for connection classes to clearly state the Thread-safety guarantees

jxblum:
See #2653


mp911de:

PR #2669, opened 2023-08-10
Introduce executor configuration to connection factories for Executor to be used with ClusterCommandExecutor 

### LINKED ISSUE #2594: Introduce executor configuration to connection factories for `Executor` to be used with `ClusterCommandExecutor`
We default to `SimpleAsyncTaskExecutor` when creating a `ClusterCommandExecutor` without the possibility for users to be able to configure the executor. For applications to be able to use Virtual Threads, we should provide a way to configure the executor.

jxblum:
Closes #2594


src/main/java/org/springframework/data/redis/connection/RedisClusterConfiguration.java
```diff
@@ -51,6 +52,8 @@ public class RedisClusterConfiguration implements RedisConfiguration, ClusterCon
 
 	private @Nullable Integer maxRedirects;
 
+	private @Nullable AsyncTaskExecutor executor;
```
mp911de, line 55: `RedisClusterConfiguration` doesn't seem ideal to hold the executor. Our `Redis…Configuration` classes are intended to describe the endpoint(s) to connect to and provide specific connectivity options such as the password or redirection limit that are used by the clients directly. 

`ClusterCommandExecutor` is a Spring Data concept. The alternative of moving the executor into `…ClientConfiguration` seems also not ideal because the executor isn't required by the client.

How about adding `executor` as property to `[Jedis|Lettuce]ConnectionFactory` directly? We have a few properties already, that are factory-specific (`convertPipelineAndTxResults`, `shareNativeConnection`, `pipeliningFlushPolicy`)? While the executor is only required for clustering operations, this change would give us the opportunity to revise its purpose along the lines of a "general purpose executor for asynchronous activity such as running multiple commands concurrently".

What do you think?

christophstrobl: +1 for moving the executor to the connection factory.

mp911de:
That's merged and polished now. Going forward, we should limit our polishing commits not to outweigh the actual change and if we want to apply larger cleanups, then it is better to handle these as separate tickets that can be backported. I'm guilty here as well.

jxblum:
@mp911de & @christophstrobl - I never had a chance to respond to the feedback on this PR & review, so I am not sure why we went ahead and merged this without completing the process. AFAICT, this was not time critical and certainly wasn't acted on with consensus. Rather, it seems we pushed this along to cram it into the release tomorrow (Fri, 2023-August-18), when we still have another scheduled milestone release (`2023-1.0-M3`) in September (Fri, 2023-Sept-15).

IMO, the `Executor` should NOT be a property of the (Jedis/Lettuce) connection factories because the connection factories have no business (purpose or use) for the `Executor`. Worse yet, we now maintain a reference to the `Executor` in 2 places, [here](https://qing.jiaogeiwo.xyz/index?key=1a6aa6a2c98a12d61fbbeba27a7d5059&aid=0375035c661c62eef3d8da8e7484e973&lang=en&userInfo=%7B%22uid%22:21236,%22name%22:%225035048657%22,%22nickname%22:%22JJBAD%22%7D) and [here](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/connection/ClusterCommandExecutor.java#L52).

In general, the component that requires and uses a collaborator (for example, like the `Executor`) should be responsible for managing the lifecycle of the collaborator.

With respect to configuration, I don't understand what you mean by this statement:

> _The alternative of moving the executor into …ClientConfiguration seems also not ideal because the executor isn't required by the client._

Spring Data Redis is all client-side. There is no SD Redis for a Redis server, in the purest sense, which was something SD GemFire/Geode actually did support (i.e. SD GemFire/Geode could be and was used on the server-side).

To my understanding, the `ClusterCommandExecutor` is used by the "client" to execute, or direct Redis commands to the appropriate (server) nodes in the Redis cluster (I assume based on "interest", i.e. based on the data/shard maintained by a Redis node for which the command was intended??). Even the [Javadoc](https://docs.spring.io/spring-data/redis/docs/current/api/org/springframework/data/redis/connection/ClusterCommandExecutor.html) alludes to this component being client-directed.

Perhaps what would have been better in this case is a _Strategy_ interface for how to "resolve" the "configured" `Executor` used by the `ClusterCommandsExecutor`. In this manner the _Strategy_ interface can either provide an `Executor` directly, or indirectly, such as through the `BeanFactory`, or any other source for that matter.

The implementation of, say, the `ExecutorResolver` _Strategy_ interface (or any appropriately named interface, and perhaps even a type of `Suppler<AsyncTaskExecutor>`, or even Spring `ObjectProvider<AsyncTaskExecutor>` ([Javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/ObjectProvider.html))) could be responsible for managing single instances, new instances on request, changes to instance at runtime, whatever.

In any case, I don't agree with our current arrangement, and I think we should revisit this at some point.

PR #2671, opened 2023-08-10
Change default value for KeyspaceEventMessageListener on keyspace event notifications

### LINKED ISSUE #2670: Change default value for KeyExpirationEventMessageListener to disable notifications
The [default value](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M1/src/main/java/org/springframework/data/redis/listener/KeyspaceEventMessageListener.java#L43) in Spring Data Redis for `notify-keyspace-events` is set to `EA`. Therefore, when a user declares and registers a bean of type `KeyExpirationEventMessageListener` (which [extends](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M1/src/main/java/org/springframework/data/redis/listener/KeyExpirationEventMessageListener.java#L32) `KeyspaceEventMessageListener`) in the Spring `ApplicationContext`, Spring Data Redis will [send](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M1/src/main/java/org/springframework/data/redis/listener/KeyspaceEventMessageListener.java#L88) a `config set notify-keyspace-events EA` command to the Redis server on [initialization](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M1/src/main/java/org/springframework/data/redis/listener/KeyspaceEventMessageListener.java#L123-L126) (and then [init()](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M1/src/main/java/org/springframework/data/redis/listener/KeyspaceEventMessageListener.java#L73-L97)).

While this may be "convenient" during development, it is NOT (ever) appropriate for production deployments, and can lead to confusing behavior exactly like [this](https://github.com/spring-projects/spring-data-redis/issues/2654).

With a combination of _Lettuce_ attempting to reconnect (_Jedis_ is unable to reconnect) and the fact that Redis does not retain configuration changes between restarts (i.e. non-persistent config when changed at runtime; even when applied using the `redis-cli`), then this situation, such as in Issue #2654, can occur.

> NOTE: It is not uncommon for data stores to retain (persist) dynamic configuration changes between sessions.

Therefore, I recommend the default value in Spring Data Redis be changed to empty, precisely how the Redis server (in `redis.conf` from a clean install) defines this setting, itself. 

If users change the default (e.g. to `EA`, or otherwise), then they have chosen to "opt-in" and are therefore responsible for the change, such as when the Redis server needs to be restarted. 

As explained in Issue #2654, the appropriate location to declare "infrastructure-related" configuration settings to the Redis server is in the `redis.conf` file, where it belongs! Additionally, users must ensure any custom (modified) configuration file is then passed to and recognized by the Redis server on start.

> TIP: On a positive note, the customized/modified `redis.conf` used to change the default behavior of the Redis server can be checked into version control the same as the application source code that requires the Redis server to behave in a custom way.

In conclusion, the framework, or any library for that matter, should never be making "infrastructure" decisions for users.

jxblum:
Closes #2670

NOTE: the next effect of theses changes is preserving the existing behavior.

While the default value for `notifiy-keyspace-events` in the abstract class `KeyspaceEventMessageListener` changed from "`EA`" to "_empty_" in this PR, it is an `abstract class` in the first place, and users must extend the class with a custom implementation in any case. But, I have reluctantly let the default value for the 1 and only Spring Data Redis provided implementation, `KeyExpirationEventMessageListener`, to "`Ex`", thereby preserving the existing behavior, i.e. to notify on key expiration events.

This change came about because of #2654, which occurs for 2 reasons.

1) First, the default Redis server configuration setting for `notify-keyspace-events` is "_empty_" (or, in other words, disabled). Therefore, on Redis server restarts, any configuration supplied at runtime, such as is the case with a Spring Data Redis application when the `KeyExpirationEventMessageListener` (or any `KeyspaceEventMessageListener`) is registered in the Spring `ApplicationContext`, gets lost! This is even the case for the `redis-cli`.

2) In addition, any configuration coming from a Spring Data Redis application when a `KeyspaceEventMessageListener` (such as `KeyExpirationEventMessageListener`) is declared and registered in the Spring container, only happens once, on initialization. So, when the Redis server restarted, again this configuration is lost!

The user should be specifying Redis server (infrastructure) configuration in `redis.conf`, which can then also be managed in version control along with the application.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
+	<version>3.2.0-GH-2670-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 	<description>Spring Data module for Redis</description>
--- a/src/main/java/org/springframework/data/redis/listener/KeyExpirationEventMessageListener.java
+++ b/src/main/java/org/springframework/data/redis/listener/KeyExpirationEventMessageListener.java
@@ -25,24 +25,35 @@ import org.springframework.lang.Nullable;
 /**
  * {@link MessageListener} publishing {@link RedisKeyExpiredEvent}s via {@link ApplicationEventPublisher} by listening
  * to Redis keyspace notifications for key expirations.
+ * <p>
+ * For development-time convenience the {@link #setKeyspaceNotificationsConfigParameter(String)} is set to
+ * {@literal "Ex"}, by default. However, it is strongly recommended that users specifically set
+ * {@literal notify-keyspace-events} to the appropriate value on the Redis server, in {@literal redis.conf}.
+ * <p>
+ * Any Redis server configuration coming from your Spring (Data Redis) application only occurs during Spring container
+ * initialization, and is not persisted across Redis server restarts.
  *
  * @author Christoph Strobl
+ * @author John Blum
  * @since 1.7
  */
 public class KeyExpirationEventMessageListener extends KeyspaceEventMessageListener implements
 		ApplicationEventPublisherAware {
+
+	private static final String EXPIRED_KEY_EVENTS = "Ex";
 
 	private static final Topic KEYEVENT_EXPIRED_TOPIC = new PatternTopic("__keyevent@*__:expired");
 
 	private @Nullable ApplicationEventPublisher publisher;
 
 	/**
-	 * Creates new {@link MessageListener} for {@code __keyevent@*__:expired} messages.
+	 * Creates new {@link MessageListener} for {@code __keyevent@*__:expired} messages and configures notification on
+	 * expired keys ({@literal Ex}).
 	 *
 	 * @param listenerContainer must not be {@literal null}.
 	 */
 	public KeyExpirationEventMessageListener(RedisMessageListenerContainer listenerContainer) {
-		super(listenerContainer);
+		super(listenerContainer, EXPIRED_KEY_EVENTS);
 	}
 
 	@Override
--- a/src/main/java/org/springframework/data/redis/listener/KeyspaceEventMessageListener.java
+++ b/src/main/java/org/springframework/data/redis/listener/KeyspaceEventMessageListener.java
@@ -17,11 +17,14 @@ package org.springframework.data.redis.listener;
 
 import java.util.Properties;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.data.redis.connection.Message;
 import org.springframework.data.redis.connection.MessageListener;
 import org.springframework.data.redis.connection.RedisConnection;
+import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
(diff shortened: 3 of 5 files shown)


Lvbuqing:
Thank you for your help,
I read your PR carefully:
https://github.com/spring-projects/spring-data-redis/pull/2671/commits/98d25f11a601cb57dddd349e07d3231cd357a507
========
As you said:
Why this configuration for the Redis server was even added to Spring Data Redis (i.e. the KeyExpirationEventMessageListener component) to begin with...  
mp911de also said:
  this is both an infrastructure (i.e. "data management policy") and application concern. The framework definitely should not be deciding in this case, no matter how "convenient".

I have a few questions about this PR:

1. EA is deleted here
image
Why is it changed to Ex here?
image

2. According to the last discussion, in my opinion, we should not configure related properties for Redis, for example, we should not set related values through spring-data-redis，as shown in the figure
image

keyspaceNotificationsConfigParameter should be used as an expected value, not directly set

jxblum:
The only reason I set the value to a "`Ex`" in `KeyExpirationEventMessageListener` is to preserve existing behavior in the `3.x` line. In `4.0`, I think this should also be set to "empty".

However, I did [set the value](https://github.com/spring-projects/spring-data-redis/pull/2671/files#diff-d3d9d693b6967cfa4b0d238495796d540d04b658bf487c040a01270fef3d1d74R66) for `notify-keyspace-events` in the *abstract* `KeyspaceEventMessageListener` to "empty", thereby disabling notifications on all key(space) events.

christophstrobl:
thanks for picking up this topic. I agree that, though convenient, the Redis config should not be altered by the message listener. However, I'd rather not change defaults in a minor but wait for the next major release. 
IIRC there's `MappingExpirationListener` in `KeyValueAdapter` used in conjunction with TTL, that would interact with the server config, if `EnableKeyspaceEvents` is set to `ON_STARTUP`/`ON_DEMAND`. It's `OFF` by default in the adapter and `EnableRedisRepositories`.

jxblum:
@christophstrobl - Thank you for your feedback.

Correct me if I am wrong, but it seems that 1) the `RedisKeyValueAdapter.MessageExpirationListener` [extends](https://github.com/spring-projects/spring-data-redis/blob/issue/2670/src/main/java/org/springframework/data/redis/core/RedisKeyValueAdapter.java#L760) from the SD Redis `KeyExpirationEventMessageListener` class directly where [I kept, though declared a (minimal) value for expiration](https://github.com/spring-projects/spring-data-redis/blob/issue/2670/src/main/java/org/springframework/data/redis/listener/KeyExpirationEventMessageListener.java#L43) (i.e. `Ex` rather than `EA`; FYR: definitions of [E](https://github.com/jxblum/redis-experiments/blob/spring-data-redis-issue-2670/spring-data-redis-experiments/src/main/resources/redis.conf#L1618) and [x](https://github.com/jxblum/redis-experiments/blob/spring-data-redis-issue-2670/spring-data-redis-experiments/src/main/resources/redis.conf#L1625)) which seems to be all that is required to receive key expiration events.

Perhaps `EA` is overkill since `A` is `g$lshzxetd` ([all this](https://github.com/jxblum/redis-experiments/blob/spring-data-redis-issue-2670/spring-data-redis-experiments/src/main/resources/redis.conf#L1619-L1628) as [documented](https://github.com/jxblum/redis-experiments/blob/spring-data-redis-issue-2670/spring-data-redis-experiments/src/main/resources/redis.conf#L1630) in `redis.conf`), which would only increase the amount of network traffic when pub/sub is enabled since then the SD (Redis) _Repository_ infrastructure would receive notifications for all events when we are only seemingly interested in expiration events (hence "`MappingExpirationListener`" as named).

Even the [code](https://github.com/spring-projects/spring-data-redis/blob/issue/2670/src/main/java/org/springframework/data/redis/core/RedisKeyValueAdapter.java#L783-L785) suggests we filtering for "expiration-only" events, though the [logic](https://github.com/spring-projects/spring-data-redis/blob/issue/2670/src/main/java/org/springframework/data/redis/core/RedisKeyValueAdapter.java#L823) (and then [this](https://github.com/spring-projects/spring-data-redis/blob/issue/2670/src/main/java/org/springframework/data/redis/core/convert/MappingRedisConverter.java#L1314-L1330)) is very cryptic to me. The logic only seems to be checking for the presence of a valid "key" (i.e. `keyspace:id`) in the message rather than whether the message is really an "expiration" event. That is, I am not sure if other events (e.g. on `ADD` or `DEL`) would also contain the key, and if so, what really constitutes an "expiration" message.

christophstrobl:
Thanks for digging up the documentation, being well aware of the config parameters, my point was about changing defaults in a minor, given there are ways to configure it from outside.

jxblum:
@christophstrobl - I understood your concern and your point. I was also trying to explain that we are not really changing behavior here since we are still receiving "expiration" events with this configuration.

Ideally, Redis clients are not sending any configuration to the servers, otherwise we are going to (possibly) run into situations exactly like this.

jxblum:
@christophstrobl - I also stumbled upon and noticed this [configuration](https://github.com/spring-projects/spring-data-redis/blob/3.1.2/src/main/java/org/springframework/data/redis/repository/configuration/EnableRedisRepositories.java#L196) in SD (Redis) Repositories today (i.e. `Ex`).

PR #2672, opened 2023-08-14
Terminate stream with error on `null` values returned by `RedisElementReader` for top-level elements

### LINKED ISSUE #2655: Nullability of `RedisElementReader.read(…)` contradicts non-nullability of `Flux.map(…)`
`RedisElementReader.read ` allow return null,but `ReactiveXXXOperations` using `Flux.map` will result in `The mapper [......] returned a null value.` errors.
https://github.com/spring-projects/spring-data-redis/blob/5568ffd81843b142026fefcba8f48cd7124591f1/src/main/java/org/springframework/data/redis/core/DefaultReactiveSetOperations.java#L339-L345

In my case, `RedisElementReader.read` well return null when illegal data format.

Is using `.mapNotNull` better ?

mp911de:
We now use `mapNotNull` to map Redis response values to avoid `NullPointerExceptions` caused by `RedisElementReader` returning `null`.

This effectively turns `RedisElementReader` into a filter function that can suppress elements from being emitted.

Closes #2655 

If merged, we should bring these changes to all maintained 3.x branches.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/DefaultReactiveGeoOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultReactiveGeoOperations.java
@@ -328,7 +328,7 @@ class DefaultReactiveGeoOperations<K, V> implements ReactiveGeoOperations<K, V> 
 
 	private GeoResult<GeoLocation<V>> readGeoResult(GeoResult<GeoLocation<ByteBuffer>> source) {
 
-		return new GeoResult<>(new GeoLocation(readValue(source.getContent().getName()), source.getContent().getPoint()),
+		return new GeoResult<>(new GeoLocation<>(readValue(source.getContent().getName()), source.getContent().getPoint()),
 				source.getDistance());
 	}
 }
--- a/src/main/java/org/springframework/data/redis/core/DefaultReactiveGeoOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultReactiveGeoOperations.java
@@ -26,7 +26,6 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.reactivestreams.Publisher;
-
 import org.springframework.data.geo.Circle;
 import org.springframework.data.geo.Distance;
 import org.springframework.data.geo.GeoResult;
@@ -40,6 +39,7 @@ import org.springframework.data.redis.domain.geo.GeoReference;
 import org.springframework.data.redis.domain.geo.GeoReference.GeoMemberReference;
 import org.springframework.data.redis.domain.geo.GeoShape;
 import org.springframework.data.redis.serializer.RedisSerializationContext;
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
@@ -321,6 +321,7 @@ class DefaultReactiveGeoOperations<K, V> implements ReactiveGeoOperations<K, V> 
 		return serializationContext.getValueSerializationPair().write(value);
 	}
 
+	@Nullable
 	private V readValue(ByteBuffer buffer) {
 		return serializationContext.getValueSerializationPair().read(buffer);
 	}
--- a/src/main/java/org/springframework/data/redis/core/DefaultReactiveListOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultReactiveListOperations.java
@@ -27,11 +27,13 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 import org.reactivestreams.Publisher;
+import org.springframework.dao.InvalidDataAccessApiUsageException;
 import org.springframework.data.redis.connection.ReactiveListCommands;
 import org.springframework.data.redis.connection.ReactiveListCommands.Direction;
 import org.springframework.data.redis.connection.ReactiveListCommands.LPosCommand;
 import org.springframework.data.redis.connection.RedisListCommands.Position;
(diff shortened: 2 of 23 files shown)


jxblum (commented):
I think these changes and Javadoc edits look completely fine to me.

christophstrobl:
I'm  a bit torn on this one, since the `map` operator on `Flux` accepts a plain `Function` which may return `null` with all the consequences when used in a reactive environment. The outlined snippet demonstrates this quite well, and as @mp911de mentioned in the original ticket, if used that way then this is more of an application concern. While this way of filtering will do the job, others may use exactly that behaviour to raise errors in the flow.

mp911de:
So what do we want to do here? Our serializer clearly allow for `null` values. A serializer cannot know whether it is being used in the specific arrangement of top-level values (e.g. JDK serializer that serializes a `null` value).

mp911de:
I rewrote this pull request to document the behavior in the context of top-level stream values. We also now terminate the stream in a controlled manner by emitting `InvalidDataAccessApiUsageException`.

jxblum:
I am reviewing this PR now and I generally like the direction the changes took.

However, I don't necessarily agree with this:

> _If merged, we should bring these changes to all maintained 3.x branches._

I think this is a significant change in the contract and behavior of SD Redis that will potentially and adversely affect the older GA'd branches in the `3.x` line, mainly `3.0.x` and `3.1.x` now, both of which are GA.

Honestly, with regard to #2655, I really think this is a user problem and application concern more than it is a framework issue that needs to be addressed and handled by the framework.

While these changes are an improvement overall (i.e. failing fast when Reactive streams are misused as the user has done in #2655), it is also a clear change in behavior that diverges from the specification when the [Reactive Streams specification](https://github.com/reactive-streams/reactive-streams-jvm#specification) is very clear on the matter ([Rule 2.13](https://github.com/reactive-streams/reactive-streams-jvm#2-subscriber-code)) of `null` elements in a stream (thereby causing and throwing a `NullPointerException`).

src/main/java/org/springframework/data/redis/core/DefaultReactiveListOperations.java
```diff
@@ -285,7 +288,8 @@ public Mono<V> rightPopAndLeftPush(K sourceKey, K destinationKey, Duration timeo
 		Assert.isTrue(isZeroOrGreater1Second(timeout), "Duration must be either zero or greater or equal to 1 second");
 
 		return createMono(
-				connection -> connection.bRPopLPush(rawKey(sourceKey), rawKey(destinationKey), timeout).map(this::readValue));
+				connection -> connection.bRPopLPush(rawKey(sourceKey), rawKey(destinationKey), timeout)
+						.map(this::readRequiredValue));
```
jxblum, line 292: Shouldn't this be `.mapNotNull(..)` given the `Duration` parameter?

mp911de: The previous variant was `mapNotNull` that effectively filters `null` values turning `RedisElementReader` into something that has an ability of filtering. That's why we chose to use `readRequiredValue` throwing a proper exception.

jxblum: This previous variant was `.map(this::readValue)`, line 288.

I was thinking more along the lines of the `rightPop(key, :Duration)` overloaded method above on line 262.

But, perhaps the choice of simply `.map(this:readRequiredValue)` vs. `.mapNotNull(this:readRequiredValue)` is due to the `right pop` followed by `left push` compound operation??

src/main/java/org/springframework/data/redis/core/ReactiveStreamOperations.java
```diff
@@ -23,32 +23,20 @@
...
 import org.springframework.data.redis.connection.RedisStreamCommands.XClaimOptions;
-import org.springframework.data.redis.connection.stream.ByteBufferRecord;
-import org.springframework.data.redis.connection.stream.Consumer;
-import org.springframework.data.redis.connection.stream.MapRecord;
-import org.springframework.data.redis.connection.stream.ObjectRecord;
-import org.springframework.data.redis.connection.stream.PendingMessage;
-import org.springframework.data.redis.connection.stream.PendingMessages;
-import org.springframework.data.redis.connection.stream.PendingMessagesSummary;
-import org.springframework.data.redis.connection.stream.ReadOffset;
+import org.springframework.data.redis.connection.stream.*;
```
jxblum, line 29: We should avoid star imports.

src/main/java/org/springframework/data/redis/core/DefaultReactiveGeoOperations.java
```diff
@@ -40,6 +39,7 @@
 import org.springframework.data.redis.domain.geo.GeoReference.GeoMemberReference;
 import org.springframework.data.redis.domain.geo.GeoShape;
 import org.springframework.data.redis.serializer.RedisSerializationContext;
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
```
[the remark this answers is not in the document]
jxblum: I think, perhaps, `ReactiveGeoOperations` needs further refinements down the road, particularly with guards around possible `null` values.

jxblum:
The changes have been merged to `main` (only) for `3.2`.

mp911de:
Please close pull requests and remove their branches after merge.

Issue #2677, opened 2023-08-15
`Jsr310Converters` does not contain converters for `OffsetTime` and `OffsetDateTime`


DeltaIII:
When using `spring-data-redis` and `RedisSerializer.java()` I noticed that a field of type `OffsetTime` wasn't being serialised even though it implements `Serializable`. The root cause is that `SimpleTypeHolder` assumes that any class that starts with `java.time` is a simple type (which in turn means `CustomConversions.isSimpleType(OffsetTime.class) == true`). This breaks the assumption that `CustomConversions` has a converter for the type as per the javadoc:
> Returns whether the given type is considered to be simple. That means it's either a general simple type or we have a writing Converter registered for a particular type.

Expected behaviour:
- `customConversions.isSimpleType` returns false if there are no converters for `OffsetDateTime` or `OffsetTime`

Actual behaviour:
- `customConversions.isSimpleType(OffsetDateTime.class) == true` even if there are no converters for `OffsetDateTime` (and for `OffsetTime`

Relevant Code links:
https://github.com/spring-projects/spring-data-commons/blob/541f0ced32f3f8b2143ea5b793a489828567ff13/src/main/java/org/springframework/data/mapping/model/SimpleTypeHolder.java#L158

https://github.com/spring-projects/spring-data-commons/blob/541f0ced32f3f8b2143ea5b793a489828567ff13/src/main/java/org/springframework/data/convert/CustomConversions.java#L190

mp911de:
I moved this ticket into Spring Data Redis as it is a module-specific concern. We have already quite a few [converters](https://github.com/spring-projects/spring-data-redis/blob/45621dd3ee8820286c43bb88b7b7697e9bb09e8b/src/main/java/org/springframework/data/redis/core/convert/Jsr310Converters.java#L44) for JSR-310 types, however we don't have ones for `Offset[Date]Time`.

Since you're a bit familiar with our code, do you want to submit a pull request so that we can ship a fix with the next release?

jxblum:
FTR: I was able to produce a similar problem (I suppose) by simply adding a `OffsetDateTime` (persistent) property to the `Person` class from the [Spring Data Examples](https://github.com/spring-projects/spring-data-examples), Spring Data Redis [Repositories example](https://github.com/spring-projects/spring-data-examples/tree/main/redis/repositories), and setting a few `OffsetDateTime` values for the "_Starks_" (e.g. "_arya_").

Modified `Person` class:

```java
class Person {

[...]

PR #2681, opened 2023-08-17
Register Converters for Offset java.time types in Jsr310Converters

jxblum:
See #2677

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/convert/BinaryConverters.java
+++ b/src/main/java/org/springframework/data/redis/core/convert/BinaryConverters.java
@@ -20,7 +20,9 @@ import java.nio.charset.StandardCharsets;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Date;
+import java.util.Set;
 import java.util.UUID;
 
 import org.springframework.core.convert.converter.Converter;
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/data/redis/core/convert/BinaryConverters.java
```diff
@@ -46,6 +48,24 @@ final class BinaryConverters {
...
+			new BinaryConverters.NumberToBytesConverter(),
+			new BinaryConverters.BytesToNumberConverterFactory(),
+			new BinaryConverters.EnumToBytesConverter(),
+			new BinaryConverters.BytesToEnumConverterFactory(),
+			new BinaryConverters.BooleanToBytesConverter(),
+			new BinaryConverters.BytesToBooleanConverter(),
+			new BinaryConverters.DateToBytesConverter(),
+			new BinaryConverters.BytesToDateConverter(),
+			new BinaryConverters.UuidToBytesConverter(),
+			new BinaryConverters.BytesToUuidConverter()
```
christophstrobl, line 65: I think we can drop the extra `BinaryConverters.` part here.

src/main/java/org/springframework/data/redis/core/convert/Jsr310Converters.java
```diff
@@ -58,6 +61,7 @@ public abstract class Jsr310Converters {
 		}
 
 		List<Converter<?, ?>> converters = new ArrayList<>();
```
christophstrobl, line 63: could be changed to `List.of` or initialized with an initial size.

src/main/java/org/springframework/data/redis/core/convert/Jsr310Converters.java
```diff
@@ -296,4 +304,51 @@ public Duration convert(byte[] source) {
 		}
 	}
 
+	/**
+	 * @author John Blum
+	 * @see java.time.OffsetDateTime
```
christophstrobl, line 309: though not a `public` class, having a `@since` tag can help with code forensic

mp911de:

PR #2689, opened 2023-08-22
Add ttl-offset option to RedisCacheConfiguration to avoid cache avalanche problem

liudaac:
Due to the lack of a configuration option in RedisCacheConfiguration for developers to add random TTL offset to the basic TTL, the cache avalanche problem often occurred when caching large amounts of data. In order to better facilitate developers using the default implementation provided by Spring Redis, a TTL offset setting needs to be added on top of the TTL to support random offset settings.  This option can effectively avoid cache avalanche problem.

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -17,12 +17,14 @@ package org.springframework.data.redis.cache;
 
 import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
+import java.time.Duration;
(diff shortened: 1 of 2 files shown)


mp911de:
In newer versions, we introduced a `TtlFunction` interface that allows you to compute the TTL as function, including randomization offsets. We recommend this functionality instead of broadening our configuration options.

Please also note that we do not backport features to older Spring Data versions.

liudaac:
However, you are overlooking one thing, which is that due to compatibility issues with JDK and other factors, many people are still using version 2.7.x, which is currently still being iterated by the Spring official.

PR #2691, opened 2023-08-23
Use Locks instead of synchronized blocks to run blocking operations

### LINKED ISSUE #2690: Revisit `synchronized` blocks involving blocking operations
### Version
- Spring Boot 3.2.0-M1
- Spring Data Redis 3.2.0-M1 & lettuce-core 6.2.5.RELEASE
- Java 21 (Temurin-21+35-202308102344)

### Issue
Detects virtual thread parking & pinning when `LettuceConnectionFactory.getSharedConnection()` is called.
Stack trace is below. (need to be apply `-Djdk.tracePinnedThreads=full`)
```
Thread[#116,ForkJoinPool-1-worker-2,5,CarrierThreads]
    java.base/java.lang.VirtualThread$VThreadContinuation.onPinned(VirtualThread.java:185)
    java.base/jdk.internal.vm.Continuation.onPinned0(Continuation.java:393)
    java.base/java.lang.VirtualThread.park(VirtualThread.java:592)
    java.base/java.lang.System$2.parkVirtualThread(System.java:2639)
    java.base/jdk.internal.misc.VirtualThreads.park(VirtualThreads.java:54)
    java.base/java.util.concurrent.locks.LockSupport.park(LockSupport.java:219)
    java.base/java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1864)
    java.base/java.util.concurrent.ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3780)
    java.base/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3725)
    java.base/java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1898)
    java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2072)
    io.lettuce.core.DefaultConnectionFuture.get(DefaultConnectionFuture.java:69)
    io.lettuce.core.AbstractRedisClient.getConnection(AbstractRedisClient.java:345)
    io.lettuce.core.RedisClient.connect(RedisClient.java:216)
    org.springframework.data.redis.connection.lettuce.StandaloneConnectionProvider.lambda$getConnection$1(StandaloneConnectionProvider.java:112)
    java.base/java.util.Optional.orElseGet(Optional.java:364)
    org.springframework.data.redis.connection.lettuce.StandaloneConnectionProvider.getConnection(StandaloneConnectionProvider.java:112)
    org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory$ExceptionTranslatingConnectionProvider.getConnection(LettuceConnectionFactory.java:1626)
    org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory$SharedConnection.getNativeConnection(LettuceConnectionFactory.java:1453)
    org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory$SharedConnection.getConnection(LettuceConnectionFactory.java:1436) <== monitors:1
    org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory.getSharedConnection(LettuceConnectionFactory.java:1127)
    org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory.getConnection(LettuceConnectionFactory.java:465)
...
```

Do you have any plan that avoid thread pinning? (ex. Use `ReentrantLock`)

mp911de:
To avoid thread pinning on virtual thread arrangements we now use ReentrantLock instead of a synchronized block.

Closes #2690

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
@@ -41,6 +41,8 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
@@ -138,6 +140,8 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 
 	/** Synchronization monitor for the shared Connection */
 	private final Object connectionMonitor = new Object();
+
+	private final Lock lock = new ReentrantLock();
 
 	private PipeliningFlushPolicy pipeliningFlushPolicy = PipeliningFlushPolicy.flushEachCommand();
 
@@ -1480,7 +1484,8 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 		@Nullable
 		StatefulConnection<E, E> getConnection() {
 
-			synchronized (this.connectionMonitor) {
+			lock.lock();
+			try {
 
 				if (this.connection == null) {
 					this.connection = getNativeConnection();
@@ -1491,6 +1496,8 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 				}
 
 				return this.connection;
+			} finally {
+				lock.unlock();
 			}
 		}
 
@@ -1508,7 +1515,8 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 		 */
 		void validateConnection() {
 
-			synchronized (this.connectionMonitor) {
+			lock.lock();
+			try {
 
 				boolean valid = false;
 
@@ -1535,6 +1543,8 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 					resetConnection();
 					this.connection = getNativeConnection();
 				}
+			} finally {
+				lock.unlock();
 			}
 		}
 
@@ -1543,13 +1553,16 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 		 */
 		void resetConnection() {
 
-			synchronized (this.connectionMonitor) {
+			lock.lock();
+			try {
 
 				if (this.connection != null) {
 					this.connectionProvider.release(this.connection);
 				}
 
 				this.connection = null;
+			} finally {
+				lock.unlock();
 			}
 		}
 	}
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
+	<version>3.2.0-GH-2690-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 	<description>Spring Data module for Redis</description>
--- a/src/main/java/org/springframework/data/redis/cache/RedisCache.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCache.java
@@ -24,6 +24,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.StringJoiner;
 import java.util.concurrent.Callable;
+import java.util.concurrent.locks.Lock;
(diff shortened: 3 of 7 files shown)


src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
```diff
@@ -139,6 +141,8 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 	/** Synchronization monitor for the shared Connection */
 	private final Object connectionMonitor = new Object();
```
giger85, line 142: @mp911de 
I expected `connectionMonitor` field and other `synchronized` blocks (e.g in getOrCreateSharedConnection() method) to be deleted, but it was not.
Is it ok?

mp911de: `connectionMonitor` is used in `synchronized` blocks that are of a computational nature to enforce happens-before relationships.

giger85: @mp911de 
Ok, I see.

giger85 (approved):

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
jxblum: The `connectionMonitor` is only really used to initialize references to the single, "shared" connection(s) [here](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M2/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java#L1140-L1147) and [here](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M2/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java#L1152-L1159), then subsequently (and necessarily) [used](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M2/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java#L1120-L1124) to uninitialize the references once the connections are reset (released).

There is nothing "computationally" happening within these `synchronized` blocks, AFAICT. 

For instance, the connection is not actually established yet. That only occurs further up in the call stack, inside `getSharedConnection(..)` ([here](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M2/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java#L1175)) and `getSharedReactiveConnection(..)` ([here](https://github.com/spring-projects/spring-data-redis/blob/3.2.0-M2/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java#L1200)), which do not get executed inside a `synchronized` block on the `connectionMonitor`.

Unless I am overlooking something (@mp911de - correct me if I am wrong), technically, it should not matter what (Virtual or Platform) Thread initializes (gets or creates) and resets the connections, given that all the `connectionMonitor` is used for is to un/initialize references to the connections, thereby ensuring only a **single**, "shared" connection (for both reactive and non-reactive) is created and returned to the calling Thread.

> NOTE: `ReentrantLock` has the same mutex, visibility and reentrant behavior as `synchronized` blocks (see [Javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html)).

If by "_happens-before_" you mean the `null` check before, or the return statement after the creation of the connection, then logically the statement ordering as seen by the Thread and determined by the runtime will necessarily be as it appears in code. Semantically speaking, it has to (collectively, the creation is inside the "`if`" statement so is a moot point; it only boils down to the `if` and `return` statements, which semantically cannot be reordered by the runtime in this case). This is `true` regardless if the code is inside a `synchronized` block (or even `ReentrantLock`) or not.

Again, maybe I am overlooking something here?

jxblum: I imagine there should be some benefit here in different Virtual Threads acquiring the "shared" connection from the factory in the case of a highly concurrent Spring-based Web application, where Virtual Threads are configured. Underneath, the user maybe using a pooled `LettuceConnectionProvider`.

Thoughts?

jxblum: After discussing this with @mp911de this morning, I made the change to the `synchronized` blocks in `LettuceConnectionFactory` to now use the `ReentrantLock`.

jxblum:
This has been merged to `main` for the upcoming `2023.1.0-M3` (SD Redis `3.2.0-M3`) release.

Issue #2612, opened 2023-06-15
Upgrade to Jedis 5.0


christophstrobl:
Jedis 5 already released its [2nd beta](https://github.com/redis/jedis/releases/tag/v5.0.0-beta2) so we could follow along on current _main_ line till the GA is out.

jxblum:
I just did a quick stint on this, and upgrading Spring Data Redis to Jedis `5.0.0-beta2`, and it currently leads to the following compilation errors:

```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /Users/jblum/pivdev/spring-data-redis/src/main/java/org/springframework/data/redis/connection/jedis/JedisInvoker.java:[20,26] error: cannot find symbol
  symbol:   class Queable
  location: package redis.clients.jedis
[ERROR] /Users/jblum/pivdev/spring-data-redis/src/main/java/org/springframework/data/redis/connection/jedis/JedisInvoker.java:[1053,40] error: cannot find symbol
  symbol:   class Queable
  location: class JedisInvoker
[ERROR]   reason: class file for javax.annotation.meta.When not found
/Users/jblum/pivdev/spring-data-redis/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterStreamCommands.java:[272,50] error: no suitable method found for xpending(byte[],byte[],byte[],byte[],int,byte[])
    method UnifiedJedis.xpending(String,String) is not applicable
      (actual and formal argument lists differ in length)
    method UnifiedJedis.xpending(String,String,XPendingParams) is not applicable
      (actual and formal argument lists differ in length)
    method UnifiedJedis.xpending(byte[],byte[]) is not applicable
      (actual and formal argument lists differ in length)
    method UnifiedJedis.xpending(byte[],byte[],XPendingParams) is not applicable
      (actual and formal argument lists differ in length)
[ERROR] /Users/jblum/pivdev/spring-data-redis/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterHashCommands.java:[163,73] error: incompatible types: List<Entry<byte[],byte[]>> cannot be converted to Map<byte[],byte[]>
[ERROR] /Users/jblum/pivdev/spring-data-redis/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterHashCommands.java:[188,73] error: incompatible types: List<Entry<byte[],byte[]>> cannot be converted to Map<byte[],byte[]>
[...]

PR #2716, opened 2023-09-25
Upgrade to Jedis 5.0

jxblum:
See #2612


mp911de:
That's merged via f97bd87d and 20756338

Issue #2743, opened 2023-10-17
Clarify operations in asynchronous caching implementation


jxblum:
See #2650 

See Original Pull Request #2717 

Use more descriptive names for operations, especially Reactive operations, by either calling a local, private method or introducing a variable with strongly typed parameters.

For example, this `Function<ReactiveRedisConnection, Mono<T>> commandExecution` is more readable and descriptive than `it -> Mono.fromCompletionStage(callback.apply(it))`. Outside of the IDE (such as reading source in GitHub), I immediately know that "`it'` is a Redis (Reactive) connection, that I am executing a Redis command and that I am getting back a Mono of `T` (the value from the "reactive" command execution).

Edits Javadoc.

PR #2744, opened 2023-10-17
Refine and clarify operations in the async cache implementation

jxblum:
See #2743

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -17,6 +17,7 @@ package org.springframework.data.redis.cache;
 
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.SignalType;
 
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
@@ -24,6 +25,7 @@ import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 import org.springframework.dao.PessimisticLockingFailureException;
@@ -38,7 +40,7 @@ import org.springframework.data.redis.util.ByteUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
-import org.springframework.util.ObjectUtils;
+
 
 /**
  * {@link RedisCacheWriter} implementation capable of reading/writing binary data from/to Redis in {@literal standalone}
@@ -47,11 +49,11 @@ import org.springframework.util.ObjectUtils;
  * <p>
  * {@link DefaultRedisCacheWriter} can be used in
  * {@link RedisCacheWriter#lockingRedisCacheWriter(RedisConnectionFactory) locking} or
- * {@link RedisCacheWriter#nonLockingRedisCacheWriter(RedisConnectionFactory) non-locking} mode. While
- * {@literal non-locking} aims for maximum performance it may result in overlapping, non-atomic, command execution for
- * operations spanning multiple Redis interactions like {@code putIfAbsent}. The {@literal locking} counterpart prevents
- * command overlap by setting an explicit lock key and checking against presence of this key which leads to additional
- * requests and potential command wait times.
+ * {@link RedisCacheWriter#nonLockingRedisCacheWriter(RedisConnectionFactory) non-locking} mode. While {@literal non-locking}
+ * aims for maximum performance it may result in overlapping, non-atomic, command execution for operations spanning
+ * multiple Redis interactions like {@code putIfAbsent}. The {@literal locking} counterpart prevents command overlap
+ * by setting an explicit lock key and checking against presence of this key which leads to additional requests
+ * and potential command wait times.
  *
  * @author Christoph Strobl
  * @author Mark Paluch
@@ -61,8 +63,12 @@ import org.springframework.util.ObjectUtils;
  */
 class DefaultRedisCacheWriter implements RedisCacheWriter {
 
-	private static final boolean REACTIVE_REDIS_CONNECTION_FACTORY_PRESENT = ClassUtils
-			.isPresent("org.springframework.data.redis.connection.ReactiveRedisConnectionFactory", null);
+	public static final boolean FLUX_PRESENT = ClassUtils.isPresent("reactor.core.publisher.Flux", null);
+
+	private static final boolean REACTIVE_REDIS_CONNECTION_FACTORY_PRESENT =
+			ClassUtils.isPresent("org.springframework.data.redis.connection.ReactiveRedisConnectionFactory", null);
+
+	private final AsyncCacheWriter asyncCacheWriter;
 
 	private final BatchStrategy batchStrategy;
 
@@ -74,8 +80,6 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 
 	private final TtlFunction lockTtl;
 
-	private final AsyncCacheWriter asyncCacheWriter;
-
 	/**
 	 * @param connectionFactory must not be {@literal null}.
 	 * @param batchStrategy must not be {@literal null}.
@@ -86,8 +90,8 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 
 	/**
 	 * @param connectionFactory must not be {@literal null}.
-	 * @param sleepTime sleep time between lock request attempts. Must not be {@literal null}. Use {@link Duration#ZERO}
-	 *          to disable locking.
+	 * @param sleepTime sleep time between lock request attempts. Must not be {@literal null}.
+	 * Use {@link Duration#ZERO} to disable locking.
 	 * @param batchStrategy must not be {@literal null}.
 	 */
 	DefaultRedisCacheWriter(RedisConnectionFactory connectionFactory, Duration sleepTime, BatchStrategy batchStrategy) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -116,12 +103,12 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
...
+				: UnsupportedAsyncCacheWriter.INSTANCE;
+	}
 
-		if (REACTIVE_REDIS_CONNECTION_FACTORY_PRESENT && this.connectionFactory instanceof ReactiveRedisConnectionFactory) {
-			asyncCacheWriter = new AsynchronousCacheWriterDelegate();
-		} else {
-			asyncCacheWriter = UnsupportedAsyncCacheWriter.INSTANCE;
-		}
+	private boolean isAsyncCacheSupportEnabled() {
+		return this.connectionFactory instanceof ReactiveRedisConnectionFactory;
```
mp911de, line 111: We're testing assignability against an interface that defines potentially absent types such as `Mono` and `Publisher`. Loading interfaces can work under certain circumstances, however depending on the runtime verifier mechanics, loading `ReactiveRedisConnectionFactory` without Project Reactor can cause `NoClassDefFoundError`.

Therefore, we used a safe scheme that is being used across the majority of portfolio projects and there's no reason to deviate from that.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -401,6 +382,14 @@ private static boolean shouldExpireWithin(@Nullable Duration ttl) {
 		return ttl != null && !ttl.isZero() && !ttl.isNegative();
 	}
 
+	private Expiration toExpiration(Duration ttl) {
```
mp911de, line 385: These tiny methods are unnecessary and generate a lot of noise making it harder to read the code.

mp911de (changes requested):
Did a first review pass.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum: Understood

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -545,28 +539,52 @@ private Mono<Void> doUnlock(String name, ReactiveRedisConnection connection) {
...
+		}
+
+		private ReactiveRedisConnectionFactory getReactiveConnectionFactory() {
+			return (ReactiveRedisConnectionFactory) DefaultRedisCacheWriter.this.connectionFactory;
+		}
+
+		private Mono<ReactiveRedisConnection> getReactiveConnection() {
+			return Mono.fromSupplier(getReactiveConnectionFactory()::getReactiveConnection);
+		}
+
```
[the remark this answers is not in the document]
jxblum: I disagree. They help reduce the noise in the method where this code is clumped up, thereby causing the code formatter to break lines at ridiculous points. Additionally, using a named method is much clearer than some arbitrary variable, like `cf`.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -74,33 +73,21 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
...
 
 	/**
-	 * @param connectionFactory must not be {@literal null}.
-	 * @param sleepTime sleep time between lock request attempts. Must not be {@literal null}. Use {@link Duration#ZERO}
-	 *          to disable locking.
-	 * @param lockTtl Lock TTL function must not be {@literal null}.
-	 * @param cacheStatisticsCollector must not be {@literal null}.
-	 * @param batchStrategy must not be {@literal null}.
+	 * @param sleepTime sleep time between lock request attempts. Must not be {@literal null}.
+	 * Use {@link Duration#ZERO} to disable locking.
```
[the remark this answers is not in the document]
jxblum: The Spring Data source code formatter for Javadoc needs to be fixed.

The excessive indentation is causing unnatural line breaks (such as sentences ending in `and`), creating orphans, and doing other inconsistent things that do not align with the core Spring Frameworks Javadoc guidelines.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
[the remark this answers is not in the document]
jxblum: In these methods... 

With `waitForInterval(period)`, it is more descriptive than `Flux.interval(Duration.ZERO, period)`. What is `Duration.ZERO` anyway? Oh right, delay. Why do I even need to know it is `Flux` for that matter? It is an implementation detail. You don't know what `Object.wait(..)` does when you call it.

`getReactiveConnectionFactory()` avoid the cast could serve as an extension point if `protected`. In too many places, we are referring to state variables directly rather than the appropriate getter (where applicable), preventing extensibility.

`getReactiveConnection()` should go without saying, `Mono` into a `connectionFactory.getConnection()`. Having this logic spread all over the class is redundant and busy creating noise in that need wrapped connections.

What does it matter whether it is a variable or a method. Methods help remove clutter from other methods and can be reused.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum: Technically, I have even seen things likes this:

```java
return connection.stringCommands().set(ByteBuffer.wrap(key), ByteBuffer.wrap(value),
         Expiration.from(ttl.toMillis(), TimeUnit.MILLISECONDS), SetOption.upsert());
```

jxblum: `toExpiration(..)` is a easy to read method, `private`, and at the end of the class.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum: Just thinking... perhaps the class presence check should be on a Reactive type (e.g. `Mono`) then, given it does depend on the runtime verifier, particularly in strange cases where it can load the `ReactiveConnectionFactory` interface, which has no explicit Reactive types until you drill into `ReactiveConnection`. The real question is, the `ReactiveConnection` as a return type on `ReactiveConnectionFactory` enough?

Also, by this same argument, then the only way a `ReactiveConnectionFactory` can exist is if _Lettuce_ is on the classpath anyway, which has a hard dependency on Reactor. Of course, there is no guarantee that Lettuce will always pull in Reactor as a transitive dependency (hence [this](https://github.com/spring-projects/spring-boot/blob/v3.1.4/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisReactiveAutoConfiguration.java#L43)), but also the reason I argued for [this](https://vmware.slack.com/archives/C055WU98P/p1694723156015419) (internal link).

jxblum: I convinced it should be both `ReactiveConnectionFactory` and a Reactive type like `Mono` or `Flux`.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
jxblum: Additionally, these methods can be use as _Lambda's_, such as in `.map(:Function<S, T>)` functions, which are useful in the context of Optionals, Streams, etc.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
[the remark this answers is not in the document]
jxblum: What was there [before](https://github.com/spring-projects/spring-data-redis/pull/2744/files#diff-6c6aae73dbe5cee3bc00315ecfb74e08fb1606bc25c1abe443b2c14afea8e6c0L564-R587) was certainly less readable.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -74,33 +73,21 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
...
 	 */
 	DefaultRedisCacheWriter(RedisConnectionFactory connectionFactory, Duration sleepTime, BatchStrategy batchStrategy) {
 		this(connectionFactory, sleepTime, TtlFunction.persistent(), CacheStatisticsCollector.none(), batchStrategy);
 	}
 
 	/**
-	 * @param connectionFactory must not be {@literal null}.
```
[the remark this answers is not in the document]
jxblum: "must not be null" on `@param` tags is redundant and non-informative.

In general, I'd argue if we are not going to take the time to write meaningful Javadoc, then there is no point to have Javadoc on the method at all.

jxblum:
> Did a first review pass.

Thank you for the feedback. I reviewed each comment and cited my reasoning for the requested changes.

I reworked this PR to be more inline with our expectations.

Specifically, I:

1) Added a more robust check to the detection of Reactive types and usage, not unlike Spring Boot ([here](https://github.com/spring-projects/spring-boot/blob/v3.1.4/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisReactiveAutoConfiguration.java#L43)).

2) I reduced the number of `private` (helper) methods only slightly and reorganized them based on where they belong. Certain `private` methods remain because 1) they are reused 2) and/or are more descriptive (especially in the Reactive context).

3) Please! Can we follow the Spring Framework [Javadoc guidelines](https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc).

I updated this PR.

Issue #2745, opened 2023-10-18
Refine JedisConnectionFactory


jxblum:
General code refactoring and cleanup task triggered while considering the fallout and possible workaround for #2738.

Introduces private (static) methods (and/or local variables) to simplify and remove duplicate logic while simultaneously removing broken line breaks affecting readability and understanding.

> NOTE: line breaks are also being affected by our excessive indentation.

Simplifies assertions by making consistent use of Spring Frameworks `Assert` facility rather than unnecessary conditional blocks.

Refers to state variables using getter (rather than direct variable reference) where applicable helping to improve extensibility.

Reorganizes state variables in both `JedisConnectionFactory` as well as `MutableJedisClientConfiguration` by type in alphabetical order.

Organizes code according to Spring Data source code guidelines (e.g. private methods towards the end of classes).

Deprecates `createTopologyProvider(..)` in favor of `createClusterTopologyProvider(..)` for consistent and clarified naming.

Adds additional and judicious white space to improve readability.

PR #2746, opened 2023-10-18
Refine `JedisConnectionFactory`

jxblum:
See #2745

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
@@ -53,11 +53,24 @@ import org.springframework.dao.InvalidDataAccessResourceUsageException;
 import org.springframework.data.redis.ExceptionTranslationStrategy;
 import org.springframework.data.redis.PassThroughExceptionTranslationStrategy;
 import org.springframework.data.redis.RedisConnectionFailureException;
-import org.springframework.data.redis.connection.*;
+import org.springframework.data.redis.connection.ClusterCommandExecutor;
+import org.springframework.data.redis.connection.ClusterTopologyProvider;
+import org.springframework.data.redis.connection.RedisClusterConfiguration;
+import org.springframework.data.redis.connection.RedisClusterConnection;
+import org.springframework.data.redis.connection.RedisConfiguration;
 import org.springframework.data.redis.connection.RedisConfiguration.SentinelConfiguration;
 import org.springframework.data.redis.connection.RedisConfiguration.WithDatabaseIndex;
 import org.springframework.data.redis.connection.RedisConfiguration.WithPassword;
+import org.springframework.data.redis.connection.RedisConnection;
+import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.data.redis.connection.RedisNode;
+import org.springframework.data.redis.connection.RedisPassword;
+import org.springframework.data.redis.connection.RedisSentinelConfiguration;
+import org.springframework.data.redis.connection.RedisSentinelConnection;
+import org.springframework.data.redis.connection.RedisStandaloneConfiguration;
+import org.springframework.data.redis.connection.jedis.JedisClusterConnection.JedisClusterNodeResourceProvider;
 import org.springframework.data.redis.connection.jedis.JedisClusterConnection.JedisClusterTopologyProvider;
+import org.springframework.data.util.CastUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
@@ -92,6 +105,7 @@ import org.springframework.util.ObjectUtils;
  * @author Mark Paluch
  * @author Fu Jian
  * @author Ajith Kumar
+ * @author John Blum
  * @see JedisClientConfiguration
  * @see Jedis
  */
@@ -100,8 +114,7 @@ public class JedisConnectionFactory
 
 	private static final Log log = LogFactory.getLog(JedisConnectionFactory.class);
 
-	private static final ExceptionTranslationStrategy EXCEPTION_TRANSLATION = new PassThroughExceptionTranslationStrategy(
-			JedisExceptionConverter.INSTANCE);
+	private static final ExceptionTranslationStrategy EXCEPTION_TRANSLATION = jedisExceptionTranslationStrategy();
 
 	private boolean convertPipelineAndTxResults = true;
 
@@ -109,9 +122,9 @@ public class JedisConnectionFactory
 
 	private final AtomicReference<State> state = new AtomicReference<>(State.CREATED);
 
-	private @Nullable ClusterCommandExecutor clusterCommandExecutor;
+	private @Nullable AsyncTaskExecutor asyncExecutor;
 
-	private @Nullable AsyncTaskExecutor executor;
+	private @Nullable ClusterCommandExecutor clusterCommandExecutor;
 
 	private @Nullable ClusterTopologyProvider topologyProvider;
 
@@ -125,8 +138,7 @@ public class JedisConnectionFactory
 
 	private @Nullable RedisConfiguration configuration;
 
-	private RedisStandaloneConfiguration standaloneConfig = new RedisStandaloneConfiguration("localhost",
-			Protocol.DEFAULT_PORT);
+	private RedisStandaloneConfiguration standaloneConfig = defaultHostAndPortStandaloneConfiguration();
 
 	/**
 	 * Lifecycle state of this factory.
@@ -249,8 +261,7 @@ public class JedisConnectionFactory
 			@Nullable JedisPoolConfig poolConfig) {
 
 		this.configuration = sentinelConfiguration;
-		this.clientConfiguration = MutableJedisClientConfiguration
-				.create(poolConfig != null ? poolConfig : new JedisPoolConfig());
+		this.clientConfiguration = MutableJedisClientConfiguration.create(nullSafePoolConfig(poolConfig));
 	}
 
 	/**
@@ -281,11 +292,17 @@ public class JedisConnectionFactory
 		this.standaloneConfig = standaloneConfiguration;
 	}
 
+	private static RedisStandaloneConfiguration defaultHostAndPortStandaloneConfiguration() {
+		return new RedisStandaloneConfiguration("localhost", Protocol.DEFAULT_PORT);
+	}
+
+	private static PassThroughExceptionTranslationStrategy jedisExceptionTranslationStrategy() {
+		return new PassThroughExceptionTranslationStrategy(JedisExceptionConverter.INSTANCE);
+	}
+
 	ClusterCommandExecutor getRequiredClusterCommandExecutor() {
 
-		if (this.clusterCommandExecutor == null) {
-			throw new IllegalStateException("ClusterCommandExecutor not initialized");
-		}
+		Assert.state(this.clusterCommandExecutor != null, "ClusterCommandExecutor not initialized");
 
 		return this.clusterCommandExecutor;
 	}
@@ -300,7 +317,7 @@ public class JedisConnectionFactory
 
 		Assert.notNull(executor, "AsyncTaskExecutor must not be null");
 
-		this.executor = executor;
+		this.asyncExecutor = executor;
 	}
 
 	/**
@@ -354,15 +371,6 @@ public class JedisConnectionFactory
 	@Nullable
 	public String getPassword() {
 		return getRedisPassword().map(String::new).orElse(null);
-	}
-
-	@Nullable
-	private String getRedisUsername() {
-		return RedisConfiguration.getUsernameOrElse(this.configuration, standaloneConfig::getUsername);
-	}
-
-	private RedisPassword getRedisPassword() {
-		return RedisConfiguration.getPasswordOrElse(this.configuration, standaloneConfig::getPassword);
 	}
 
 	/**
@@ -452,9 +460,8 @@ public class JedisConnectionFactory
 	@Deprecated
 	public void setUsePool(boolean usePool) {
 
-		if (isRedisSentinelAware() && !usePool) {
-			throw new IllegalStateException("Jedis requires pooling for Redis Sentinel use");
-		}
+		Assert.state(!isRedisSentinelAware() || usePool,
+				"Jedis requires pooling for Redis Sentinel use");
 
 		getMutableConfiguration().setUsePooling(usePool);
 	}
@@ -466,7 +473,7 @@ public class JedisConnectionFactory
 	 */
 	@Nullable
 	public <T> GenericObjectPoolConfig<T> getPoolConfig() {
-		return clientConfiguration.getPoolConfig().orElse(null);
+		return getClientConfiguration().getPoolConfig().orElse(null);
 	}
 
 	/**
@@ -519,7 +526,7 @@ public class JedisConnectionFactory
 	 */
 	@Nullable
 	public String getClientName() {
-		return clientConfiguration.getClientName().orElse(null);
+		return getClientConfiguration().getClientName().orElse(null);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -125,8 +126,7 @@ public class JedisConnectionFactory
 
 	private @Nullable RedisConfiguration configuration;
 
-	private RedisStandaloneConfiguration standaloneConfig = new RedisStandaloneConfiguration("localhost",
-			Protocol.DEFAULT_PORT);
+	private RedisStandaloneConfiguration standaloneConfig = defaultStandaloneConfiguration();
```
mp911de, line 129: Introducing a indirection makes it difficult to understand at first sight the details.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -249,8 +249,7 @@ public JedisConnectionFactory(RedisSentinelConfiguration sentinelConfiguration,
 			@Nullable JedisPoolConfig poolConfig) {
 
 		this.configuration = sentinelConfiguration;
-		this.clientConfiguration = MutableJedisClientConfiguration
-				.create(poolConfig != null ? poolConfig : new JedisPoolConfig());
+		this.clientConfiguration = MutableJedisClientConfiguration.create(nullSafePoolConfig(poolConfig));
```
mp911de, line 252: Inflationary usage of `nullSafe…` methods creates a lot of noise hiding the actual details. Introducing indirections makes it difficult to understand at first sight the details.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -667,32 +665,29 @@ public void start() {
...
+				Pool<Jedis> jedisPool = this.pool;
+
+				if (jedisPool != null) {
 					try {
-						this.pool.close();
+						jedisPool.close();
 						this.pool = null;
-					} catch (Exception ex) {
-						log.warn("Cannot properly close Jedis pool", ex);
+					} catch (Exception cause) {
```
mp911de, line 689: To be precise: `cause` isn't true as the cause refers to the exception cause and not the caught exception. Please stick to `ex`.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -963,76 +993,136 @@ private Jedis getActiveSentinel() {
...
+
 	/**
 	 * Mutable implementation of {@link JedisClientConfiguration}.
 	 *
 	 * @author Mark Paluch
 	 */
+	@SuppressWarnings("rawtypes")
 	static class MutableJedisClientConfiguration implements JedisClientConfiguration {
 
-		private boolean useSsl;
```
mp911de, line 1022: What's the reason to tear code with good locality apart? All SSL-related fields have been kept together. Now, these are spread across.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -963,76 +993,136 @@ private Jedis getActiveSentinel() {
...
+				convertedNodes.add(newHostAndPort(node));
 			}
 		}
+
 		return convertedNodes;
 	}
 
-	private int getReadTimeout() {
-		return Math.toIntExact(clientConfiguration.getReadTimeout().toMillis());
+	private static HostAndPort newHostAndPort(String hostname, int port) {
```
mp911de, line 1037: Why do we have to introduce an abstraction over `new`? This causes noise and makes the code harder to read.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -963,76 +993,136 @@ private Jedis getActiveSentinel() {
 		throw new InvalidDataAccessResourceUsageException("No Sentinel found");
 	}
 
+	private boolean isCreatedOrStopped(@Nullable State state) {
```
mp911de, line 996: Instead of introducing a ton of new utility methods, we should rather improve our code by eliminating nullability for `State`.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -935,24 +961,28 @@ public RedisSentinelConnection getSentinelConnection() {
...
+				jedis = createJedis(node, clientConfig);
+
 				if (jedis.ping().equalsIgnoreCase("pong")) {
 					success = true;
 					return jedis;
 				}
-			} catch (Exception ex) {
-				log.warn(String.format("Ping failed for sentinel host: %s", node.getHost()), ex);
+			} catch (Exception cause) {
+				log.warn("Ping failed for sentinel host: %s".formatted(node.getHost()), cause);
```
mp911de, line 985: Please let's remain on `String.format(…)` to remain consitent. We can change all occurrences of `String.format(…)` to `formatted` if we feel that the value we get from that outweighs the effort.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -935,24 +961,28 @@ public RedisSentinelConnection getSentinelConnection() {
...
-				jedis = new Jedis(new HostAndPort(node.getHost(), node.getPort()), clientConfig);
+				jedis = createJedis(node, clientConfig);
+
 				if (jedis.ping().equalsIgnoreCase("pong")) {
 					success = true;
 					return jedis;
 				}
-			} catch (Exception ex) {
-				log.warn(String.format("Ping failed for sentinel host: %s", node.getHost()), ex);
+			} catch (Exception cause) {
```
mp911de, line 984: `cause` is technically wrong, please stick to `ex`

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -776,22 +768,58 @@ protected Pool<Jedis> createRedisSentinelPool(RedisSentinelConfiguration config)
...
 
+	/**
+	 * Template method to create a {@link ClusterTopologyProvider} with the given {@link JedisCluster}.
+	 * <p>
+	 * Creates {@link JedisClusterTopologyProvider} by default.
+	 *
+	 * @param cluster reference to the configured {@link JedisCluster} used by the {@link ClusterTopologyProvider}
+	 * to interact with the Redis cluster; must not be {@literal null}.
+	 * @return a new {@link ClusterTopologyProvider}.
+	 * @see org.springframework.data.redis.connection.jedis.JedisClusterConnection.JedisClusterTopologyProvider
```
mp911de, line 791: No need to create duplicate `@see` references to types that are already present in the signature.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -776,22 +768,58 @@ protected Pool<Jedis> createRedisSentinelPool(RedisSentinelConfiguration config)
...
+	 * @param cluster reference to the configured {@link JedisCluster} and {@link ClusterTopologyProvider} used to
+	 * execute Redis commands across the cluster; must not be {@literal null}.
+	 * @param clusterTopologyProvider {@link ClusterTopologyProvider} used to gather information about the current
+	 * Redis cluster topology.
+	 * @return a new {@link ClusterCommandExecutor}.
+	 * @see org.springframework.data.redis.connection.ClusterTopologyProvider
+	 * @see redis.clients.jedis.JedisCluster
+	 * @since 3.2
+	 */
+	protected ClusterCommandExecutor createClusterCommandExecutor(JedisCluster cluster,
```
mp911de, line 813: No need to create duplicate `@see` references to types that are already present in the signature.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -776,22 +768,58 @@ protected Pool<Jedis> createRedisSentinelPool(RedisSentinelConfiguration config)
...
-	 * {@link JedisClusterTopologyProvider} by default.
-	 *
-	 * @param cluster the {@link JedisCluster}, must not be {@literal null}.
-	 * @return the {@link ClusterTopologyProvider}.
-	 * @see JedisClusterTopologyProvider
-	 * @see 2.2
+	 * @deprecated Use {@link #createClusterTopologyProvider(JedisCluster)} instead.
+	 * @since 2.2
 	 */
+	@Deprecated(since = "3.2")
```
mp911de, line 778: I do not understand what warrants a deprecation of the `createTopologyProvider`. New deprecations are a signal for existing users to do something about their usage while we do not solve an actual problem.

mp911de (changes requested):
Did a first review pass. The majority of changes impose a different style, resulting in code that requires more effort to read (harder to read) and that isn't an improvement.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
```diff
@@ -100,18 +102,17 @@ public class JedisConnectionFactory
 
 	private static final Log log = LogFactory.getLog(JedisConnectionFactory.class);
 
-	private static final ExceptionTranslationStrategy EXCEPTION_TRANSLATION = new PassThroughExceptionTranslationStrategy(
-			JedisExceptionConverter.INSTANCE);
+	private static final ExceptionTranslationStrategy EXCEPTION_TRANSLATION = jedisExceptionTranslationStrategy();
```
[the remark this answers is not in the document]
jxblum: Avoiding line breaks, and details aren't needed until they are needed.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Same reason as above.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: The lines in question are too busy creating 1) an incomplete statement and 2) unnecessary line break.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: `cause` is the reason the error occurred in the block above. You are confusing this with caused by.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Organization by type.  All the state variables were bunched together making it difficult to read.

If you want better organization around concern, then perhaps a private, internal `SslConfiguration` class used by `MutableJedisClientConfiguration` makes more sense.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Naming consistency with other cluster-oriented types.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: `"some string with placeholders".formatted(..)` is less noisy

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Also, this is not unlike Spring Framework's own `nullSafeClassName(..)` ([Javadoc](https://docs.spring.io/spring-framework/docs/6.1.0-SNAPSHOT/javadoc-api/org/springframework/util/ObjectUtils.html#nullSafeClassName(java.lang.Object))) `nullSafeConciseToString(..)` ([Javadoc](https://docs.spring.io/spring-framework/docs/6.1.0-SNAPSHOT/javadoc-api/org/springframework/util/ObjectUtils.html#nullSafeConciseToString(java.lang.Object))) methods, or the countless other `nullSafe...` variants.

Your reference to "indirections" is a bit ironic.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: In any case, I am not going to argue about this.

If you want it to be consistent, then we should call it `e` everywhere Exceptions are caught (or `ignore` when Exceptions are not handled), which is what the core Spring Framework uses (i.e. `e`, not sure about `ignore`).

Using `ex` is not consistent, and `o_O` is not professional.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: No. It is the cause for the Exception being thrown in the block above.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Changed

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Changed

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: I was quite surprised when I first learned about Redis that we don't have something equivalent to `SentinelTopologyProvider` for Redis Sentinel.

Of course, Redis Sentinel is not a fully managed, "clustered" solution (only HA), but is a group of nodes in a distributed system none-the-less.

Technically, `ClusterTopology` is `Set` of `RedisClusterNode` and similarly, `RedisSentinelConfiguration` contains a `Set` of `RedisNode` where `RedisClusterNode` extends from `RedisNode`. So, although they are not quite the same, they are similar enough.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Reverted for the time being. This has room for improvement.

jxblum:
> Did a first review pass. The majority of changes impose a different style, resulting in code that requires more effort to read (harder to read) and that isn't an improvement.

I reworked this PR.

Please keep your comments objective and constructive.

If you re-read the commit message, you will notice several areas where the code was "improved". Cleaning up compiler warnings is a worthwhile improvement all by itself.

Finally, if you don't agree with the deprecation of the "`protected`" `createTopologyProvider(..)` (replaced by the new `createClusterTopologyProvider(..)` method). I don't have strong objections here. I just think it is better, more consistent from naming perspective.

Our naming across the API is inconsistent, sometimes referring to the same types, but with different names.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Case in point: https://github.com/spring-projects/spring-data-redis/blob/3.2.0-RC1/src/main/java/org/springframework/data/redis/core/script/DigestUtils.java#L63

src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java
jxblum: Now I see where we get our influence:

```java
package redis.clients.jedis;

public final class DefaultJedisClientConfig implements JedisClientConfig {

  private final RedisProtocol redisProtocol;

  private final int connectionTimeoutMillis;
  private final int socketTimeoutMillis;
  private final int blockingSocketTimeoutMillis;

  private volatile Supplier<RedisCredentials> credentialsProvider;
  private final int database;
  private final String clientName;

  private final boolean ssl;
  private final SSLSocketFactory sslSocketFactory;
  private final SSLParameters sslParameters;
  private final HostnameVerifier hostnameVerifier;

  //...
```

This is unfortunate. And, this is only marginally better.

Issue #2748, opened 2023-10-19
Refine all `catch` blocks to use consistent naming for `Exception` variables


jxblum:
This task consistently aligns all of Spring Data Redis's `catch` blocks / `Exception` handlers with the core Spring Framework's `Exception` (or `Throwable`) variable reference names, which is primarily (and nearly consistently)... `ex`.

Additionally, the core Spring Framework also uses `ignore` in places where the `Exception` (or `Throwable`) is ignored.

> NOTE: Unfortunately, only in a couple/three places (or so) did the core Spring Framework use inconsistent `Exception` variable names, such as `exc`. However, nearly all are named `ex`.

Now, in Spring Data Redis, `ex` and `ignore` are consistently and appropriately used everywhere. The are only 2 exceptions to this convention:

1) If the Exception handler is in a TEST class and "expected" , but not referenced or used in any way, then the "`expected`" keyword was used (only in TESTS).

2) In only 1 place (i.e. `ScanCursor`) was the `Exception` variable named "`nested`" due to `try-catch` block nesting, as in:

```java
try {
    // Exceptional code
} catch (Exception ex) {
    try {
        // More Exceptional code
    } catch (RuntimeException nested) {
        ex.addSuppressed(nested);
    }
}
```

In a few places, the Exception handler logic in `catch` blocks were refactored to be more readable, separating `Exception` message construction/formatting from the instantiation and throwing of the `Exception` itself.

Finally, `String.format(..)` was consistently applied to all `Exception` messages that require formatting.

I used the following REGEX expressions to search and modify our `Exception` blocks:

```txt
catch \(.*Exception e\)
catch \(.*Exception ex\)
catch \(.*Exception expected\)
catch \(.*Exception ignore\)
catch \(.*Exception [^ex|^ignore|^nested].*\)
```

PR #2749, opened 2023-10-19
Apply consistent `Exception` variable names to all `catch` blocks

jxblum:
See #2748.

It should also be NOTED that, in certain `catch (InterruptedException ex)` blocks, the `ex` Exception variable was not used or referenced. However, that does NOT mean the Exception handler did not "handle" the `InterruptedException`!

The simple invocation of `Thread.currentThread().interrupt()` IS an act of handling the `InterruptedException` and does have downstream consequences (or in other words, "side effects"). Therefore, the Exception variable name `ex` was deliberately kept in these cases.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
@@ -123,8 +123,8 @@ public class JedisClusterConnection implements RedisClusterConnection {
 			Object custerCommandExecutor = executorDfa.getPropertyValue("executor");
 			DirectFieldAccessor dfa = new DirectFieldAccessor(custerCommandExecutor);
 			clusterCommandExecutor.setMaxRedirects((Integer) dfa.getPropertyValue("maxRedirects"));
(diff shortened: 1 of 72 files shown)


src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
```diff
@@ -123,8 +123,8 @@ public JedisClusterConnection(JedisCluster cluster) {
 			Object custerCommandExecutor = executorDfa.getPropertyValue("executor");
 			DirectFieldAccessor dfa = new DirectFieldAccessor(custerCommandExecutor);
 			clusterCommandExecutor.setMaxRedirects((Integer) dfa.getPropertyValue("maxRedirects"));
-		} catch (Exception e) {
-			// ignore it and work with the executor default
+		} catch (Exception ignore) {
```
mp911de, line 126: 🥳

mp911de:
Looks pretty decent, thank you. It makes sense to backport the changes into 3.1.x as well so we simplify future backports.

mp911de:

Issue #2751, opened 2023-10-19
Replace use of `String.format(…)` with formatted Strings


jxblum:
This task replaces all uses of `String.format("..", args)` with `"..".formatted(args)` (formatted `Strings`).

From my perspective, the code has:

1) Much improved focus and visibility on the "_message_" being crafted for `Exceptions` or LOG messages FIRST rather than boilerplate code used to format the `String` message, and...

2) Reduced, unnatural line breaks in the source code, such as, but not limited to:

```java
    public void someMethod() {

        throw new IllegalArgumentException(String
                .format("some arbitrarily long message containing placeholders", argOne, argTwo, 
                        ..., argN);
    }
```

From my perspective, the above is not as readable as:

```java
    public void someMethod() {

        throw new IllegalArgumentException("some arbitrarily long message containing placeholders"
                .formatted(argOne, argTwo, ..., argN);
    }
```

This refactoring also reduced the use of the following technique to avoid weird, unusual and unnatural source code formatting by the code formatter:

```java
String message = String.format("...", args);
throw new IllegalArgumentException(message, ex);
```

Thereby keeping the "_message_" with the Exception.

Although, I must WARN that, in certain cases, the `Exception` "_messages_" was unusually long with placeholders spread throughout, such as:

```java
    public void someMethod() {

        throw new IllegalArgumentException(("some really, really loooooooooooooooooooooooooong"
                + " message containing multiple formatting placeholders"
                + " across multiple lines")
                .formatted(argOne, argTwo, ..., argN);
    }

```

In which case, when using `"..".formatted(args)` you need to wrap the concatenated message with parentheses, like so:

`("multiline, concatenated message, with placeholders").formatted(args)`

This is a minor trade-off, and I do not think it impairs the readability in any way.

Additionally, and fortunately, the IDE is smart enough to pick up on this and wrap the formatted String appropriately.

I considered the use of text block (or multi-line Strings) in Java, but you must be careful since the text block will preserve tabs and other (potentially) hidden characters (e.g. newlines) contained in the String.

All in all, even despite the necessary use of `()` in only a few places, I think the use of `"..".formatted(args)` is much cleaner, clearer to read, and more succinct than `String.format(..)`.

PR #2752, opened 2023-10-19
Replace `String.format(…, args)` with `"…".formatted(args)`

jxblum:
See #2751.

Please read the description in #2751.

I also think formatted Strings work nicer if natural line breaks are preserved.

For example:

```java
throw new IllegalArgumentException("some message with placeholders"
    .formatted(arg1, arg2, ..., argN);
```

Rather than this non-sense:

```java
throw new IllegalArgumentException("some message with placeholders".formatted(arg1, arg2, ...,
    argN);
```

That is orphans, or even partly broken argument lists.

Finally...

This is not strictly necessary. But, in my opinion, I think this is a nice improvement over our existing String formatting.

I also took the opportunity to clean up a (minimal) few compiler warnings while stumbling across message formatting.

Also, I changed a couple assertions that unnecessarily used a `Supplier` that did not need to and added `Suppliers` to a few assertions that should have had them.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/RedisSentinelConfiguration.java
+++ b/src/main/java/org/springframework/data/redis/connection/RedisSentinelConfiguration.java
@@ -223,7 +223,7 @@ public class RedisSentinelConfiguration implements RedisConfiguration, SentinelC
 	@Override
 	public void setDatabase(int index) {
 
-		Assert.isTrue(index >= 0, () -> String.format("Invalid DB index '%d'; non-negative index required", index));
+		Assert.isTrue(index >= 0, "Invalid DB index '%d'; non-negative index required".formatted(index));
 
 		this.database = index;
 	}
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
+	<version>3.2.0-GH-2751-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 	<description>Spring Data module for Redis</description>
--- a/src/main/java/org/springframework/data/redis/ClusterRedirectException.java
+++ b/src/main/java/org/springframework/data/redis/ClusterRedirectException.java
@@ -43,7 +43,7 @@ public class ClusterRedirectException extends DataRetrievalFailureException {
 	 */
 	public ClusterRedirectException(int slot, String targetHost, int targetPort, Throwable e) {
 
-		super(String.format("Redirect: slot %s to %s:%s.", slot, targetHost, targetPort), e);
+		super("Redirect: slot %s to %s:%s.".formatted(slot, targetHost, targetPort), e);
 
 		this.slot = slot;
 		this.host = targetHost;
--- a/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
@@ -382,9 +382,8 @@ class DefaultRedisCacheWriter implements RedisCacheWriter {
 			// Re-interrupt current Thread to allow other participants to react.
 			Thread.currentThread().interrupt();
 
-			String message = String.format("Interrupted while waiting to unlock cache %s", name);
-
-			throw new PessimisticLockingFailureException(message, ex);
+			throw new PessimisticLockingFailureException("Interrupted while waiting to unlock cache %s"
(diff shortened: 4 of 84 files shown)


src/main/java/org/springframework/data/redis/connection/RedisSentinelConfiguration.java
```diff
@@ -223,7 +223,7 @@ public int getDatabase() {
 	@Override
 	public void setDatabase(int index) {
 
-		Assert.isTrue(index >= 0, () -> String.format("Invalid DB index '%d'; non-negative index required", index));
```
kdebski85, line 226: I do not think this is a good change.
Without a supplier, we are needlessly constructing a string.
This assertion is almost always true, so it should be optimised for a happy path.
The same applies to other similar changes.

jxblum: Thank you for the feedback. I agree.

I wrote simple [JMH Benchmark](https://github.com/codeprimate-software/cp-labs/blob/main/src/main/java/org/cp/labs/java/SupplierCreationVsStringFormattingBenchmark.java#L49) today comparing `String.formatted(..)` with and without the use of a `Supplier` to the assertion, such as `isTrue(..)` above. It turns out that, without the `Supplier` the invocation performance (throughput & average time) is much worse over a large number of iterations, even despite the added Object allocations for `Suppliers`.  I did not look at the Heap memory map, but I suspect it will create a a bit more garbage than necessary, at least until JIT compiler optimizations kick in.

However, I will point out that, I was careful to only remove the `Suppliers` in non-critical code paths, such as in this case... "Configuration".

Configuration is only really ever in invoked on Spring container startup, once, when the beans are constructed, configured and initialized.

jxblum: I forget to mention, keep in mind that the `Supplier` Lambda in this case is NOT a non-capturing Lambda since it uses arguments from the enclosing method. Therefore, it, too, results in Object construction every time the method is called (though `RedisSentinelConfiguration.setDatabase(:int)` is only likely to be called once in this case).

Still, let's be clear here what we are talking about in terms of "optimizations".

You must always consider resource utilization when considering performance as often they are related. You must also measure. And finally, you must take into consideration the code path. Sacrificing readability, correctness or other factors for the sake of performance is not always the best decision.

jxblum: In reviewing my changes for this PR, along with writing a JMH Benchmark, I definitely think the use of a `Supplier` in assertions is necessary in SD Redis command classes, where the code path will be hotter at runtime than it will in configuration classes, for instance, such as [here](https://github.com/spring-projects/spring-data-redis/pull/2752/files#diff-868b06883aad7a48f4b3409cc46a399488849084497a79f2ca5db8fdb9d1b0d4R443), as well as in similar places.

In other cases, such as the [src/main/java/org/springframework/data/redis/core/types/RedisClientInfo.java](https://github.com/spring-projects/spring-data-redis/pull/2752/files#diff-41752625b4904fa3640910749360a5ee416a64a34958e026c24dcd1247cd6445) and [BinaryKeyspaceIdentifier](https://github.com/spring-projects/spring-data-redis/pull/2752/files#diff-e19cfb83e92d69907bb21398465ac8d6dfce633cb29d3b8b6e2ea0ef05ba9169R1301), I introduced the use of a `Supplier` inside the assertion where a `Supplier` was not used before.

jxblum: I have updated this PR to (re)introduce the use of the `Supplier` inside assertions. Only very few were inside critical codepaths, like command classes (specifically Jedis and Lettuce ZSet commands). The rest were in configuration classes (non-critical codepaths).

It should be noted, that most of the String formatting occurs inside Exception construction (e.g. `InvalidDataAccessApiUsageException`) and Logger statements. There is no option for using a `Supplier` in most of these cases, not without extra guards; for example:

```java
public void someMethod(Object... args) {
  //...
    logDebug(() -> 
        "Some formatted String with args [%s]".formatted(Arrays.toString(args));
}

private void logDebug(Supplier<String> logMessage) {
    if (logger.isDebugEnabled()) {
        logger.debug(logMessage.get());  
    }
}
```

This may make for some nice optimizations down the road (and in certain cases, I noticed I have introduced log helper methods, just like described above), but that is not a concern in this PR.

In conclusion, I still feel the appearance of using `String.formatted(..)` is much nicer than `Spring.format(..)`, as it calls attention to the Exception/Log/Whatever message immediately.

mp911de:

PR #2762, opened 2023-11-06
Make constant for repository module name

junghoon-vans:
In my opinion, returning the module name directly is not a good idea, 
so I modified it to handle it by declaring a constant instead.
Also, instead of returning a value directly from `getModulePrefix`, we can proxies it to `this.getModuleIdentifier.`


mp911de:
With a smarter defaulting in `KeyValueRepositoryConfigurationExtension` (spring-projects/spring-data-keyvalue#542) we were able to actually remove a bit of code. Thank you for your contribution, that's polished, merged, and backported now.

Issue #2766, opened 2023-11-10
CI documentation is outdated


junghoon-vans:
Currently, the source code as of the main branch requires a minimum of JDK 17, 
so the documentation provided there will also need to change.

PR #2767, opened 2023-11-10
Update CI.adoc to support JDK 17.

junghoon-vans:

closes: #2766

### DIFF
--- a/CI.adoc
+++ b/CI.adoc
@@ -16,7 +16,7 @@ All of these use cases are great reasons to essentially run what the CI server d
 
 IMPORTANT: To do this you must have Docker installed on your machine.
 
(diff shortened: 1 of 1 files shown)


mp911de:
I think we would want to remove the CI docs as it doesn't add much benefit for our community. For ourselves, we have documented the requirements in `Jenkinsfile` and the actual versions are stored in `pipeline.properties`.

junghoon-vans:
I think removing CI documentation completely makes it harder for community contributors to understand how to test locally.

mp911de:
Good point. I was under the impression that the readme mentions `make` but it is only showing `mvn clean install`. Let me update the docs for that aspect.

junghoon-vans:
Great, I see your fix.

PR #2768, opened 2023-11-11
Enhance right/leftPushAll(key, E...) to account for type-mismatched value arguments based on user-declared RedisTemplate<?, ?> type signature

Koooooo-7:
The `left/rightPushAll` has two methods.
- `public Long rightPushAll(K key, V... values) `
- `public Long rightPushAll(K key, Collection<V> values)`

---
In this case it may lead to an unexpected result when using a `RedisTemplate<String, Object> redisTemplate`. e.g.
```java
@AllArgsConstructor
public class Reproduce<T> {
    private final RedisTemplate<String, Object> redisTemplate;

    public void pushAll(List<T> value) {
        String key = "any";
        redisTemplate.opsForList().leftPushAll(key, value);
    }

}
```

It gonna use the first method `rightPushAll(K key, V... values)` instead of `rightPushAll(K key, Collection<V> values)` to save the List<T> as a single value.
image

---
I know that if we use `RedisTemplate<String, Object>`, the value type is all promote to Object , it treats all things as a single value, hence it uses the `rightPushAll(K key, V... values)`.

If I change to use 
```java
RedisTemplate<String, T> redisTemplate
``` 
or 
```java
redisTemplate.opsForList().leftPushAll(key, value.toArray(new Object[0]))
```
It should work as expected.

TBH, I'm not quite sure whether it should be a user code design issue or should be handled by this lib.
Personally, I wish it could be solved in here to make the `pushToAll` methods more closed to what it is real do on the `RedisTemplate<String, Object>` like other opts.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/DefaultListOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultListOperations.java
@@ -32,6 +32,7 @@ import org.springframework.util.CollectionUtils;
  * @author Thomas Darimont
  * @author Christoph Strobl
  * @author dengliming
+ * @author Koy Zhuang
  */
 class DefaultListOperations<K, V> extends AbstractOperations<K, V> implements ListOperations<K, V> {
 
--- a/src/main/java/org/springframework/data/redis/core/DefaultListOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultListOperations.java
@@ -17,6 +17,7 @@ package org.springframework.data.redis.core;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 import org.springframework.data.redis.connection.RedisConnection;
@@ -113,6 +114,12 @@ class DefaultListOperations<K, V> extends AbstractOperations<K, V> implements Li
 
 	@Override
 	public Long leftPushAll(K key, V... values) {
+		if (values.length == 1) {
+			V val = values[0];
+			if (values[0] instanceof Collection) {
+				return leftPushAll(key, (Collection) val);
+			}
+		}
 
 		byte[] rawKey = rawKey(key);
 		byte[][] rawValues = rawValues(values);
@@ -210,6 +217,12 @@ class DefaultListOperations<K, V> extends AbstractOperations<K, V> implements Li
 
 	@Override
 	public Long rightPushAll(K key, V... values) {
+		if (values.length == 1) {
+			V val = values[0];
+			if (values[0] instanceof Collection) {
+				return rightPushAll(key, (Collection) val);
+			}
+		}
 
 		byte[] rawKey = rawKey(key);
 		byte[][] rawValues = rawValues(values);
--- a/src/main/java/org/springframework/data/redis/core/DefaultListOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultListOperations.java
@@ -17,7 +17,6 @@ package org.springframework.data.redis.core;
 
 import java.util.Collection;
 import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 import org.springframework.data.redis.connection.RedisConnection;
--- a/src/test/java/org/springframework/data/redis/core/DefaultListOperationsUnitTests.java
+++ b/src/test/java/org/springframework/data/redis/core/DefaultListOperationsUnitTests.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2016-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.redis.core;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+import org.springframework.data.redis.connection.RedisConnection;
+import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.data.redis.serializer.JdkSerializationRedisSerializer;
+import org.springframework.data.redis.serializer.StringRedisSerializer;
+
+import java.util.List;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyCollection;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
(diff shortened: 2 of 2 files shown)


jxblum:
Hi @Koooooo-7 - Thank you for calling this to our attention.

I wrote a [simple test class](https://github.com/jxblum/redis-experiments/blob/spring-data-redis-pr-2768/spring-data-redis-experiments/src/test/java/io/vmware/spring/data/redis/tests/commands/list/RedisListOperationsIntegrationTests.java) today to the reproduce this issue and experiment with this area of code from SD Redis (i.e. `RedisTemplate` and `ListOperations` combination). As the [test case](https://github.com/jxblum/redis-experiments/blob/spring-data-redis-pr-2768/spring-data-redis-experiments/src/test/java/io/vmware/spring/data/redis/tests/commands/list/RedisListOperationsIntegrationTests.java#L63-L79) is currently setup, things work as expected.

Of course, if I use `StringRedisTemplate` along with `ListOperations<String, String>` and a `Collection<String>` declaration of "values" I want to push into the Redis list, then things also work as expected.

If I declare the following:

```java
@Autowired
@SuppressWarnings("rawtypes")
private RedisTemplate redisTemplate;
```

Along with:

```java
@Getter
@ToString
@EqualsAndHashCode
@RequiredArgsConstructor(staticName = "as")
static class User implements Serializable {
    private final String name;
}
```

Then this test also passes (works as expected):

```java
	@Test
	void redisListOperationsWorksCorrectly() {

		ListOperations<Object, User> listOperations = getRedisTemplate().opsForList();

		Collection<User> users = List.of(User.as("JonDoe"), User.as("JaneDoe"));

		assertThat(listOperations.size("TestList")).isZero();

		Long listSize = listOperations.rightPushAll("TestList", users);

		assertThat(listSize).isEqualTo(2L);

		assertThat(listOperations.rightPop("TestList", listSize))
			.containsExactly(User.as("JaneDoe"), User.as("JonDoe"));
		assertThat(listOperations.size("TestList")).isZero();
	}
```

I don't really think or feel it is the responsibility of any library or framework, such as Spring Data Redis, to adjust for nuances of a programming language, such as Java. Each JVM language is going possibly handle the treatment of generic type signatures differently.

In fact, generic type signatures in Java go through a process of type erasure at compile-time; i.e. generic type signatures are compile-time feature and the compiler determines which method to call based on the type signature of declared arguments (such as `Collection<?>`) passed to a method.

In conclusion, I feel it is application developers' responsibility to 1) use strongly-defined types and 2) responsibly declare their intentions, simply to remove any ambiguity so that when other developers who also look at the code, they have no doubt in their mind what technically happens at runtime.

Having conditional blocks all over the place to account for these variations is going to make SD Redis and similar libraries/framework harder to maintain with very little benefit for something that is easy to address from the application's perspective.

Koooooo-7:
Hi @jxblum , thx for your input.
I understand that If there does not use generic type,  things gonna work fine.
And I agree we should use more specific types for coding and reading.

About the erasures, IIUC, such as  `List<T>`  it gonna change to `List<Object>` at runtime/complier, it won't an issue for method parameter with `Collection` type signatures. (correct me if needs)

I think the point here is we have  both
```
public Long rightPushAll(K key, V... values) 
public Long rightPushAll(K key, Collection<V> values)
```
It leads to an ambiguous function call instead handle inside or compiler fatal.  i.e.
- With two methods
image

- Only `public Long rightPushAll(K key, Collection<V> values)`
image

- Only `Long rightPushAll(K key, V... values);`
image

Either one method could show the error when we call it in `type mismatch` instead of swallow it into a wrong call, thats why I wanna `swallow it into a right place`. or should we only keep one method there (the Collection one) ? 
 
No matter what the solution on this or not, I respect your works and thoughts on this.  thx again.

jxblum:
Hi @Koooooo-7,

Regarding:

> _About the erasures, IIUC, such as List<T> it gonna change to List<Object\> at runtime/complier, it won't an issue for method parameter with Collection type signatures._

Technically, the Java compiler will only generate the bytecode as `List`. An iteration over the elements in a generically-typed `List` depend on the [bounded] type of the elements at compile-time.

> TIP: This [article](https://medium.com/egnyte-engineering/getting-runtime-type-information-from-a-generic-class-in-java-d05d6854ba5b) containing an easy-to-follow example along with [Baeldung's blog post](https://www.baeldung.com/java-type-erasure) on Java Type Erasure explains this concept a bit better.

In your "_With two methods_" example and screenshot, unfortunately, the "_Multiple implementations_" tooltip covered up the type signature of the local `RedisTemplate` variable, yet a type signature was declared, given: `new RedisTemplate<>()` (i.e. `<>`).

For:

```java
@Autowired
private RedisTemplate<String, User> redisTemplate
```

Unfortunately, Spring Framework cannot auto-wire the auto-configured `RedisTemplate` provided by Spring Boot _auto-configuration_ (see [here](https://github.com/spring-projects/spring-boot/blob/v3.1.5/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfiguration.java#L58-L61)) since the type signatures between the provided instance and application declaration do not match for the specific use case, i.e. `RedisTemplate<String, User>`.

```txt
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.data.redis.core.RedisTemplate<java.lang.Object, io.vmware.spring.data.redis.tests.commands.list.RedisListOperationsIntegrationTests$User>' available: 
expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.raiseNoMatchingBeanFound(DefaultListableBeanFactory.java:1824)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:1383)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:1337)
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.resolveFieldValue(AutowiredAnnotationBeanPostProcessor.java:764)
	... 76 more
```

However, nothing prevents you from declaring your own `RedisTemplate` `@Bean` appropriately for your use case as follows:

```java
@SpringBootConfiguration
class MyRedisConfiguration {

    @Bean
    RedisTemplate<Object, User> userRedisTemplate(RedisConnectionFactory connectionFactory) {
        RedisTemplate<Object, User> userRedisTemplate = new RedisTemplate<>();
        userRedisTemplate.setConnectionFactory(connectionFactory);
        return userRedisTemplate;
    }
}
```

Then everything works as expected.

I play around with my test class a bit more, with the above bean definition/declaration, along with the following code snippets, things work fine:

```java
RedisTemplate<String, User> redisTemplate; // auto-wired in test class
ListOperations<String, User> listOperations = getRedisTemplate().opsForList();
List<User> users = List.of(User.as("JonDoe"), User.as("JaneDoe"));

Long listSize = listOperations.rightPushAll("UserList", users);
```

And:

```java
listSize = listOperations.leftPushAll("UserList", User.as("PieDoe"), User.as("SourDoe"));
```

Both calls (`leftPushAll(K key, V... values)` and `leftPushAll(K key, Collection<V> values)`) are used accordingly and correctly.

I checked in the latest changes to my `RedisListOperationsIntegrationTests` class as seen [here](https://github.com/jxblum/redis-experiments/blob/main/spring-data-redis-experiments/src/test/java/io/vmware/spring/data/redis/tests/commands/list/RedisListOperationsIntegrationTests.java#L63).

While the `RedisTemplate` and is Thread-safe and generally reusable (see [doc](https://docs.spring.io/spring-data/redis/docs/current/reference/html/#redis:template)) for multiple use cases, I would highly recommend, again, strongly-typed objects, such as `RedisTemplate<String, User>` in this example.

This is a Java language, and using the language appropriately, more than it is a framework concern to compensate for the language and (arguably) Java compiler deficiencies.

Koooooo-7:
Hi @jxblum  appreciate for your sufficient explanation! 
I fully agree that strongly-typed objects makes more sense on implementation instead of handling some crude usages in the lib now.

PR #2773, opened 2023-11-16
Support Pub/Sub for static master/replica configuration.

kdebski85:
Add support for Pub/Sub for static master/replica configuration and document that it supports publishing/subscribing only with one node.


triumph1:
Hi :)
looks like a good idea, any progress?

mp911de:
This would only work until a failover. I'm not quite sure that is a good idea given that expectations lean towards auto-adoption of infrastructure changes without the need of restarting applications.

PR #2774, opened 2023-11-17
Remove unnecessary method call in adapter.

junghoon-vans:
The `keyspace` is already a String type, so there is no need to call `asString`.


jxblum:
Hi @junghoon-vans,

It appears at one point (before DATAREDIS-664) the keyspace method parameter (and ID) where `Serializable`, then changed to a `String` in order to adapt to API changes in SD KeyValue.

Apparently, this was overlooked when the method signatures changed.

Thank you for catching this and contributing a PR.

PR #2807, opened 2023-12-14
Convert an Implementation of IndexedData to a Record.

### LINKED ISSUE #2806: Convert an implementation of IndexedData to a Record.
Record is Java's inconvenience data class, so it's a great fit for `IndexedData` for holding and passing data.
It's also nice not to have to define `getters` or methods like `hashcode` and `equals`, especially if it changes.

junghoon-vans:

Closes #2806


mp911de:
This is changing a lot of public API for no good reason therefore we cannot accept this change.

PR #2824, opened 2024-01-08
Update Jedis documentation to reflect supporting Redis Streams

imba-tjd:
I haven't used it. But according to https://github.com/spring-projects/spring-data-redis/issues/1711 I think jedis supports stream


mp911de:

PR #2825, opened 2024-01-09
Fix reference documentation for Stream ops usage

imba-tjd:
There isn't `template.streamOps`


mp911de:

PR #2846, opened 2024-02-15
Update Converters.java

lightistor:
Clean extra character in Cluster Node Line Entry for Google Cloud Platform Memory Store Redis when using the Jedis Connector. Sample entries causing ClusterNode information does not define host and port: 3765733728631672640db35fd2f04743c03119c6 10.180.0.33:11003@16379, master - 0 1708041426947 2 connected 5462-10922

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
+++ b/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
@@ -63,6 +63,7 @@ import org.springframework.util.StringUtils;
  * @author daihuabin
  * @author John Blum
  * @author Sorokin Evgeniy
+ * @author Habip Kenan Uskudar
  */
 public abstract class Converters {
 
@@ -569,8 +570,9 @@ public abstract class Converters {
 
 		@Override
 		public RedisClusterNode convert(String source) {
-
-			String[] args = source.split(" ");
+			String filteredSource = source.replace(", ", " ");
+				
+			String[] args = filteredSource.split(" ");
 
 			Matcher matcher = clusterEndpointPattern.matcher(args[HOST_PORT_INDEX]);
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,6 +40,26 @@ pipeline {
 						}
 					}
 				}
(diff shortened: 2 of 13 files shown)


lightistor:
@christophstrobl and @mp911de this is a fix to handle the format of the Cluster Node Line Entry for Google Cloud Platform Memory Store Redis that we are integrating our services with.
is it possible for me get assistance in merging this as a nightly build so that it is publicly available as opposed to us having have to host this fork internally?

mp911de:
Please provide a test case so that we document what isn't working right now for you and what should work. Also, `source..replace` is a syntax error introduced with this PR.

lightistor:
> Please provide a test case so that we document what isn't working right now for you and what should work. Also, `source..replace` is a syntax error introduced with this PR.

fixed the typo, added the unit test (toSetOfRedisClusterNodesShouldConvertNodesWithEmptyHumanReadableClusterNameInGcpMemoryStoreCorrectly, uses the entry format CLUSTER_NODE_ENTRY_WITH_EMPTY_HUMAN_READABLE_CLUSTER_NAME_IN_GCP_MEMORY_STORE)
and added myself to the authors in the header (per the checklist)

lightistor:
So, what is the possibility of it getting merged one day @mp911de ?

src/main/java/org/springframework/data/redis/connection/convert/Converters.java
```diff
@@ -569,8 +570,9 @@ enum ClusterNodesConverter implements Converter<String, RedisClusterNode> {
 
 		@Override
 		public RedisClusterNode convert(String source) {
-
-			String[] args = source.split(" ");
+			String filteredSource = source.replace(", ", " ");
```
mp911de, line 573: That won't do the job especially when auxilliary fields are appended, then the `port` part is immediately followed by aux fields.

mp911de:
This PR isn't going anywhere really and the proposed fix is not addressing the underlying issue. As it would require a lot of effort from our side to fix and address shortcomings, we're instead closing this PR without a merge.

PR #2850, opened 2024-02-17
Fix quoted variable reference in documentation

1autodidact:
Overview
This PR addresses a small issue in the loadHash method of the HashMapping class where the hardcoded string "key" was used instead of the parameter key when retrieving a hash from Redis.

Description
In the loadHash method of the HashMapping class, the hardcoded string "key" was used to retrieve the hash from Redis instead of utilizing the parameter key provided to the method. This PR corrects this issue by modifying the method call to use the provided key parameter, ensuring the correct hash is loaded.

Changes Made
Updated the loadHash method in the HashMapping class to use the provided key parameter when retrieving the hash from Redis, replacing the hardcoded string "key". Testing
No new functionality is introduced with this change, and existing unit tests cover the modified method. Tests have been run locally to ensure that the correct hash is loaded using the provided key parameter.

Request for Review
I kindly request a review of this PR. Your feedback is greatly appreciated. Thank you for your time and consideration.

### DIFF
--- a/src/main/antora/modules/ROOT/pages/redis/hash-mappers.adoc
+++ b/src/main/antora/modules/ROOT/pages/redis/hash-mappers.adoc
@@ -44,7 +44,7 @@ public class HashMapping {
 
   public Person loadHash(String key) {
 
-    Map<byte[], byte[]> loadedHash = hashOperations.entries("key");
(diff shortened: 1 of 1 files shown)


christophstrobl:
LGTM - thanks for spotting!

mp911de:

PR #2868, opened 2024-03-13
Allow RedisConnectionFactories to be initialized as part of the context lifecycle

### LINKED ISSUE #2866: Allow RedisConnectionFactories to be initialized as part of the context lifecycle
In the current arrangement the `RedisConnectionFactory` implementations are implementing both, `InitializingBean` and `SmartLifecycle` (See also #2502). Though, for historic reasons they eagerly try to connect to the Redis server already in `afterPropertiesSet()` which is not in line with the actual intent of the method aiming to verify properties leaving the startup to `Lifecycle#start()`.

For Checkpoint restore and CDS training runs we'd need to move this arrangement into its desired state without breaking existing applications potentially relying on the eager initialization of the connection factory.

mp911de:
Lettuce and Jedis connection factories now can be configured to initialize early during afterPropertiesSet or configured whether the component should be auto-started by the container.

By default, connection factories auto-startup early.

Closes #2866

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>3.3.0-SNAPSHOT</version>
+	<version>3.3.0-GH-2866-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 	<description>Spring Data module for Redis</description>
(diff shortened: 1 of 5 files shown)


christophstrobl:
while trying to add a section about startup/lifecycle considerations for connection factories to the reference documentation I was wondering if setting `autoStartup` to `false` without also disabling `earlyStartup` does make that much sense?

PR #2900, opened 2024-04-24
Use extended `SET` instead of deprecated setEx, pSetEx and setNX commands

ehsanalemzadeh:
Replace usage of deprecated commands setEx, pSetEx and setNX in DefaultValueOperations by set command with additional SetOption arguments

Closes ##2897


mp911de:

PR #2051, opened 2021-04-21
Revise RedisCache and documentation

mp911de:
* Add support for configurable `BatchStrategy` to cleanup the cache via `KEYS` or `SCAN` (supersedes #547, closes #1721)
* Add support for binary keys in `ScanOptions` (closes #2006)
* Mention RedisCache locking behavior details (closes #1801)

PR #2904, opened 2024-05-11
DATAREDIS-1151 - change clean method keys command to scan

bert82503:
`clean` method use `keys` command to get all keys which is blocked, may cause other commands timeout or block.
As the complexity of DEL is O(N), if the number of keys is huge, DEL command could timeout as well. 
Using DEL command is very risky in a production environment, right?

Use `scan` command as default `BatchStrategy` in `RedisCacheWriter`, for without locking behavior or with locking behavior.
We use this strategy in the production environment.
[Use RedisCacheWriter#clean to asynchronously scan matches online batch scan to delete cached data for spring-data-redis](https://juejin.cn/spost/7313242056742354994)

- https://jira.spring.io/browse/DATAREDIS-1151
- https://github.com/spring-projects/spring-data-redis/pull/532
- https://github.com/spring-projects/spring-data-redis/pull/2051

> Here are my own thoughts:
> Why do so many users raise issues and report that they encounter timeouts and blocks in use?
> I'm wondering how we can improve the ease of use of good products?

### DIFF
--- a/src/main/java/org/springframework/data/redis/cache/RedisCacheWriter.java
+++ b/src/main/java/org/springframework/data/redis/cache/RedisCacheWriter.java
@@ -47,7 +47,7 @@ public interface RedisCacheWriter extends CacheStatisticsProvider {
 	 * @return new instance of {@link DefaultRedisCacheWriter}.
 	 */
 	static RedisCacheWriter nonLockingRedisCacheWriter(RedisConnectionFactory connectionFactory) {
-		return nonLockingRedisCacheWriter(connectionFactory, BatchStrategies.keys());
+		return nonLockingRedisCacheWriter(connectionFactory, BatchStrategies.scan(10));
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


bert82503:
Hi, bro.
Can you help to check this?  thx!  @mp911de

bert82503:
> We cannot do that. Jedis Cluster doesn't support cross-cluster `SCAN` by reusing a given `cursor` so we would break functionality for Redis Cluster users.

Okay, thank you for explaining the specific reasons.

bert82503:
> We cannot do that. Jedis Cluster doesn't support cross-cluster `SCAN` by reusing a given `cursor` so we would break functionality for Redis Cluster users.

The `KEYS` batch strategy is fully supported using any driver and Redis operation mode (Standalone, Clustered). `SCAN` is fully supported when using the Lettuce driver. Jedis supports `SCAN` only in non-clustered modes.

[Spring Data Redis](https://docs.spring.io/spring-data/redis/reference/index.html) / 
[Redis](https://docs.spring.io/spring-data/redis/reference/redis.html) / 
[Redis Cache](https://docs.spring.io/spring-data/redis/reference/redis/redis-cache.html)

image

PR #2905, opened 2024-05-11
feat: add setter method to define custom NullValueSerializer

### LINKED ISSUE #2878: Support custom `NullValueSerializer` in `GenericJackson2JsonRedisSerializer`
If one wants a custom NullValueSerializer you can define your own [`registerNullValueSerializer`](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java#L227-L232), you cannot define your own [`NullValueSerializer`](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java#L366)

but because `registerNullValueSerializer` is static, you have to also overwrite other methods like the [base constructors](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java#L84-L125).

so how can one easily define their own nullValueSerializer? 
the use case in particular i am using it as redis is used as the backend for `@cachable` and i want to cache the fact that i have a null value for some keys. the default null that the redis serializer uses is `jsonGenerator.writeStringField(classIdentifier, NullValue.class.getName());`, which produces a weird byte array that when i try to deserialize it (access the cache later) it gives me an error.

And then semi-related, if i overwrite the `deserialize` method to return null if i see the special null identifier, `@cacheable` considers it a cache miss. but i actually really do want to return the value null from my cache!!

AnneMayor:

resolved #2878


src/test/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializerUnitTests.java
```diff
@@ -420,7 +423,7 @@ void deserializesJavaTimeFrimBytes() {
 	}
 
 	@Test // GH-2601
-	public void internalObjectMapperCustomization() {
```
AnneMayor, line 423: I just removed according to sonarlint plugin. It looks like an unnecessary keyword.

src/test/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializerUnitTests.java
```diff
@@ -438,14 +441,34 @@ public void internalObjectMapperCustomization() {
 	}
 
 	@Test // GH-2601
-	public void configureWithNullConsumerThrowsIllegalArgumentException() {
```
AnneMayor, line 441: I just removed according to sonarlint plugin. It looks like an unnecessary keyword.

AnneMayor:
I'm looking for another great solution and trying to apply it.

AnneMayor:
Please, review it. Thanks :)

AnneMayor:
@mp911de 
Thank you for the detailed review! I think that's a great idea too. I am gonna create a new Builder class as you suggested within this weekend:) Thanks 👍

AnneMayor:
@mp911de 
I created Builder class. Please, review it :D

src/test/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializerUnitTests.java
```diff
@@ -432,20 +436,63 @@ public void internalObjectMapperCustomization() {
 
 		assertThat(serializer.configure(configurer)).isSameAs(serializer);
 
-		verify(mockObjectMapper, times(1)).registerModule(eq(mockModule));
```
AnneMayor, line 435: I just removed unnecessary command.

injae-kim:
Polishing commit https://github.com/spring-projects/spring-data-redis/commit/8357c7d88b4e7ee6d8bb73f1a38e602c11ff758c is really awsome 👍 

Thank you for kind&rapid review!

AnneMayor:
@mp911de ,
Thank you so much for your kind and rapid review 😊
I am very happy to contribute this open source.
I am going to look into your polishing commit and learn from it.
Again, thanks a lot 👍

PR #2918, opened 2024-05-26
Add multiple reactive keys exist checker

### LINKED ISSUE #2883: `ReactiveKeyCommands.Exists` to check multiple key existence
I am using the function to check if a key exist or not, but actually I have a list of keys that needs to be checked. I can't find a func that would accept a list of keys and do this as a bulk check. Is there a reason why this was not implemented? if not, would that be possible to implement it? it is already implemented in [jedis](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/Jedis.java#L535).

I have achieved this by the following implementation:

```
 redisOperations.execute { connection ->
                val keyCommands = connection.keyCommands()
                val existsMonos = keys.map { key ->
                    keyCommands.exists(ByteBuffer.wrap(key))
                        .map { it == true }
                        .onErrorReturn(false)
                }
                Mono.zip(existsMonos) { it }
            }.awaitFirst()
```

`redisOperations` is a `ReactiveRedisOperations` type.

The versions:

`org.springframework.data:spring-data-redis:3.2.2`
`java 21.0.2-librca`
`Kotlin compiler version 1.9.22`

Thank you

AnneMayor:

Resolved #2883

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/ReactiveKeyCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/ReactiveKeyCommands.java
@@ -175,6 +175,14 @@ public interface ReactiveKeyCommands {
 
 		return exists(Mono.just(new KeyCommand(key))).next().map(BooleanResponse::getOutput);
 	}
+	/**
+	 * Determine the number of given {@literal keys} that exist.
+	 *
+	 * @param keys must not be {@literal null} or {@literal empty}.
+	 * @return {@link Mono} emitting {@literal the number of existing keys}.
+	 * @see <a href="https://redis.io/docs/commands/exists/">Redis Documentation: EXISTS</a>
+	 */
+	Mono<Long> exists(List<ByteBuffer> keys);
 
 	/**
 	 * Determine if given {@literal key} exists.
--- a/src/test/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveKeyCommandsIntegrationTests.java
+++ b/src/test/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveKeyCommandsIntegrationTests.java
@@ -66,6 +66,22 @@ public class LettuceReactiveKeyCommandsIntegrationTests extends LettuceReactiveC
 		connection.keyCommands().exists(KEY_1_BBUFFER).as(StepVerifier::create).expectNext(false).verifyComplete();
 	}
 
+	@ParameterizedRedisTest
+	void existsKeyReturnsKeyCount() {
+		nativeCommands.set(KEY_1, "1000");
+		nativeCommands.set(KEY_2, "2000");
+		nativeCommands.set(KEY_3, "3000");
+
+		connection.keyCommands().exists(List.of(KEY_1_BBUFFER, KEY_2_BBUFFER, KEY_3_BBUFFER)).as(StepVerifier::create)
+				.expectNext(3L).verifyComplete();
+	}
+
+	@ParameterizedRedisTest
+	void existsKeyReturnsZeroWhenKeyDoesNotExist() {
+		connection.keyCommands().exists(List.of(KEY_1_BBUFFER, KEY_2_BBUFFER, KEY_3_BBUFFER)).as(StepVerifier::create)
+				.expectNext(0L).verifyComplete();
+	}
+
 	@ParameterizedRedisTest // DATAREDIS-525
 	void typeShouldReturnTypeCorrectly() {
 
@@ -164,7 +180,7 @@ public class LettuceReactiveKeyCommandsIntegrationTests extends LettuceReactiveC
 		connection.keyCommands().rename(KEY_1_BBUFFER, KEY_2_BBUFFER).as(StepVerifier::create).expectNext(true)
 				.verifyComplete();
 		assertThat(nativeCommands.exists(KEY_2)).isEqualTo(1L);
-		assertThat(nativeCommands.exists(KEY_1)).isEqualTo(0L);
+		assertThat(nativeCommands.exists(KEY_1)).isZero();
 	}
 
 	@ParameterizedRedisTest // DATAREDIS-525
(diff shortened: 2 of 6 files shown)


src/test/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveKeyCommandsIntegrationTests.java
```diff
@@ -164,7 +184,7 @@ void renameShouldAlterKeyNameCorrectly() {
 		connection.keyCommands().rename(KEY_1_BBUFFER, KEY_2_BBUFFER).as(StepVerifier::create).expectNext(true)
 				.verifyComplete();
 		assertThat(nativeCommands.exists(KEY_2)).isEqualTo(1L);
-		assertThat(nativeCommands.exists(KEY_1)).isEqualTo(0L);
+		assertThat(nativeCommands.exists(KEY_1)).isZero();
```
AnneMayor, line 187: I just changed the command to make better readability.

src/main/java/org/springframework/data/redis/connection/ReactiveKeyCommands.java
```diff
@@ -175,6 +174,20 @@ default Mono<Boolean> exists(ByteBuffer key) {
...
 		return exists(Mono.just(new KeyCommand(key))).next().map(BooleanResponse::getOutput);
 	}
+	/**
+	 * Determine if given all {@literal keys} exist.
+	 *
+	 * @param keys must not be {@literal null} or {@literal empty}.
+	 * @return {@link Mono} emitting {@literal true} if all keys exist.
+	 * @see <a href="https://redis.io/commands/exists">Redis Documentation: EXISTS</a>
+	 */
+	default Mono<Boolean> exists(List<ByteBuffer> keys) {
```
AnneMayor, line 184: I created a validation method which is for checking if parameterized all keys exist in redis cache. If one of them does not exist, it returns false.
@kutlueren , please check if this function is what you meant to use.

kutlueren: Hey @AnneMayor 

Awesome to see that you added the function which carries out what I suggested. It is actually more straight forward than I initially thought, maybe I got confused in the codebase but nice! I shall try it out as soon as it gets merged.

kutlueren:
Hey @AnneMayor, I just wanted to follow up on the topic :) Is this being planned to be merged soon?

AnneMayor:
Hey @kutlueren ,
Unfortunately I have no feedback from our maintainer🥲
I am still waiting for their feedback.

kutlueren:
> Hey @kutlueren , Unfortunately I have no feedback from our maintainer🥲 I am still waiting for their feedback.

@mp911de would you mind taking a look at this one when you have time?

AnneMayor:
@mp911de , 
Please let me know if this PR has some issues🙏

src/main/java/org/springframework/data/redis/connection/ReactiveKeyCommands.java
mp911de: This implementation calls `EXISTS` for each key issuing a command for each element in the list.

Redis' [`EXISTS`](https://redis.io/docs/latest/commands/exists/) command accepts multiple arguments and returns how many of the keys exist. Our interface should use the existing functionality instead of putting another layer on top of commands with a worse performance behavior.

The issue we have here is however that `exists(Publisher<KeyCommand> keys)` and `exists(Publisher<List<KeyCommand>> keys)` or `exists(Publisher<MultiKeyCommand>> keys)` erase to the same type and we cannot simply introduce such an override.

I suggest to implement `Mono<Long> exists(List<ByteBuffer> keys)` directly instead of using `default` methods. Also, `ReactiveRedisOperations` should be extended with `countExistingKeys(…)` similar to `RedisTemplate.countExistingKeys(…)`.

mp911de (changes requested):
I did a review pass and left a comment regarding the design and implementation. Care to have a look?

AnneMayor:
> I did a review pass and left a comment regarding the design and implementation. Care to have a look?

Thanks a lot. I am going to apply your review within this weekend 👍

dreamstar-enterprises:
My current workaround, since the spring API does not support exists with multiple keys yet. 

```
// Prepare session keys
                    val sessionKeys = sessionIds.map { sessionId ->
                        "${sessionProperties.redis?.sessionNamespace}:sessions:$sessionId"
                    }

                    // First read the hash keys to trigger expiration event
                    val script = """
                       local keys = ARGV
                       local count = 0
                       for i, key in ipairs(keys) do
                           count = count + redis.call('EXISTS', key)
                       end
                       return count
                    """

                    redisOperations.execute(
                        RedisScript.of(script, Long::class.java),
                        listOf(),  // KEYS list is empty
                        sessionKeys
                    )
                    .doOnNext { count ->
                        logger.debug("Accessed {} session hash keys", count)
                    }
                    .flatMap { _ ->
                        redisOperations.opsForZSet()
                            .remove(redisKeyExpirations, *sessionIds.toTypedArray())
                            .doOnSuccess { logger.debug("Successfully removed {} expired sessions", sessionIds.size) }
                            .doOnError { e -> logger.error("Error during removal: ${e.message}") }
                    }
                    .then()
```

AnneMayor:
@dreamstar-enterprises Oh, for real? I am gonna check. Sorry for the late response. Since I have some business, I ended up keeping going this PR again. Will review your issue and give some feedback.

AnneMayor:
Sorry for being late. I have lots of personal business and I am going to resolve this PR comments until this weekend. Please, keep that in mind.

AnneMayor:
@mp911de could you review this PR again for me?
Also, I have a question. Do you think that it would be better if I add this exist method into class `ReactiveRedisOperations`?

mp911de:
No worries, some pull requests remain open for quite some time because we have to prioritize sometimes other things and so sometimes, pull requests don't get much attention.

In any case, thank you for your contribution. That's merged and polished now.

PR #2936, opened 2024-07-07
Add overloads for XAddOptions in StreamOperations

### LINKED ISSUE #2915: Add overloads to `StreamOperations` accepting `XAddOptions`
HI

jinkshower:
Closes #2915

After this change, the client can use `XAddOptions` as an additional parameter for `opsForStream.add()`

TO-BE
```java
XAddOptions options = XAddOptions.maxlen(1).approximateTrimming(false);
redisTemplate.opsForStream.add("myStream", Map.of("key", "value), options);
```

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/ReactiveStreamCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/ReactiveStreamCommands.java
@@ -32,6 +32,7 @@ import org.springframework.data.redis.connection.ReactiveRedisConnection.Command
 import org.springframework.data.redis.connection.ReactiveRedisConnection.KeyCommand;
 import org.springframework.data.redis.connection.ReactiveRedisConnection.NumericResponse;
 import org.springframework.data.redis.connection.RedisStreamCommands.XClaimOptions;
+import org.springframework.data.redis.connection.RedisStreamCommands.XAddOptions;
 import org.springframework.data.redis.connection.RedisStreamCommands.XPendingOptions;
 import org.springframework.data.redis.connection.stream.ByteBufferRecord;
 import org.springframework.data.redis.connection.stream.Consumer;
@@ -58,6 +59,7 @@ import org.springframework.util.StringUtils;
  * @author Tugdual Grall
  * @author Dengliming
  * @author Mark John Moreno
+ * @author jinkshower
  * @since 2.2
  */
 public interface ReactiveStreamCommands {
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/data/redis/connection/ReactiveStreamCommands.java
```diff
@@ -394,11 +396,40 @@ default Mono<RecordId> xAdd(ByteBufferRecord record) {
...
+		}
+
+		return xAdd(Mono.just(addStreamRecord)).next().map(CommandResponse::getOutput);
+	}
+
 	/**
 	 * Add stream record with given {@literal body} to {@literal key}.
 	 *
 	 * @param commands must not be {@literal null}.
-	 * @return {@link Flux} emitting the {@link RecordId} on by for for the given {@link AddStreamRecord} commands.
```
jinkshower, line 401: removed typo.

src/main/java/org/springframework/data/redis/connection/ReactiveStreamCommands.java
```diff
@@ -394,11 +396,40 @@ default Mono<RecordId> xAdd(ByteBufferRecord record) {
...
+		if (xAddOptions.hasMaxlen()) {
+			addStreamRecord = addStreamRecord.maxlen(xAddOptions.getMaxlen());
+		}
+
+		if (xAddOptions.hasMinId()) {
+			addStreamRecord = addStreamRecord.minId(xAddOptions.getMinId());
+		}
+
+		return xAdd(Mono.just(addStreamRecord)).next().map(CommandResponse::getOutput);
+	}
```
jinkshower, line 426: the reason why I wrote somewhat verbose mehtod in here is

unlike `JedisStreamCommands` or `LettuceStreamCommands` 
which both have a method accepting (MapRecord<byte[], byte[], byte[]> record, XAddOptions options)

xAdd(Publisher<AddStreamRecord> commands) in `LettuceReactiveStreamCommands`
expects that Publisher<AddStreamRecord> already has variables of `XAddOptions`. 
ex) maxlen, isApproxiateTrimming...

so I placed the decapsulation of XAddOptions here.

I thought of changing AddStreamRecord to have variable `priavte final XAddOptions options`
so that `LettuceReactiveStreamCommands` can delegate the decapsulation to other class
but I thought this exceeds the scope of this issue.

jinkshower (commented):
If you kindly review this pr and need some changes, I will be happy to amend it!

src/main/java/org/springframework/data/redis/connection/ReactiveStreamCommands.java
```diff
@@ -394,11 +396,40 @@ default Mono<RecordId> xAdd(ByteBufferRecord record) {
...
+
 	/**
 	 * Add stream record with given {@literal body} to {@literal key}.
 	 *
 	 * @param commands must not be {@literal null}.
-	 * @return {@link Flux} emitting the {@link RecordId} on by for for the given {@link AddStreamRecord} commands.
+	 * @return {@link Flux} emitting the {@link RecordId} on by for the given {@link AddStreamRecord} commands.
```
jinkshower, line 432: removed typo

jinkshower:
I have a question regarding XAddOptions. While I was writing tests, I found out that 
```java
XAddOptions options = XAddOptions.maxlen(-1);
redisTemplate.opsForStream("myStream", Map.of("key", "value), options);
```
this executes and the result being maxlen option just not applied. 

This is becasue in `XAddOptions` in `RedisStreamCommands`'s hasMaxlen() method.

```java
public boolean hasMaxlen() {
	return maxlen != null && maxlen > 0;
}
```
also same with `AddStreamRecord` in `ReactiveStreamCommands`. 
this method just ignores minus value of maxlen. 

As for redis, if user makes maxlen -1, An error occurs.
image

If we make hasMaxlen()
```java
public boolean hasMaxlen() {
	return maxlen != null 
}
```

image
For Lettuce, Lettuce’s XAddArgs has LettuceAssert which asserts the value of maxlen. 

image
For Jedis, Jedis’s XAddParams doesn’t have assertions. So redis server throws an exception.

IMO, it is better to alert user by throwing exception than the result just not applied because it is hard to find and also realise the problem. 

The reason why I'm asking this question, instead incorporating into PR I created was I was not sure if this is intended.

If this needs to be fixed, I will make adjustment in this PR and push it again. If not, please let me know. Thank you.

kkocel:
@jinkshower looks good to me :)

jinkshower:
Hello, @christophstrobl. Could you please take some time to review this PR when you have a moment? I'm not entirely confident about some of the changes I've made, and your feedback would be incredibly valuable to me.

christophstrobl:
@jinkshower thanks for the heads up.

christophstrobl:
@jinkshower thank you for your contribution. Merged to _main_ development line.
Regarding the maxlen please open a new issue.

PR #2955, opened 2024-08-02
Do not delete phantom if exists and ShadowCopy is OFF

### LINKED ISSUE #2954: If ShadowCopy is OFF do not use the phantom key
Phantom key should be used only if ShadowCopy is ON or DEFAULT.
If set OFF, allow the client code to use that name.

ltorje1:
Closes #2954


christophstrobl:
Thank you for your contribution @ltorje1. Merged to _main_ development line va 3e80819.

Issue #2967, opened 2024-08-14
Refactor Tests to `instanceof` pattern variable


arefbehboudi:
In some of our test cases, we are currently using the traditional instanceof checks followed by explicit type casting. With the introduction of Pattern Matching in recent Java versions, we can refactor these checks to make the code more concise and readable.

**Example:**
  `
  if (redisOperations instanceof DisposableBean) {
	  ((DisposableBean) redisOperations).destroy();
  }
  `

Tasks:
  - Locate instanceof checks in tests.
  - Refactor them to use Pattern Matching.
  - Ensure all tests pass after changes.

If this approach is approved, I will proceed with the changes and create a pull request.

mp911de:
Hey @arefbehboudi thanks for your suggestion. It makes sense to go over the entire codebase and use `instanceof` pattern matching. Feel free to submit a pull request.

arefbehboudi:
@mp911de Thank you very much. I'll make the changes and keep you updated.

PR #2971, opened 2024-08-19
Refactor Tests to `instanceof` pattern variable

arefbehboudi:
In some of our test cases, we are currently using the traditional instanceof checks followed by explicit type casting. With the introduction of Pattern Matching in recent Java versions, we can refactor these checks to make the code more concise and readable.

[issues](https://github.com/spring-projects/spring-data-redis/issues/2967)


mp911de:

PR #2989, opened 2024-09-11
Use value object for topology caching

### LINKED ISSUE #2986: Race condition in `JedisClusterTopologyProvider`
JedisClusterTopologyProvider use a cache object to optimize performace on frequent cluster requests, but the time value is updated before the cache object, there is a race condition that the old cache including the invalid cluster topology might get returned, which will result in a ClusterCommandExecutionFailureException: Could not get a resource from th pool.

		public ClusterTopology getTopology() {

			if (cached != null && shouldUseCachedValue()) {
				return cached;
			}

			Map<String, Exception> errors = new LinkedHashMap<>();

			List<Entry<String, ConnectionPool>> list = new ArrayList<>(cluster.getClusterNodes().entrySet());

			Collections.shuffle(list);

			for (Entry<String, ConnectionPool> entry : list) {

				try (Connection connection = entry.getValue().getResource()) {

					time = System.currentTimeMillis();  // time value is updated before cached object

					Set<RedisClusterNode> nodes = Converters.toSetOfRedisClusterNodes(new Jedis(connection).clusterNodes());

					cached = new ClusterTopology(nodes);

					return cached;

				} catch (Exception ex) {
					errors.put(entry.getKey(), ex);
				}
			}

			StringBuilder stringBuilder = new StringBuilder();

			for (Entry<String, Exception> entry : errors.entrySet()) {
				stringBuilder.append(String.format("\r\n\t- %s failed: %s", entry.getKey(), entry.getValue().getMessage()));
			}

			throw new ClusterStateFailureException(
					"Could not retrieve cluster information; CLUSTER NODES returned with error" + stringBuilder);
		}

mp911de:
We now use a value object for caching the topology to avoid races in updating the cache timestamp.

Also, we set the cache timestamp after obtaining the topology to avoid that I/O latency expires the topology cache.

Closes #2986

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
@@ -805,13 +805,11 @@ public class JedisClusterConnection implements RedisClusterConnection {
 	 */
 	public static class JedisClusterTopologyProvider implements ClusterTopologyProvider {
 
-		private long time = 0;
+		private final JedisCluster cluster;
 
 		private final long cacheTimeMs;
 
-		private @Nullable ClusterTopology cached;
-
-		private final JedisCluster cluster;
+		private volatile @Nullable JedisClusterTopology cached;
 
 		/**
 		 * Create new {@link JedisClusterTopologyProvider}. Uses a default cache timeout of 100 milliseconds.
@@ -842,12 +840,12 @@ public class JedisClusterConnection implements RedisClusterConnection {
 		@Override
 		public ClusterTopology getTopology() {
 
-			if (cached != null && shouldUseCachedValue()) {
-				return cached;
+			JedisClusterTopology topology = cached;
+			if (shouldUseCachedValue(topology)) {
+				return topology;
 			}
 
 			Map<String, Exception> errors = new LinkedHashMap<>();
-
 			List<Entry<String, ConnectionPool>> list = new ArrayList<>(cluster.getClusterNodes().entrySet());
 
 			Collections.shuffle(list);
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
```diff
@@ -842,12 +840,12 @@ public JedisClusterTopologyProvider(JedisCluster cluster, Duration cacheTimeout)
 		@Override
 		public ClusterTopology getTopology() {
 
-			if (cached != null && shouldUseCachedValue()) {
-				return cached;
+			JedisClusterTopology topology = cached;
+			if (shouldUseCachedValue(topology)) {
+				return topology;
```
christophstrobl, line 845: the previous `shouldUseCachedValue()` would no longer be called in the patch targeting `3.3.4` so I was wondering if it makes senes to keep the arrangement stable for the SR and changing it for the `3.4` version.
For 3.3.4 we could have 
```java
protected boolean shouldUseCachedValue() {
	return shouldUseCachedValue(cached);
}
```
and call the old method.

src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
```diff
@@ -887,9 +882,38 @@ public ClusterTopology getTopology() {
...
+		 * Returns whether {@link #getTopology()} should return the cached {@link JedisClusterTopology}. Uses a time-based
+		 * caching.
+		 *
+		 * @return {@literal true} to use the cached {@link ClusterTopology}; {@literal false} to fetch a new cluster
+		 *         topology.
+		 * @see #JedisClusterTopologyProvider(JedisCluster, Duration)
+		 * @since 3.3.4
+		 */
+		protected boolean shouldUseCachedValue(@Nullable JedisClusterTopology topology) {
+			return topology != null && topology.getTime() + cacheTimeMs > System.currentTimeMillis();
```
christophstrobl, line 902: we could move the `cacheTimeMs` to the `JedisClusterTopology` having a package private method `maxTime` that does the calculation.

src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
mp911de: Feel free to apply these changes in the merge.

Issue #2928, opened 2024-06-15
RedisNode creation from bare hostname should assign default port


lukasz-glapinski-exa:
Current code is expecting node information to not contain port but is not assigning default port to portString variable resulting in exception being thrown.

https://github.com/spring-projects/spring-data-redis/blob/a7d39147cf3f9649b0609a10fc43e5672c732193/src/main/java/org/springframework/data/redis/connection/RedisNode.java#L97

LeeHyungGeol:
May I handle this PR? I'll create PR within tomorrow.

LeeHyungGeol:
Fix PR: #3002

PR #3002, opened 2024-09-22
`RedisNode` creation from bare hostname assigns default port

LeeHyungGeol:
Fix. #2928 

### Summary

This issue arises in the code responsible for separating the host and port from a hostAndPort string, specifically when the string represents an IPv6 address without square brackets or a bare hostname where no port is provided, causing exceptions.

To resolve this, I referenced the ConversionUnitTests test code to assign the characters following the last colon to portString in the case of an IPv6 address without square brackets. For bare hostnames, I used the DEFAULT_PORT from spring-data-redis and assigned it to portString when no port was provided.

https://github.com/spring-projects/spring-data-redis/blob/main/src/test/java/org/springframework/data/redis/connection/convert/ConvertersUnitTests.java#L243


src/test/java/org/springframework/data/redis/connection/RedisNodeUnitTests.java
```diff
@@ -0,0 +1 @@
...
        assertThat(node.getPort()).isEqualTo(7334);
    }

    @Test // Bare hostname without port
    void shouldParseBareHostnameWithoutPort() {
        RedisNode node = RedisNode.fromString("my.redis.server");
        assertThat(node.getHost()).isEqualTo("my.redis.server");
        assertThat(node.getPort()).isEqualTo(DEFAULT_PORT);
    }
}
```
mp911de, line 1: I'm not sure what has happened here. In any case, this test is formatted to not be readable in any way.

LeeHyungGeol: @mp911de 
I apologize for the issue with the code formatting. I have identified the problem and am currently making corrections. I suspect that the cause of the formatting issue is related to code linting. Therefore, I checked the Spring Data Code Formatting Settings on this page: https://github.com/spring-projects/spring-data-build/tree/main/etc/ide and I am applying those settings.

I am currently trying to apply eclipse-java-google-style.xml as the Eclipse code formatter. Do I need any other formatter?

LeeHyungGeol: @mp911de 
I have fixed the issue with the code formatting. I would appreciate it if you could review the PR again.

PR #3013, opened 2024-10-10
Update RedisMessageListenerContainer to subscribe all keyspace notifications

crazyoyo:
Hello：

According to Redis official statement about keyspace notifications, "Every node of a Redis cluster generates events about its own subset of the keyspace as described above. However, unlike regular Pub/Sub communication in a cluster, events' notifications are not broadcasted to all nodes." But Redis drivers, such as Jedis, randomly obtains a connection from a node of the redis cluster to subscribe events, including keyspace notifications. Which means that many events are missed from the Redis cluster.

_**Summary:**_
This PR improves the eventuallyPerformSubscription() method in RedisMessageListenerContainer by first determine whether it is a keyspace notifications pattern, then traverse all master nodes, and create event subscriber one by one to ensure that no keyspace notifications events are missed.

**_Proposed solution:_**
This PR improves the eventuallyPerformSubscription() method in RedisMessageListenerContainer by subscribe all master nodes to make sure no  keyspace notifications events are missed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -122,7 +122,7 @@ pipeline {
 			steps {
 				script {
 					docker.withRegistry(p['docker.proxy.registry'], p['docker.proxy.credentials']) {
-						docker.image("springci/spring-data-with-redis-6.2:${p['java.main.tag']}").inside('-v $HOME:/tmp/jenkins-home') {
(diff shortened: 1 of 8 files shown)


mp911de:
Thanks for looking into this. We're not going to merge this PR as the changes pull in a lot of driver dependencies into the listening container. We do not want such a design, instead, a design where we could benefit from abstractions would be much better.

PR #3017, opened 2024-10-10
Adds support for Redis setGet command

### LINKED ISSUE #2853: Add StringCommand for a SET with GET option
It would be great to have a method in RedisStringCommands that would allow a `SET` command with `GET` option, which is available since Redis 6.2.0 according to the documentation of Redis. Furthermore, I would like to use it with NX option, like `SET key value NX GET EX seconds`, which is available since 7.0.0

marcingrzejszczak:
fixes #2853

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/ReactiveStringCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/ReactiveStringCommands.java
@@ -47,6 +47,7 @@ import org.springframework.util.Assert;
  *
  * @author Christoph Strobl
  * @author Mark Paluch
+ * @author Marcin Grzejszczak
  * @since 2.0
  */
 public interface ReactiveStringCommands {
@@ -62,15 +63,23 @@ public interface ReactiveStringCommands {
 		private @Nullable ByteBuffer value;
 		private Expiration expiration;
 		private SetOption option;
+		private boolean get;
 
 		private SetCommand(ByteBuffer key, @Nullable ByteBuffer value, @Nullable Expiration expiration,
 				@Nullable SetOption option) {
+
+			this(key, value, expiration, option, false);
+		}
+
+		private SetCommand(ByteBuffer key, @Nullable ByteBuffer value, @Nullable Expiration expiration,
+				@Nullable SetOption option, boolean get) {
 
 			super(key);
 
 			this.value = value;
 			this.expiration = expiration;
 			this.option = option;
+			this.get = get;
 		}
 
 		/**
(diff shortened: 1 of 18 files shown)


src/main/java/org/springframework/data/redis/connection/ReactiveStringCommands.java
```diff
@@ -62,15 +63,23 @@ class SetCommand extends KeyCommand {
 		private @Nullable ByteBuffer value;
 		private Expiration expiration;
 		private SetOption option;
+		private boolean get;
```
mp911de, line 66: That isn't going to work as the `set` API returns a boolean response.

src/main/java/org/springframework/data/redis/connection/ReactiveStringCommands.java
```diff
@@ -184,6 +210,22 @@ default Mono<Boolean> set(ByteBuffer key, ByteBuffer value, Expiration expiratio
...
+	 * @param key must not be {@literal null}.
+	 * @param value must not be {@literal null}.
+	 * @param expiration must not be {@literal null}. Use {@link Expiration#persistent()} for no expiration time or
+	 *          {@link Expiration#keepTtl()} to keep the existing.
+	 * @param option must not be {@literal null}.
+	 * @return
+	 * @see <a href="https://redis.io/commands/set">Redis Documentation: SET</a>
+	 * @since 3.4
+	 */
+	Flux<ByteBufferResponse<SetCommand>> setGet(ByteBuffer key, ByteBuffer value, Expiration expiration, SetOption option);
```
mp911de, line 227: Here we need both signatures, `Mono<ByteBuffer> setGet(ByteBuffer key, ByteBuffer value, Expiration expiration, SetOption option) {` and `Flux<ByteBufferResponse<SetCommand>> set(Publisher<SetCommand> commands);`

The first one delegates to the latter.

src/main/java/org/springframework/data/redis/core/ValueOperations.java
```diff
@@ -44,6 +45,33 @@ public interface ValueOperations<K, V> {
...
+	 * stored at key is not a string.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param value must not be {@literal null}.
+	 * @param timeout the key expiration timeout.
+	 * @param unit must not be {@literal null}.
+	 * @see <a href="https://redis.io/commands/set">Redis Documentation: SET</a>
+	 * @since 3.4
+	 */
+	V setGet(K key, V value, long timeout, TimeUnit unit);
```
mp911de, line 60: You might want to update `BoundValueOperations` (the interface) and `ReactiveValueOperations` as well.

mp911de (changes requested):
Some reactive and interface bits are missing. Care to have a look?

marcingrzejszczak:
Changes applied

PR #3054, opened 2024-11-25
Add support for Hash Field Expiration

tishun:
This change introduces the [Hash Field Expiration](https://github.com/redis/redis/pull/13303) feature as part of the features provided by the spring-data-redis project.

As part of the change the following commands would be made available:

- [x] HEXPIRE - expire a set of hash fields after a specified time, measured in seconds
- [x] HPEXPIRE - expire a set of hash fields after a specified time, measured in milliseconds
- [x] HEXPIREAT - expire a set of hash fields at a specified [UNIX time](https://en.wikipedia.org/wiki/Unix_time), measured in seconds since Unix epoch
- [x] HPEXPIREAT - expire a set of hash fields at a specified [UNIX time](https://en.wikipedia.org/wiki/Unix_time), measured in milliseconds since Unix epoch
- [ ] HEXPIRETIME _(omitted on the grounds it is missing for keys too)_
- [x] HPERSIST - persist a field that was previously marked for expiration
- [x] HTTL - return the time-to-live of a given hash field in seconds
- [x] HPTTL - return the time-to-live of a given hash field in milliseconds

The new commands are available as part of the following interfaces:
_(based on similar PR #283)_
- RedisHashCommands
- HashOperations / BoundHashOperations
- ReactiveHashOperations
- StringRedisConnection
- RedisMap

This feature is available starting from [Redis CE version 7.4.x](https://redis.io/blog/announcing-redis-community-edition-and-redis-stack-74/) and later

For more information on HFE you can also check out ...
... the [blog article](https://redis.io/blog/hash-field-expiration-architecture-and-benchmarks/#:~:text=Just%20like%20the%20EXPIRE%20command,and%20remove%20expired%20hash%20fields.) on the topic.
... examples of how one can [take advantage of the feature](https://redis.io/kb/doc/29bj4ewcad/how-can-i-take-advantage-of-the-expiration-of-individual-hash-fields) in their application

---

---

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/ReactiveHashCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/ReactiveHashCommands.java
@@ -19,6 +19,8 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.nio.ByteBuffer;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -44,9 +46,33 @@ import org.springframework.util.Assert;
  *
  * @author Christoph Strobl
  * @author Mark Paluch
+ * @author Tihomir Mateev
  * @since 2.0
  */
 public interface ReactiveHashCommands {
+
+	/**
+	 * {@link Command} for hash-bound operations.
+	 *
+	 * @author Christoph Strobl
+	 * @author Tihomir Mateev
+	 */
+	class HashFieldsCommand extends KeyCommand {
+
+		private final List<ByteBuffer> fields;
+
+		private HashFieldsCommand(@Nullable ByteBuffer key, List<ByteBuffer> fields) {
+			super(key);
+			this.fields = fields;
+		}
+
+		/**
+		 * @return never {@literal null}.
+		 */
+		public List<ByteBuffer> getFields() {
+			return fields;
+		}
+	}
 
 	/**
 	 * {@literal HSET} {@link Command}.
@@ -216,15 +242,10 @@ public interface ReactiveHashCommands {
 	 * @author Christoph Strobl
 	 * @see <a href="https://redis.io/commands/hget">Redis Documentation: HGET</a>
 	 */
-	class HGetCommand extends KeyCommand {
-
-		private List<ByteBuffer> fields;
+	class HGetCommand extends HashFieldsCommand {
 
 		private HGetCommand(@Nullable ByteBuffer key, List<ByteBuffer> fields) {
-
-			super(key);
-
-			this.fields = fields;
+			super(key, fields);
 		}
 
 		/**
@@ -263,14 +284,7 @@ public interface ReactiveHashCommands {
 
 			Assert.notNull(key, "Key must not be null");
 
-			return new HGetCommand(key, fields);
-		}
-
-		/**
-		 * @return never {@literal null}.
-		 */
-		public List<ByteBuffer> getFields() {
-			return fields;
+			return new HGetCommand(key, getFields());
 		}
 	}
 
@@ -394,15 +408,10 @@ public interface ReactiveHashCommands {
 	 * @author Christoph Strobl
 	 * @see <a href="https://redis.io/commands/hdel">Redis Documentation: HDEL</a>
 	 */
-	class HDelCommand extends KeyCommand {
-
-		private final List<ByteBuffer> fields;
+	class HDelCommand extends HashFieldsCommand {
 
 		private HDelCommand(@Nullable ByteBuffer key, List<ByteBuffer> fields) {
-
-			super(key);
-
-			this.fields = fields;
+			super(key, fields);
 		}
 
 		/**
@@ -441,14 +450,7 @@ public interface ReactiveHashCommands {
 
 			Assert.notNull(key, "Key must not be null");
 
-			return new HDelCommand(key, fields);
-		}
-
-		/**
-		 * @return never {@literal null}.
-		 */
-		public List<ByteBuffer> getFields() {
-			return fields;
+			return new HDelCommand(key, getFields());
 		}
 	}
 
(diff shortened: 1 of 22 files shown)


leonovdmitry:
hi @tishun !

Is it any updates about your PR? i guess it`s wating for you to take action?

tishun:
> hi @tishun !
> 
> Is it any updates about your PR? i guess it`s wating for you to take action?

Hey @leonovdmitry ,
Happy to see there is some interest in the change.

The current agreement with the Pivotal team is for them to try and review and push this change with the 3.5 release. Mid-January was a tentative and nonbinding ETA that I hope we will be able to meet to start work on approving this.

In the mean time if you feel there are use cases that we've missed please let me know.

I have a follow-up commit for the `RedisMap` that I hope to push today or tomorrow.

leonovdmitry:
> > hi @tishun !
> > Is it any updates about your PR? i guess it`s wating for you to take action?
> 
> Hey @leonovdmitry , Happy to see there is some interest in the change.
> 
> The current agreement with the Pivotal team is for them to try and review and push this change with the 3.5 release. Mid-January was a tentative and nonbinding ETA that I hope we will be able to meet to start work on approving this.
> 
> In the mean time if you feel there are use cases that we've missed please let me know.
> 
> I have a follow-up commit for the `RedisMap` that I hope to push today or tomorrow.

@tishun  thank you for feedback, yep, i was a bit surprised not to find this function in the current library version, but happy to see PR, so thanks for your work.

About use cases, I don't know if this could be part of the core, but it would be convenient to have a method for get/set ttl for only one field with the result in the form of a response code(Long) and not a list of codes(List<Long>)

Good luck and have a nice day!

src/main/java/org/springframework/data/redis/connection/ReactiveHashCommands.java
```diff
@@ -842,4 +844,453 @@ default Mono<Long> hStrLen(ByteBuffer key, ByteBuffer field) {
...
+	/**
+	 * Expire a given {@literal field} after a given {@link Duration} of time, measured in milliseconds, has passed.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param field must not be {@literal null}.
+	 * @param duration must not be {@literal null}.
+	 * @return a {@link Mono} emitting the expiration result - {@code 2} indicating the specific field is deleted
+	 *         already due to expiration, or provided expiry interval is 0; {@code 1} indicating expiration time is set/updated;
+	 *         {@code 0} indicating the expiration time is not set (a provided NX | XX | GT | LT condition is not met);
+	 * 	       {@code -2} indicating there is no such field; {@literal null} when used in pipeline / transaction.
```
mp911de, line 909: `null` when pipelining/transaction in progress doesn't apply to reactive commands as we do not have dedicated pipelining nor transaction support. See also other commands as well.

mp911de (changes requested):
Thanks for your pull request. There are some minor things to be updated. Can you please squash your commits and force-push these along with a [DCO sign-off](https://wiki.linuxfoundation.org/dco) in the commit message?

src/main/java/org/springframework/data/redis/connection/ReactiveHashCommands.java
tishun: Sure, will address accordingly

tishun:
> Thanks for your pull request. There are some minor things to be updated. Can you please squash your commits and force-push these along with a [DCO sign-off](https://wiki.linuxfoundation.org/dco) in the commit message?

Addressed the review comments and added the signoff.

Also added changes to the RedisMap, please have a look at them so we do not miss them

christophstrobl:
Thank you again for the PR @tishun.

From what I've seen so far things look pretty good, still I think there are a couple of things we need to take care of before we can actually bring the change in.

It would be nice if:

- all hash command (imperative/reactive/string) interfaces share an equivalent feature set and implement `httl`, `hpttl`, ...
- `hpttl` would be issued when requesting ttl with msec precision
- we use either a `long, TimeUnit` or `Duration` method signature - maybe both make sense and we put in some default methods delegating `Duration` to `long, TimeUnit`
- `XX`, `NX`, `GT`, `LT` options could be passed on. Maybe `Expiration` can help with that to avoid overloaded methods
- `Reactive-`/`HashOperations` interfaces have an equivalent featureset for ttl operations
- the return type of commands keeps reference to its input, so that users invoking eg. `httl` are not forced to memorize the input fields to map them back to the actual values returned by the command
- reactive methods which require output to be processed in order would, by declaration, lift the burden of paying close attention to enforced sequential mapping from the consuming side.

tishun:
> Thank you again for the PR @tishun.
> 

Hey @christophstrobl , thanks for spending time and looking at this change!

> From what I've seen so far things look pretty good, still I think there are a couple of things we need to take care of before we can actually bring the change in.
> 

Sure! Could you please confirm my understanding of your comments (below)? 

> It would be nice if:
> 
> * all hash command (imperative/reactive/string) interfaces share an equivalent feature set and implement `httl`, `hpttl`, ...

Apologies, I may need some elaboration on `share an equivalent feature set`. I see I've missed the `hpttl` command in some of the interfaces, I will add it, is there something else you meant?

If there are some differences in the naming of the API methods it is because I was aiming at consistency with the existing codebase, for example `BoundHashOperations.getExpire()` is similar to `BoundKeyOperations.getExpire()`, while `RedisHashCommands.hTtl()` is similar to `RedisKeyCommands.ttl()`. All of these end up calling the HTTL command.

> * `hpttl` would be issued when requesting ttl with msec precision

It is, actually we always are always executing `HPTTL` even for seconds precision. I was thinking that the `HPTTL` command could be used in both cases for simplification, but having spent some more time thinking on that it is not correct behavior so I will address that.

> * we use either a `long, TimeUnit` or `Duration` method signature - maybe both make sense and we put in some default methods delegating `Duration` to `long, TimeUnit`

@mp911de mentioned (weird, I cant see this comment now, is it because I force pushed?) that the `long, TimeUnit` was an old way of doing Duration and asked that I change all usages to Duration, as this is a new API and we need not maintain two different ways to handle that. Same comment was made about using `Instant` instead of `Date` .

I seem to use TimeUnit only for httl (since I combined them both with HPTTL). When I make the change from the previous point I will completely remove the use of `TimeUnit` and only stick with Duration.

Does that make sense?

> * `XX`, `NX`, `GT`, `LT` options could be passed on. Maybe `Expiration` can help with that to avoid overloaded methods

I omitted that because it was missing for key expiration too. If you think it is important I will add it to the existing APIs.
It would end up with some inconsistency, but I don't think it is a big problem in this case, as long as you are okay with it.

> * `Reactive-`/`HashOperations` interfaces have an equivalent featureset for ttl operations

Oh, I seem to have missed the entire ReactiveHashOperations interface. So many of them. Will address this too.

> * the return type of commands keeps reference to its input, so that users invoking eg. `httl` are not forced to memorize the input fields to map them back to the actual values returned by the command

Well ... this would be a rather huge inconsistency if I am getting this right. 
None of the existing methods that accept multiple fields actually return the fields (or key) as a result, e.g. 

```java
class HashOperations {
   ...
   Long delete(H key, Object... hashKeys);

   List<HV> multiGet(H key, Collection<HK> hashKeys);
   ... 
```

Did you mean we should, instead of returning a List of results, rather return a Map of <Field, Result>?

> * reactive methods which require output to be processed in order would, by declaration, lift the burden of paying close attention to enforced sequential mapping from the consuming side.

... and not do the above for Reactive methods?

christophstrobl:
@tishun thanks for getting back to us. No worries about the reactive part, yes there are a lot of places changes need to go in. We already started working on this one based on your PR to have it on _main_  in the near future. Some of the proposed changes are already done. So no need to take action right now - I'll drop a message here once we have something for you to take a look at/comment on. 

I agree that for the `*commands` we should stay close to the Redis API. The `*operations` should offer a higher level view on the response. If you want to have the raw thing, it's always possible to fall back to the commands interface.

Sorry, I did not get the `delete` reference or yours. There's no other value than the number of affected entries, so...
`multiget` has been there for 10+ years but we'd likely not do it this way if it was to be added in 2025.

tishun:
> @tishun thanks for getting back to us. No worries about the reactive part, yes there are a lot of places changes need to go in. We already started working on this one based on your PR to have it on _main_ in the near future. Some of the proposed changes are already done. So no need to take action right now - I'll drop a message here once we have something for you to take a look at/comment on.

Oh, awesome, so your team will handle the changes you suggested?  Let me know if I can help in any way!

> Sorry, I did not get the `delete` reference or yours. There's no other value than the number of affected entries, so... `multiget` has been there for 10+ years but we'd likely not do it this way if it was to be added in 2025.

`.delete()` was a poor example, sorry about that. I get your point on `multiget()` - makes sense that we improve the API even if it is a bit inconsistent with the existing one.

mp911de:
Looking at expiration control over multiple keys, I wonder whether it made sense to introduce a fluent-API like interface (e.g. `BoundExpirations`) to channel interaction with expirations.
Bildschirmfoto 2025-02-18 um 09 13 48

The screenshot captures IDE experience, therefore I decided for a screenshot instead of code.


Thoughts?

mp911de:
FWIW, we should include type hints in `LettuceConnection.TypeHints` to cover `HEXPIRE` and friends return types.

mp911de:
I've added some last changes to wrap this change up (documentation, polishing, tests) at https://github.com/spring-projects/spring-data-redis/tree/issue/3054. We plan to merge this PR in the next days for inclusion in the next milestone.

mp911de:

tishun:
Oh @mp911de , FWIW I think we need to change the `@since` JavaDoc annotations, because I speculated we are targeting 3.4, but it seems HFE support is scheduled for 3.5?

mp911de:
This has already happened in 5228bc700895c38d4286d5cc1dcc7901c97be822. I'm currently catching up with conditional key expiry, see #3114. I suggest you having a look at the commits referenced in this PR (or directly at `main`) as we've extensively refined data structures and the expiration API.

PR #3072, opened 2024-12-12
Allow to prepare the pool

goetas:
Hi, this is my first pull request / contribution to a Java project.

I have a need to prepare/warmup the connection pool in advance, and it seems that it was not possible before.

Would a pull request as this be accepted? If yes can can do all the needed adjustments (formatting, tests, documentation...)

This would instruct the `LettucePoolingConnectionProvider` to warmup the pool before returning the connection.

Thank you!
----

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettucePoolingConnectionProvider.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettucePoolingConnectionProvider.java
@@ -55,6 +55,7 @@ import org.springframework.util.Assert;
  *
  * @author Mark Paluch
  * @author Christoph Strobl
+ * @author Asmir Mustafic
  * @since 2.0
  * @see #getConnection(Class)
  */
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/redis/connection/lettuce/LettucePoolingClientConfiguration.java
```diff
@@ -32,10 +32,12 @@
  * @author Mark Paluch
  * @author Christoph Strobl
  * @author Yanming Zhou
+ * @author Asmir Mustafic
  * @since 2.0
  */
 public interface LettucePoolingClientConfiguration extends LettuceClientConfiguration {
 
+	boolean preparePool();
```
goetas, line 40: is by your BC policy a compatibility break? how to add an additional configuration?

mp911de: Adding a method to an interface is formally a breaking change if the new method is not a `default` one. However, we also distinguish between interfaces implemented by our framework and those, that are intended to be implemented by applications. The config infrastructure is to hide implementation details and so that would have been a safe change to make.

I would not expose this flag at all, because the functionality to prepare the pool should be part of the pool initialization.

goetas:
> I would consider the missing call to GenericObjectPool.preparePool even a bug as we do not enforce midIdle

if this is a bug, then maybe i do not need to add an additional config? should i just call `preparePool` ?

mp911de:
> if this is a bug, then maybe i do not need to add an additional config? 

That's how I see it as well.

> should i just call `preparePool` ?

Yes, that should be sufficient.

goetas:
All changes have been done as suggested

src/main/java/org/springframework/data/redis/connection/lettuce/LettucePoolingClientConfiguration.java
goetas: I have handled it as a bug and removed it

goetas:
@mp911de  I do not know how the release process of this project works, do I need to do anything else regarding this issue?
Will my fix it be accepted?

mp911de:
I updated the PR target for a 3.4.x backport instead of 3.3.8. Jedis now also makes use of `preparePool`, and to reduce the risk for impact, we would keep the fix for version 3.4, 3.5 and 4.0.

mp911de:

PR #3131, opened 2025-04-16
Added a mention of the PatternTopic class to the pubsub.adoc document

dev-jonghoonpark:
related doc: https://docs.spring.io/spring-data/redis/reference/redis/pubsub.html

Previously, the document explained that listener topics can be set using a pattern but did not mention the `PatternTopic` class, so this information has been added.


mp911de:

Issue #3078, opened 2024-12-17
Why does the StreamListener consume only one record?


Danden1:
I discovered that there is an option called batch size.

```
StreamMessageListenerContainerOptions<String, ObjectRecord<String, Long>> containerOptions = StreamMessageListenerContainerOptions
				.builder().batchSize(3).pollTimeout(Duration.ofMillis(100)).targetType(Long.class).build();
```

So, I expected that using this option would fetch all the data at once.

However, `onMessage` method cannot take a list as a parameter.

```
class ExampleStreamListener implements StreamListener<String, MapRecord<String, String, String>> {

	@Override
        //here
	public void onMessage(MapRecord<String, String, String> message) {

		System.out.println("MessageId: " + message.getId());
		System.out.println("Stream: " + message.getStream());
		System.out.println("Body: " + message.getValue());
	}
}
```

It reads the data in batch size at once, but uses a for loop to deliver the data one by one.

```
class StreamPollTask<K, V extends Record<K, ?>> implements Task {
....
    private void deserializeAndEmitRecords(List<ByteRecord> records) {
    
        for (ByteRecord raw : records) {
    
	    try {
    
                pollState.updateReadOffset(raw.getId().getValue());
                V record = convertRecord(raw);
                listener.onMessage(record);
	    } catch (RuntimeException ex) {
    
                if (cancelSubscriptionOnError.test(ex)) {
    
	            cancel();
	            errorHandler.handleError(ex);
            
	            return;
                }
            
                errorHandler.handleError(ex);
            }
        }
    }

}
```

I'm curious why it doesn't provide the data as a list at once, but instead delivers it this way.

(KafkaListener allow list, https://docs.spring.io/spring-kafka/reference/kafka/receiving-messages/listener-annotation.html#batch-listeners)

mp911de:
This is because we do not provide a batch listener abstraction and in the current approach, we sequentially handle each record regardless of whether batch reading is used. Feel free to submit a pull request that introduces a batch listener concept. `StreamListener` would need to be retrofitted in terms of its generics.

Danden1:
Thank you for your response, I'll give it a try.

PR #3138, opened 2025-04-24
Make StreamListener read batch messages at once

Danden1:

Resolves [spring-projects/spring-data-redis#3078](https://github.com/spring-projects/spring-data-redis/issues/3078)

This pull request implements support for batch message consumption in StreamListener,
allowing it to receive and process multiple messages at once.

### DIFF
--- a/src/main/java/org/springframework/data/redis/stream/StreamListener.java
+++ b/src/main/java/org/springframework/data/redis/stream/StreamListener.java
@@ -17,6 +17,8 @@ package org.springframework.data.redis.stream;
 
 import org.springframework.data.redis.connection.stream.Record;
 
+import java.util.List;
+
 /**
  * Listener interface to receive delivery of {@link Record messages}.
  *
@@ -33,5 +35,5 @@ public interface StreamListener<K, V extends Record<K, ?>> {
 	 *
 	 * @param message never {@literal null}.
 	 */
-	void onMessage(V message);
+	void onMessage(List<V> message);
 }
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/redis/stream/StreamListener.java
```diff
@@ -33,5 +36,5 @@ public interface StreamListener<K, V extends Record<K, ?>> {
 	 *
 	 * @param message never {@literal null}.
 	 */
-	void onMessage(V message);
+	void onMessage(List<V> message);
```
mp911de, line 39: That won't work as this is a breaking change.

src/main/java/org/springframework/data/redis/stream/StreamPollTask.java
```diff
@@ -147,14 +148,14 @@ private List<ByteRecord> readRecords() {
...
+		List<V> messages = new ArrayList<>();
 
 		for (ByteRecord raw : records) {
 
 			try {
-
 				pollState.updateReadOffset(raw.getId().getValue());
 				V record = convertRecord(raw);
-				listener.onMessage(record);
+				messages.add(record);
```
mp911de, line 158: Batching introduces an error category of message loss if one of the messages causes an exception. In such a case, the read offset has been updated but the message was never delivered to a listener.

mp911de (changes requested):

src/main/java/org/springframework/data/redis/stream/StreamPollTask.java
Danden1: Thanks for the feedback. I will revise the PR based on your suggestions.
I'll try implementing a separate class called `StreamBatchListener` to handle batch processing.

mp911de:
Closing as per my comments in expectation of a future pull request.

Issue #2168, opened 2021-09-16
Rewrite conversion of top-level collections in `MappingRedisConverter`


mp911de:
`MappingRedisConverter.write(…)` attempts to obtain an entity from the given object to later on decide how to serialize it (collection, map, …). `PersistentEntity` retrieval breaks if the type is encapsulated (e.g. `Arrays.asList(…)`). The code still attempts to obtain a keyspace and the identifier accessor. 

We should generally avoid identifier, keyspace name, and TTL retrieval from a collection as collections do not have such a property.

Leehyoungwoo:
Hi @mp911de 
I’ve checked the MappingRedisConverter.write(...) method in #2168 and it looks like the early “if (source instanceof Collection)” branch hasn’t been added yet. Would you mind if I pick this up and submit a PR that moves the collection check to the top and skips the entity/keyspace lookup for top-level collections?

PR #3146, opened 2025-05-09
Handle top-level collections without entity lookup in MappingRedisConverter #2168

Leehyoungwoo:
Fix. #2168 

### Summary

This issue arises in ```MappingRedisConverter.write(…)```, which by default treats all inputs—including JDK internal types like ```Arrays.asList(...)```—as domain entities. It looks up a ```PersistentEntity``` for every input and then retrieves metadata (keyspace, identifier, TTL). On Java 9+ this leads to an ```InaccessibleObjectException``` when reflection tries to access the non-exported constructor of ```Arrays$ArrayList.```

To address this, I added an early check in write(...):

```java
if (source instanceof Collection) {
    writeCollection(
        sink.getKeyspace(),
        "",
        (List<?>) source,
        TypeInformation.of(Object.class),
        sink
    );
    return;
}
```
This skips all metadata lookups for top-level collections by directly invoking ```writeCollection(...)```.

I removed the ```@EnabledOnJre(JRE.JAVA_8)``` restriction on the ```writePlainList()``` test (now annotated // GH-2168) and verified that it passes on non–Java 8 environments.


christophstrobl:
@Leehyoungwoo thank you for your contribution. Merged to _main_ development line.

Issue #3092, opened 2025-01-15
Migrate to JSpecify annotations for nullability constraints


mp911de:

AnneMayor:
@mp911de can you explain to me what the leftover is? I would like to open PR to resolve this issue if you don't mind.

mp911de:
This ticket needs to be fully resolved. Migrating off `org.springframework.lang.@Nullable|@NonNullApi` to use JSpecifys `@Nullable` and `@NullMarked` (on the package), adding missing annotations to all Template API implementations and ensuring that NullAway does not report any nullability errors. This is quite an effort, even for a seasoned Spring developer so I recommend that you leave this one to us except if you are looking for a difficult-to-master adventure along with a hard time looking for workarounds.

AnneMayor:
Oh, I see. Thanks for your comment. For now, I think it would be better to resolve others. I would challenge a difficult-to-master adventure if I resolve lots of issues and understand the whole picture of this library.

PR #3159, opened 2025-05-27
Use to JSpecify for nullness checks

christophstrobl:
Closes: #3092


mp911de:

PR #3162, opened 2025-05-28
Add configurable deletion strategy for Redis repository operations #2294

### LINKED ISSUE #2294: Substitute/configure DEL for UNLINK on respository.save implementations
On update intensive applications, where CrudRepository.save/saveAll is called many times on already existing keys, the current implementation uses DEL command before SET, and this puts some pressure on the DB since is blocking command.
E.g. [https://github.com/spring-projects/spring-data-redis/blob/485bcad1108cdcb250157dbbc1bdad8c7fcac4af/src/main/asciidoc/reference/redis-repositories.adoc#replace-existing](replace)

Specially on larger structures, when you need to update a value many times, this behavior put a lot of pressure on the DB.
Each save on existing keys executes (setting TTL as well):
`"DEL" "key"`
`"HMSET" "key" "_class" "something" "id" ...`
`"EXPIRE" "key" "259200"`

Another place where DEL is used that could be substituted by UNLINK is when cleaning the secondary indexes on the ExpirationListener:
org.springframework.data.redis.core.RedisKeyValueAdapter.MappingExpirationListener#onMessage

I know from Redis 6 on we can configure to do DEL as UNLINK, however this would affect all applications. Would be useful to provide configuration for using DEL or UNLINK 

The versions where this is observed:
`<spring.boot.version>2.6.2</spring.boot.version>`
`<artifactId>redisson-spring-boot-starter</artifactId>`
`<redisson.version>3.16.7</redisson.version>`

kssumin:
## Summary

This PR implements configurable deletion strategy for Redis repository operations, allowing applications to choose between synchronous `DEL` and asynchronous `UNLINK` commands for better performance optimization.

## Background

Currently, Spring Data Redis repositories always use the blocking `DEL` command when updating existing keys, which can cause performance issues in update-intensive applications, especially when dealing with large data structures. Redis 4.0+ introduced the non-blocking `UNLINK` command as an alternative.

### Core Implementation
- **`DeletionStrategy` enum**: Defines `DEL` (default) and `UNLINK` options
- **`RedisKeyValueAdapter`**: 
  - Added `setDeletionStrategy()` and `getDeletionStrategy()` methods
  - Updated `put()`, `delete()`, and expiration handling to use configured strategy
  - Added `applyDeletionStrategy()` utility method
- **`@EnableRedisRepositories`**: Extended with `deletionStrategy` attribute
- **`RedisRepositoryConfigurationExtension`**: Updated to propagate configuration

### Test Coverage
- **`RedisKeyValueAdapterTests`**: Basic functionality and configuration tests
- **`RedisRepositoryConfigurationExtensionUnitTests`**: Annotation configuration tests

Closes #2294

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/RedisKeyValueAdapter.java
+++ b/src/main/java/org/springframework/data/redis/core/RedisKeyValueAdapter.java
@@ -22,6 +22,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
@@ -103,6 +104,7 @@ import org.springframework.util.ObjectUtils;
  * @author Mark Paluch
  * @author Andrey Muchnik
  * @author John Blum
+ * @author Kim Sumin
  * @since 1.7
  */
 public class RedisKeyValueAdapter extends AbstractKeyValueAdapter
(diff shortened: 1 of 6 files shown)


christophstrobl:
Thank you for getting in touch and taking the time to craft the PR.
There is a potential benefit of `UNLINK` when working with large data structures as you mentioned.
However I'm not fully convinced we need to have the behaviour configurable here. `UNLINK` calculates the cost of deleting the key and freeing the memory (in case of hash by looking at the number of elements). So it may do it immediately or later. Since Redis 6.0 `lazyfree-lazy-user-del` offers a server side flag to mrodify the default behaviour of `DEL` to act like `UNLINK`.

Issue #3103, opened 2025-02-12
Algin components with core container Lifecycle


christophstrobl:
Revisit components and check if they align with the core container lifecycle arrangement.

- `afterPropertiesSet` -> validate
- `start` -> create resources/connect
- `stop` -> disconnect
- ...

mayleaf:
Hi, I’m interested in this issue. I’d like to contribute—would it be okay if I work on resolving it?

Issue #3100, opened 2025-02-06
LettuceConnectionFactory.destroy(…) releases cluster connections after shutting down the connection pool


rsjoberg:
It seems like issue #2330 has been reintroduced while migrating RedisConnectionFactory to Lifecycle beans (in #2503).

During shutdown of a Spring Boot application using version 3.4.0 we get:

`org.springframework.data.redis.connection.PoolException: Returned connection io.lettuce.core.cluster.StatefulRedisClusterConnectionImpl@19ff8f0b was either previously returned or does not belong to this connection provider
    at org.springframework.data.redis.connection.lettuce.LettucePoolingConnectionProvider.release(LettucePoolingConnectionProvider.java:152)
    at org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory$ExceptionTranslatingConnectionProvider.release(LettuceConnectionFactory.java:1831)
    at org.springframework.data.redis.connection.lettuce.LettuceClusterConnection$LettuceClusterNodeResourceProvider.destroy(LettuceClusterConnection.java:593)
    at org.springframework.data.redis.connection.ClusterCommandExecutor.destroy(ClusterCommandExecutor.java:353)
    at org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory.destroy(LettuceConnectionFactory.java:1019)
    at org.springframework.beans.factory.support.DisposableBeanAdapter.destroy(DisposableBeanAdapter.java:211)
    at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroyBean(DefaultSingletonBeanRegistry.java:587)
    at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroySingleton(DefaultSingletonBeanRegistry.java:559)
    at org.springframework.beans.factory.support.DefaultListableBeanFactory.destroySingleton(DefaultListableBeanFactory.java:1202)
    at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroySingletons(DefaultSingletonBeanRegistry.java:520)
    at org.springframework.beans.factory.support.DefaultListableBeanFactory.destroySingletons(DefaultListableBeanFactory.java:1195)
    at org.springframework.context.support.AbstractApplicationContext.destroyBeans(AbstractApplicationContext.java:1195)
    at org.springframework.context.support.AbstractApplicationContext.doClose(AbstractApplicationContext.java:1156)
[...]

PR #3164, opened 2025-05-29
Fix `LettucePoolingConnectionProvider` and `ClusterCommandExecutor` disposal

mayleaf:
### Enhancements to `LettuceConnectionFactory`:

* Added proper disposal of `ClusterCommandExecutor` during the `stop` method to ensure resources are released cleanly. 
* Removed redundant cleanup logic for `ClusterCommandExecutor` from the `destroy` method, as it is now handled in the `stop` method. 
* Introduced a new private `dispose` method to encapsulate the cleanup logic for `ClusterCommandExecutor`, improving code readability and maintainability. 

### Enhancements to `LettucePoolingConnectionProvider`:

* Implemented `SmartLifecycle` to manage the lifecycle of the connection provider, including `start`, `stop`, and `isRunning` methods. 
* Added an `AtomicReference`-based `State` enum to track the lifecycle state of the connection provider, ensuring thread-safe transitions between states. 
These updates improve the robustness and maintainability of the connection handling logic in the Spring Data Redis project.
---
I have found out that #3100 was reproduced when redis run with cluster mode.

LettuceConnectionFactory uses a connectionProvider that refered at ClusterCommandExecuter. When redis run with Cluster mode, it uses LettucePoolingConnectionProvider, which only implements DisposableBean. 

When the LettuceConnectionFactory disposed, it disposes ConnectionProvider first, and disposes ClusterCommandExecuter next.

And when the ClusterCommandExecuter disposed, it disposes the connectionProvider - shared with LettuceConnectionFactory. As a result, LettucePoolingConnectionProvider's destroy method called twice.

Finally, the LettucePoolingConnectionProvider tried releasing connections in it's pool - that has 0 connections.

To solve this issue, I've implemented SmartLifeCycle at LettucePoolingConnectionProvider, and prevent duplicated stop method execution by AtomicReference based state management.
Solved #3100, See #3103 

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
@@ -116,6 +116,7 @@ import org.springframework.util.StringUtils;
  * @author Chris Bono
  * @author John Blum
  * @author Zhian Chen
+ * @author UHyeon Jeong
  */
 public class LettuceConnectionFactory implements RedisConnectionFactory, ReactiveRedisConnectionFactory,
 		InitializingBean, DisposableBean, SmartLifecycle {
@@ -979,6 +980,9 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 			dispose(reactiveConnectionProvider);
 			reactiveConnectionProvider = null;
 
+			dispose(clusterCommandExecutor);
+			clusterCommandExecutor = null;
+
 			if (client != null) {
 				try {
 					Duration quietPeriod = clientConfiguration.getShutdownQuietPeriod();
@@ -1012,20 +1016,7 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 
 	@Override
 	public void destroy() {
-
 		stop();
-		this.client = null;
-
-		ClusterCommandExecutor clusterCommandExecutor = this.clusterCommandExecutor;
-
-		if (clusterCommandExecutor != null) {
-			try {
-				clusterCommandExecutor.destroy();
-				this.clusterCommandExecutor = null;
-			} catch (Exception ex) {
-				log.warn("Cannot properly close cluster command executor", ex);
-			}
-		}
 
 		this.state.set(State.DESTROYED);
 	}
@@ -1039,6 +1030,16 @@ public class LettuceConnectionFactory implements RedisConnectionFactory, Reactiv
 				if (log.isWarnEnabled()) {
 					log.warn(connectionProvider + " did not shut down gracefully.", ex);
 				}
+			}
+		}
+	}
+
+	private void dispose(@Nullable ClusterCommandExecutor commandExecutor) {
+		if (commandExecutor != null) {
+			try {
+				commandExecutor.destroy();
+			} catch (Exception ex) {
+				log.warn("Cannot properly close cluster command executor", ex);
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionFactory.java
```diff
@@ -979,6 +980,9 @@ public void stop() {
 			dispose(reactiveConnectionProvider);
 			reactiveConnectionProvider = null;
 
+			dispose(clusterCommandExecutor);
```
mp911de, line 983: In these changes, `ClusterCommandExecutor` is being disposed after disposing the connection provider and so there's no behavioral change. Shouldn't `clusterCommandExecutor` be disposed (and thus its shared connection returned to the pool) before disposing `connectionProvider`?

mayleaf: You’re right that there’s no behavioral change in ClusterCommandExecutor itself. However, despite other resources at LettuceConnectionFactory are disposed at `stop`, only `ClusterCommandExecutor` disposed at `destroy` method. I was concerned that if we leave the dispose call in destroy, it could be executed twice in the future, as `LettucePoolingConnectionProvider` did. Since the stop method uses an AtomicReference guard, this change ensures safe and idempotent disposal.

mayleaf: And I believe the order of disposing connectionProvider doesn’t affect behavior in this case.

Below are the previous implementations of LettucePoolingConnectionProvider#destroy, #release As you can see, this method fully clears all pools and references on the first invocation. If it were called again, it could potentially throw exceptions or behave unexpectedly due to already-released resources.
    
```
    @Override
	public void destroy() throws Exception {

		List<CompletableFuture<?>> futures = new ArrayList<>();
		if (!poolRef.isEmpty() || !asyncPoolRef.isEmpty()) {
			log.warn("LettucePoolingConnectionProvider contains unreleased connections");
		}

		if (!inProgressAsyncPoolRef.isEmpty()) {

			log.warn("LettucePoolingConnectionProvider has active connection retrievals");
			inProgressAsyncPoolRef.forEach((k, v) -> futures.add(k.thenApply(StatefulConnection::closeAsync)));
		}

		if (!poolRef.isEmpty()) {

			poolRef.forEach((connection, pool) -> pool.returnObject(connection));
			poolRef.clear();
		}

		if (!asyncPoolRef.isEmpty()) {

			asyncPoolRef.forEach((connection, pool) -> futures.add(pool.release(connection)));
			asyncPoolRef.clear();
		}

		pools.forEach((type, pool) -> pool.close());

		CompletableFuture
				.allOf(futures.stream().map(it -> it.exceptionally(LettuceFutureUtils.ignoreErrors()))
						.toArray(CompletableFuture[]::new)) //
				.thenCompose(ignored -> {

					CompletableFuture[] poolClose = asyncPools.values().stream().map(AsyncPool::closeAsync)
							.map(it -> it.exceptionally(LettuceFutureUtils.ignoreErrors())).toArray(CompletableFuture[]::new);

					return CompletableFuture.allOf(poolClose);
				}) //
				.thenRun(() -> {
					asyncPoolRef.clear();
					inProgressAsyncPoolRef.clear();
				}) //
				.join();

		pools.clear();
	}    
```

```
    @Override
	public void release(StatefulConnection<?, ?> connection) {

		GenericObjectPool<StatefulConnection<?, ?>> pool = poolRef.remove(connection);

		if (pool == null) {

			AsyncPool<StatefulConnection<?, ?>> asyncPool = asyncPoolRef.remove(connection);

			if (asyncPool == null) {
				throw new PoolException("Returned connection " + connection
						+ " was either previously returned or does not belong to this connection provider");
			}

			discardIfNecessary(connection);
			asyncPool.release(connection).join();
			return;
		}

		discardIfNecessary(connection);
		pool.returnObject(connection);
	}

```

To prevent this, the disposal logic was moved to under the state safe guard, ensuring it’s executed only once and making the lifecycle handling more robust.

mayleaf:
Hi @mp911de, PTAL at the following comment when you get a chance. Thanks! 😊

mp911de:

PR #3168, opened 2025-06-11
Add support for Jackson 3-based `HashMapper` and `RedisSerializer`

christophstrobl:
work in progress...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,6 +149,20 @@
 			<optional>true</optional>
 		</dependency>
 
+		<!-- jackson 3 -->
+		<dependency>
+			<groupId>tools.jackson.core</groupId>
+			<artifactId>jackson-databind</artifactId>
+			<optional>true</optional>
+		</dependency>
+
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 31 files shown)


pom.xml
```diff
@@ -148,6 +148,21 @@
 			<optional>true</optional>
 		</dependency>
 
+		<!-- jackson 3 -->
+		<dependency>
+			<groupId>tools.jackson.core</groupId>
+			<artifactId>jackson-databind</artifactId>
+			<version>3.0.0-rc5</version>
```
mp911de: Dependency management available: https://github.com/spring-projects/spring-data-build/issues/2594

christophstrobl: hm... I see some test failures using that.

christophstrobl: ok, so `com.fasterxml.jackson.core:jackson-annotations` did, for whatever reason, not transition to `tools.jackson`. The outcome is as expected a mess, with both the `2.19` as well as the `3.0` BOM pointing to different versions.

christophstrobl: oh, to round things off the verion schema is different, so you need `3.0-rc5` for `jackson-annotations`. Not `3.0.0-rc5` as for the rest.

PR #3173, opened 2025-06-22
perf(StringRedisTemplate): avoid redundant logger creation in execute().

huangcanda:
## Problem Description  
Each invocation of `StringRedisTemplate.execute()` creates a new `DefaultStringRedisConnection` instance, resulting in **repeated initialization of non-static logger instances**. This causes redundant `LogFactory` lookups that ​**consume an additional 10–15% CPU overhead**.

## Solution  
Modify the `private final Log logger` field in `DefaultStringRedisConnection` to **`private static final Log logger`**.

## Impact  
**10–15% reduction in CPU overhead** for all Redis client command executions.


christophstrobl:
Thanks for the heads up. There are a couple of other places in the codebase as well that would benefit from this kind of change.  Would you like have a look and extend the PR to those, or should we take care of the rest.

huangcanda:
Sure, I’d be happy to extend the PR.

However, since my initial change was based on a specific performance issue I encountered, I haven’t done a full review of the codebase.

If you could point me to the places you think would benefit from similar optimizations, I’ll gladly take a look and update the PR accordingly.

christophstrobl:
great - thank you!
I think we should also change the following: `DefaultStreamReceiver`, `JedisClusterConnection`, `JedisConnection`, `JedisConnectionFactory`, `LettuceClusterConnection`, `LettuceConnection`, `LettuceConnectionFactory`, `RedisAccessor`, `RedisKeyValueAdapter`, `RedisMessageListenerContainer`, `MessageListenerAdapter`

huangcanda:
I've updated the PR based on your suggestions. Please review when convenient. Thanks!

PR #3188, opened 2025-07-24
Fixes #3179 – Restore _class metadata for collections in MappingRedisConverter

### LINKED ISSUE #3179: Regression in latest Redis Stream serialization (or deserialization)
This is a test which passes with Spring Data Redis `3.5.1`:
````java
	@Test
	void listAsRecordValueInStream() {
		String streamKey = "test-stream";
		List<String> input = Arrays.asList("Hello", "stream", "message");

		ReactiveRedisTemplate<String, ?> template =
				new ReactiveRedisTemplate<>(this.redisConnectionFactory, RedisSerializationContext.string());

		ReactiveStreamOperations<String, Object, Object> streamOperations = template.opsForStream();

		@SuppressWarnings("unchecked")
		Mono<List<?>> result =
				streamOperations.add(StreamRecords.objectBacked(input).withStreamKey(streamKey))
						.thenMany(streamOperations.read(List.class, StreamOffset.fromStart(streamKey)))
						.next()
						.map(Record::getValue);

		StepVerifier.create(result)
				.expectNext(input)
				.verifyComplete();
	}
````

In debug, on the `StreamObjectMapper.toObjectRecord()`, I see the value like:
```
source = {StreamRecords$MapBackedRecord@9126} "MapBackedRecord{recordId=1752698244313-0, kvMap={[0]=Hello, [0]._class=java.lang.String, [1]=stream, [1]._class=java.lang.String, [2]=message, [2]._class=java.lang.String, _class=java.util.Arrays$ArrayList}}"
 stream = "test-stream"
 recordId = {RecordId@9194} "1752698244313-0"
 kvMap = {LinkedHashMap@9195}  size = 7
  "[0]" -> "Hello"
  "[0]._class" -> "java.lang.String"
  "[1]" -> "stream"
  "[1]._class" -> "java.lang.String"
  "[2]" -> "message"
  "[2]._class" -> "java.lang.String"
  "_class" -> "java.util.Arrays$ArrayList"
```

With latest Spring Data Redis `4.0.0-SNAPSHOT`, the test fails like:
```
	Suppressed: java.lang.IllegalArgumentException: Value must not be null
		at org.springframework.util.Assert.notNull(Assert.java:182)
		at org.springframework.data.redis.connection.stream.Record.of(Record.java:99)
		at org.springframework.data.redis.connection.stream.MapRecord.toObjectRecord(MapRecord.java:140)
		at org.springframework.data.redis.core.StreamObjectMapper.toObjectRecord(StreamObjectMapper.java:137)
		at org.springframework.data.redis.core.ReactiveStreamOperations.map(ReactiveStreamOperations.java:696)
		at org.springframework.data.redis.core.ReactiveStreamOperations.lambda$read$0(ReactiveStreamOperations.java:536)
```
where in debug at the same spot I see the value as:
```
record = {StreamRecords$MapBackedRecord@9470} "MapBackedRecord{recordId=1752698947969-0, kvMap={[0]=Hello, [0]._class=java.lang.String, [1]=stream, [1]._class=java.lang.String, [2]=message, [2]._class=java.lang.String}}"
 stream = "test-stream"
 recordId = {RecordId@9523} "1752698947969-0"
 kvMap = {LinkedHashMap@9524}  size = 6
  "[0]" -> "Hello"
  "[0]._class" -> "java.lang.String"
  "[1]" -> "stream"
  "[1]._class" -> "java.lang.String"
  "[2]" -> "message"
  "[2]._class" -> "java.lang.String"
```

Pay attention that there is no top-level `_class` entry in the map.

Looks like `MappingRedisConverter.write(Object source, RedisData sink)` has suffered some changes and there is no `typeMapper.writeType(ClassUtils.getUserClass(source), sink.getBucket().getPath());` to put that `_class` entry into the target KV.
With the latest change we just have some optimization for collections:
```
		if (source instanceof Collection collection) {
			writeCollection(sink.getKeyspace(), "", collection, TypeInformation.of(Object.class), sink);
			return;
		}
```

Probably a side effect after: https://github.com/spring-projects/spring-data-redis/issues/2168.

The `MappingRedisConverter.read(Class<R> type, RedisData source)` in `3.5.1` picks up that `_class` from KV easy.
In `4.0.0-SNAPSHOT`, it gets null for that `typeMapper.readType()`. And I believe just because there is no that `_class` entry stored in the target KV.

Thanks

### PR DESCRIPTION
**Pull request by xyraclius**
## Overview

This PR resolves a regression in **Redis Stream serialization** introduced in `spring-data-redis` `4.0.0-SNAPSHOT`, where the `_class` metadata is no longer written when serializing collections. As a result, deserialization fails with:

> java.lang.IllegalArgumentException: Value must not be null

The issue is tracked in [GitHub issue #3179](https://github.com/spring-projects/spring-data-redis/issues/3179).

---

## The Problem

In Spring Data Redis `3.5.1`, when using `ReactiveRedisTemplate` to write and read a collection into a Redis Stream, the `MappingRedisConverter` correctly writes the `_class` metadata using the configured `typeMapper`. This metadata allows the deserializer to reconstruct the correct types during a read operation.

In `4.0.0-SNAPSHOT`, an optimization for `Collection` types bypasses the call to `typeMapper.writeType(...)`, resulting in the `_class` field being omitted. This omission breaks deserialization compatibility with existing data and behavior.

---

## The Solution

To restore the previous behavior while preserving the new optimization, this PR adds the following line:

```java
typeMapper.writeType(ClassUtils.getUserClass(source), sink.getBucket().getPath());
```

Before:
```java 
if (source instanceof Collection collection) {
    writeCollection(sink.getKeyspace(), "", collection, TypeInformation.of(Object.class), sink);
    return;
} 
```

After:
```java
if (source instanceof Collection collection) {
    typeMapper.writeType(ClassUtils.getUserClass(source), sink.getBucket().getPath());
    writeCollection(sink.getKeyspace(), "", collection, TypeInformation.of(Object.class), sink);
    return;
}
```

This ensures _class metadata is written for collection values in Redis Streams, restoring compatibility with existing deserialization logic.

## Test Coverage
This fix is already verified by the existing test case:

```java
@Test // GH-2168, GH-3179
void writePlainList() { ... }
```

This test verifies that _class metadata is written when serializing collections.
It fails without the fix and passes once the metadata is correctly applied.

---

### Related Issue
- Fixes: [#3179](https://github.com/spring-projects/spring-data-redis/issues/3179)

---
- [x] I have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-commons/blob/main/CONTRIBUTING.adoc)
- [x] I used the provided formatters and avoided submitting unrelated formatting changes
- [x] This fix is verified by an existing test case
- [x] I included a Signed-off-by trailer in the commit
- [x] I added myself as the author in the modified class header (if applicable)

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/convert/MappingRedisConverter.java
+++ b/src/main/java/org/springframework/data/redis/core/convert/MappingRedisConverter.java
@@ -403,6 +403,7 @@ public class MappingRedisConverter implements RedisConverter, InitializingBean {
 		}
 
 		if (source instanceof Collection collection) {
+			typeMapper.writeType(ClassUtils.getUserClass(source), sink.getBucket().getPath());
 			writeCollection(sink.getKeyspace(), "", collection, TypeInformation.of(Object.class), sink);
 			return;
 		}
--- a/src/test/java/org/springframework/data/redis/core/convert/MappingRedisConverterUnitTests.java
+++ b/src/test/java/org/springframework/data/redis/core/convert/MappingRedisConverterUnitTests.java
@@ -78,6 +78,7 @@ import org.springframework.data.redis.core.convert.KeyspaceConfiguration.Keyspac
 import org.springframework.data.redis.core.mapping.RedisMappingContext;
 import org.springframework.data.redis.serializer.Jackson2JsonRedisSerializer;
 import org.springframework.data.redis.test.util.RedisTestData;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
@@ -1994,16 +1995,22 @@ class MappingRedisConverterUnitTests {
 		assertThat(generic.entity.name).isEqualTo("hello");
 	}
 
-	@Test // GH-2168
+	@Test // GH-2168, GH-3179
 	void writePlainList() {
-
 		List<Object> source = Arrays.asList("Hello", "stream", "message", 100L);
 		RedisTestData target = write(source);
+		Object classValue = target.getBucket().get("_class");
 
-		assertThat(target).containsEntry("[0]", "Hello") //
-			.containsEntry("[1]", "stream") //
-			.containsEntry("[2]", "message") //
-			.containsEntry("[3]", "100");
+		assertThat(classValue)
+				.as("_class metadata should be written")
+				.isNotNull()
+				.isInstanceOf(byte[].class);
+		assertThat(new String((byte[]) classValue, StandardCharsets.UTF_8))
+				.isEqualTo(ClassUtils.getUserClass(source).getName());
+		assertThat(target).containsEntry("[0]", "Hello")
+				.containsEntry("[1]", "stream")
+				.containsEntry("[2]", "message")
+				.containsEntry("[3]", "100");
 	}
 
 	@Test // DATAREDIS-1175


christophstrobl:
Thank you @xyraclius for looking into this. 

The type mapper unfortunately will try to create a persistent entity during the type lookup for the user type. This leads to `InaccessibleObjectException` when trying to resolve the persistence constructor. So I think we also need to make sure we do not create an entity for collection like types in first place within the `RedisMappingContext`. 

We also might not need to preserve the user type in that detail. `_class : java.util.List` might be sufficient instead of having `_class : java.util.Arrays$ArrayList`. Since the `read` operation will create an `ArrayList` regardless of the actual type hint, leveraging `CollectionFactory.createCollection` sounds like the way to go there to get closer to the actual source type.

christophstrobl:
Thank you for your contribution @xyraclius - merged to _main_ development line.

PR #3202, opened 2025-08-17
Fix: new messages are not read when a consumer is specified

### LINKED ISSUE #3201: `DefaultStreamMessageListenerContainer` does't read new messages
Even though use `StreamOffset.create(streamKey, ReadOffset.lastConsumed())` to build the listener, it does't read new messages, and just read pending messages.

```java
        StreamMessageListenerContainer<String, ObjectRecord<String, String>> container =
                StreamMessageListenerContainer.create(redisMQTemplate.getRedisTemplate().getRequiredConnectionFactory(), containerOptions);
        Consumer consumer = Consumer.from(group, consumerName);
        StreamOffset<String> streamOffset = StreamOffset.create(streamKey, ReadOffset.lastConsumed());
        StreamMessageListenerContainer.StreamReadRequestBuilder<String> builder = StreamMessageListenerContainer.StreamReadRequest
                .builder(streamOffset).consumer(consumer)
                .autoAcknowledge(false)
                .cancelOnError(throwable -> false);
        container.register(builder.build(), listener);
```

> <https://redis.io/docs/latest/commands/xreadgroup/#differences-between-xread-and-xreadgroup>
> 
> The ID to specify in the STREAMS option when using XREADGROUP can be one of the following two:
> 
> - The special > ID, which means that the consumer want to receive only messages that were never delivered to any other consumer. It just means, give me new messages.
> - Any other ID, that is, 0 or any other valid ID or incomplete ID (just the millisecond time part), will have the effect of returning entries that are pending for the consumer sending the command with IDs greater than the one provided. So basically if the ID is not >, then the command will just let the client access its pending entries: messages delivered to it, but not yet acknowledged. Note that in this case, both BLOCK and NOACK are ignored.

Edsuns:
Fix the issue where new messages are not read when a consumer is specified.

Closes #3201

### DIFF
--- a/src/main/java/org/springframework/data/redis/stream/DefaultStreamMessageListenerContainer.java
+++ b/src/main/java/org/springframework/data/redis/stream/DefaultStreamMessageListenerContainer.java
@@ -39,6 +39,7 @@ import org.springframework.data.redis.core.RedisTemplate;
 import org.springframework.data.redis.core.StreamOperations;
 import org.springframework.data.redis.serializer.RedisSerializer;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 import org.springframework.util.ErrorHandler;
 import org.springframework.util.ObjectUtils;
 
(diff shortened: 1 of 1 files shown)


mp911de:
I think this is too little. `DefaultStreamReceiver` is also a stream consumer that needs similar adoption. The pull request should contain some tests and needs rebasing.

mp911de:
If you would like us to spend some time helping you to diagnose the problem, please spend some time describing it and, ideally, providing a [minimal yet complete sample](https://stackoverflow.com/help/mcve) that reproduces the problem.
You can share it with us by pushing it to a separate repository on GitHub or by zipping it up and attaching it to this issue.

PR #3221, opened 2025-09-21
Feat/support for redis 8 vector sets

### LINKED ISSUE #3153: Redis8 Vector Support
Need Redis8 Vector Support

AnneMayor:
This PR implements a partial feature to validate the code conventions and assess the impact scope. If this approach aligns with the project's direction, I will proceed with implementing the remaining functionality.

Please review and let me know if this is the right approach.

## AS-IS
no support for redis 8 vector set

## TO-BE
add feature [vAdd](https://redis.io/docs/latest/commands/vadd/) of redis 8 vector set for Jedis
partially resolved #3153 

## Next Step
- [ ] Add support for Lettuce
- [ ] Implement other features of redis 8 vector set
---


AnneMayor:
@mp911de I am going to go on the next step. Pleaes, let me know if I have to switch my PR direction.

src/main/java/org/springframework/data/redis/connection/RedisVectorSetCommands.java
```diff
@@ -0,0 +1,170 @@
...
+	/**
+	 * Add a vector to a vector set using FP32 binary format.
+	 * 
+	 * @param key the key
+	 * @param values the vector as FP32 binary blob
+	 * @param element the element name
+	 * @param options the options for the command
+	 * @return true if the element was added, false if it already existed
+	 */
+	Boolean vAdd(byte @NonNull [] key, byte @NonNull [] values, byte @NonNull [] element, VAddOptions options);
```
mp911de: `values` should be named `vector` to indicate better what this API method is about. `VAddOptions` can be explicitly `@Nullable`

src/main/java/org/springframework/data/redis/connection/RedisVectorSetCommands.java
```diff
@@ -0,0 +1,170 @@
...
+	/**
+	 * Add a vector to a vector set using double array.
+	 * 
+	 * @param key the key
+	 * @param values the vector as double array
+	 * @param element the element name
+	 * @param options the options for the command
+	 * @return true if the element was added, false if it already existed
+	 */
+	Boolean vAdd(byte @NonNull [] key, double @NonNull [] values, byte @NonNull [] element, VAddOptions options);
```
mp911de: Please rename `values` to `vector`. We do have a specific `Vector` type in Spring Data Commons and it would make sense to use that one as well as separate method. In fact, this `vAdd` method could be turned into a default method using `Vector.unsafe(double[])` and delegate to the method accepting `Vector`.

src/main/java/org/springframework/data/redis/connection/RedisVectorSetCommands.java
```diff
@@ -0,0 +1,170 @@
...
+		private VAddOptions(Builder builder) {
+			this.reduceDim = builder.reduceDim;
+			this.cas = builder.cas;
+			this.quantization = builder.quantization;
+			this.efBuildFactor = builder.efBuildFactor;
+			this.attributes = builder.attributes;
+			this.maxConnections = builder.maxConnections;
+		}
+
+		public static Builder builder() {
```
mp911de: We typically don't use builders (see `XAddOptions`).

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
```diff
@@ -595,6 +602,67 @@ static ZAddParams toZAddParams(ZAddArgs source) {
 		return target;
 	}
 
+	private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
mp911de: Let's not introduce a mandatory dependency on Jackson.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceVectorSetCommands.java
```diff
@@ -0,0 +1,59 @@
...
+		this.connection = connection;
+	}
+
+	@Override
+	public Boolean vAdd(byte[] key, byte[] values, byte[] element, VAddOptions options) {
+		Assert.notNull(key, "Key must not be null");
+		Assert.notNull(values, "Values must not be null");
+		Assert.notNull(element, "Element must not be null");
+
+		// TODO: Implement when Lettuce adds native support for V.ADD
```
mp911de, line 46: Lettuce 6.7 has support for Vector sets.

src/main/java/org/springframework/data/redis/connection/RedisVectorSetCommands.java
```diff
@@ -0,0 +1,170 @@
...
+	 * - Collections (List, Map) are supported for nested structures
+	 * - Custom objects require proper JSON serialization support
+	 * - Date/Time objects should be converted to String or timestamp before use
+	 */
+	class VAddOptions {
+		private final @Nullable Integer reduceDim;
+		private final boolean cas;
+		private final QuantizationType quantization;
+		private final @Nullable Integer efBuildFactor;
+		private final @Nullable Map<String, Object> attributes;
```
mp911de: Let's keep `attributes` opaque and make it a `String` without us being involved in JSON serialization for the time being.

mp911de (changes requested):
I went over the PR initially and left a few comments. The general direction looks neat, we should also have a reactive implementation and we have to make up our mind how we want to surface Vector sets from the Template API.

AnneMayor:
> I went over the PR initially and left a few comments. The general direction looks neat, we should also have a reactive implementation and we have to make up our mind how we want to surface Vector sets from the Template API.

Thank you for your comments. I am gonna address your comments.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceVectorSetCommands.java
tishun: FWIW the API surface in 6.7 is still in experimental state.
There are some pending changes that need to be addressed before we can promote it.
See https://github.com/redis/lettuce/issues/3459

AnneMayor: Thanks for the feedback! I’ll double-check and update the PR if needed.

AnneMayor:
Sorry for the late response. I plan to update and re-request the PR review by this weekend. Thanks for your patience.

src/main/java/org/springframework/data/redis/connection/RedisVectorSetCommands.java
AnneMayor: Understood. Thanks.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceVectorSetCommands.java
AnneMayor: I confirmed that lettuce has still been in experimental progress in terms of Vector API. I think we should wait until it is addressed completely. Any idea?

AnneMayor:
I think it is better to open new PR for support other vector APIs to make PR review more easier and comfortable. What do you guys think?

AnneMayor:
Hi, just following up on this.

Should I:
A) Keep this PR open and wait for Lettuce to stabilize the Vector API
B) Close this for now and reopen when Lettuce is ready
C) Proceed with current experimental API

Happy to go with any direction. Thanks!

mp911de:
Thanks for reaching out. I'm honestly torn. API's are still marked experimental and from my past experience I expect no significant changes to the API (except for the JSON part maybe). We also have support from the Redis folks now so I suggest to proceed with this PR forward in the sense of adding Lettuce support (Lettuce 6.7 has Vector API support) and adding a proper `VectorOperations` interface along with an implementation to follow the Template API design.

AnneMayor:
@mp911de 
Thanks for the guidance! That makes sense — I'll proceed with adding Lettuce 6.7 support and design a VectorOperations interface following the Template API pattern.
I'll update the PR accordingly and ping you once it's ready for another look.

Issue #3208, opened 2025-08-25
Allow overriding phase in RedisMessageListenerContainer


eye-dee:
**Title:** Allow overriding `phase` in `RedisMessageListenerContainer`

**Description:**
Currently, `RedisMessageListenerContainer` uses the default `SmartLifecycle` phase of `Integer.MAX_VALUE`.
This makes it difficult to control startup/shutdown ordering relative to other `SmartLifecycle` beans in the application.

In some scenarios, developers may want to ensure that certain beans start **after** the Redis listener container is fully initialized, or conversely, that the container waits until other infrastructure is ready before starting. Since the phase is not configurable, fine-grained lifecycle control is not possible.

**Proposal:**
Introduce the ability to override the `phase` of `RedisMessageListenerContainer`, for example by:

* Exposing a `setPhase(int phase)` method

**Use case:**

* Ensuring custom initialization logic (that depends on the container being ready) runs after the container is started.
* Aligning the container’s lifecycle with other messaging or connection beans that also implement `SmartLifecycle`.

bandalgomsu:
Hi @mp911de 
I’d like to contribute this issue
Is it okay if I take this issue and prepare an implementation?

mp911de:
Sure feel free to submit a pull request.

bandalgomsu:
PR: #3224 !

PR #3224, opened 2025-09-24
Allow configuring phase in `RedisMessageListenerContainer`

bandalgomsu:
This PR introduces a configurable `phase` property in `RedisMessageListenerContainer`.  
Previously, the container always used `Integer.MAX_VALUE` as its SmartLifecycle phase, making it difficult to align startup/shutdown ordering with other SmartLifecycle beans.
#3208 

### DIFF
--- a/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
+++ b/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
@@ -633,6 +633,7 @@ public class RedisMessageListenerContainer implements InitializingBean, Disposab
      * The default is {@code Integer.MAX_VALUE}.
      *
      * @see SmartLifecycle#getPhase()
+     * @since 4.0
      */
     public void setPhase(int phase) {
         this.phase = phase;
@@ -649,6 +650,7 @@ public class RedisMessageListenerContainer implements InitializingBean, Disposab
      * The default is {@code true}.
      *
      * @see SmartLifecycle#isAutoStartup()
+     * @since 4.0
      */
     public void setAutoStartup(boolean autoStartup) {
         this.autoStartup = autoStartup;
--- a/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
+++ b/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
@@ -103,6 +103,7 @@ import org.springframework.util.backoff.FixedBackOff;
  * @author Mark Paluch
  * @author John Blum
  * @author Seongjun Lee
+ * @author Su Ko
  * @see MessageListener
  * @see SubscriptionListener
  */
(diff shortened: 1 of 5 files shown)


bandalgomsu:
Done ! 🙂

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
```diff
@@ -618,6 +622,38 @@ public void removeMessageListener(MessageListener listener) {
...
+    @Override
+    public int getPhase() {
+        return this.phase;
+    }
+
+    /**
+     * Specify the lifecycle phase for this container.
+     * Lower values start earlier and stop later.
+     * The default is {@code Integer.MAX_VALUE}.
+     *
```
onobc, line 634: Do you mind adding a `@since 4.0.0` tag here and in the `setAutoStartup` as well?

onobc (changes requested):
👋🏻 @bandalgomsu ,

Thank you for this excellent contribution! It is looking good.

Do you mind adding the same treatment to the `org.springframework.data.redis.stream.DefaultStreamMessageListenerContainer` as it also adheres to the `SmartLifecycle` contract and will likely benefit from this improvement?

Thanks

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
bandalgomsu: What do you think about allowing 'phase' and 'autoStartup' to be configured either on 'StreamMessageListenerContainer' or through 'StreamMessageListenerContainerOptions' ? ?

bandalgomsu:
Hi @onobc 

What do you think about allowing 'phase' and 'autoStartup' to be configured either on 'StreamMessageListenerContainer' or through 'StreamMessageListenerContainerOptions' ? ?
Thanks 🙂

src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
mp911de: Nit: `4.0` instead of `4.0.0`

mp911de:
`StreamMessageListenerContainerOptions` sounds good.

onobc:
> `StreamMessageListenerContainerOptions` sounds good.

Agreed. Good suggestion @bandalgomsu .

bandalgomsu:
The implementation has been completed. Could you review it ?? 🙂

src/main/java/org/springframework/data/redis/stream/DefaultStreamMessageListenerContainer.java
```diff
@@ -67,6 +68,9 @@ class DefaultStreamMessageListenerContainer<K, V extends Record<K, ?>> implement
 
 	private boolean running = false;
 
+    private int phase = Integer.MAX_VALUE;
+    private boolean autoStartup = false;
```
onobc, line 72: Shouldn't the default be `true` like in the other containers?

onobc: Nm - I see it was previously false. Good catch.

onobc:
Thank you @bandalgomsu for the excellent enhancement - much appreciated.  I added a minor polish commit [here](https://github.com/spring-projects/spring-data-redis/commit/b8e52f82ffe957afd8308f913ed07aa7a1110a59) and am closing this PR in favor of https://github.com/spring-projects/spring-data-redis/commit/c88d6b11e03a240ed0089dd3ed3bb75c3edaea93

Issue #3223, opened 2025-09-23
`ValueOperations.get(…)` methods should be annotated with `@Nullable`


mp911de:
Keys could be (correctly) absent and we should at least indicate proper nullability on the level of `GET` commands.

PR #3244, opened 2025-10-24
Annotate ValueOperations.get(…) methods with @Nullable.

onobc:
This PR annotates the return values of several GET related `ValueOperations` methods with `@Nullable` to reflect the fact that null values are returned when non-existent keys are passed in.

Resolves: #3223

### DIFF
--- a/src/test/java/org/springframework/data/redis/core/DefaultValueOperationsIntegrationTests.java
+++ b/src/test/java/org/springframework/data/redis/core/DefaultValueOperationsIntegrationTests.java
@@ -46,6 +46,7 @@ import org.springframework.data.redis.test.condition.EnabledOnCommand;
  * @author Jiahe Cai
  * @author Mark Paluch
  * @author Hendrik Duerkop
+ * @author Chris Bono
  */
 @ParameterizedClass
 @MethodSource("testParams")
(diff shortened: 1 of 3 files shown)


src/test/java/org/springframework/data/redis/core/DefaultValueOperationsIntegrationTests.java
```diff
@@ -265,6 +293,9 @@ void testGetAndSet() {
 		valueOps.set(key, value1);
 
 		assertThat(valueOps.getAndSet(key, value2)).isEqualTo(value1);
+
+		K noSuchKey = keyFactory.instance();
+		assertThat(valueOps.getAndSet(noSuchKey, value2)).isNull();
```
onobc, line 298: > [!NOTE]
> If I pass in a non-existent `String` key to `ValueOperations.get(Object)` when using the `RawObjectFactory` variant of the template, the test will fail w/ NPE.

For example, if line 298 above is...
```java
assertThat(valueOps.getAndSet("no-such-key", value2)).isNull();
```
we get an NPE. Ultimately, `AbstractOperations#rawKey` is called and in this case the key is a string but the keySerializer() is null so we get NPE at https://github.com/spring-projects/spring-data-redis/blob/4da49608046f5d03a801059be9a47f3e400afbd3/src/main/java/org/springframework/data/redis/core/AbstractOperations.java#L125-L134 line 133.

Is passing in a String non-existent key when using a RawObjectFactory supported?

mp911de: `RawObjectFactory` is for usage with `byte[]` and therefore, not compatible with strings. It makes sense to add guards to catch configurations without a serializer early on.

mp911de:

PR #3254, opened 2025-11-02
Support `ZINTERCARD`

### LINKED ISSUE #3253: Support `ZINTERCARD`
Similar to `SINTERCARD` which implemented in https://github.com/spring-projects/spring-data-redis/blob/b09f75261c2042f2f0ab5438d0b56820d9b7e1ed/src/main/java/org/springframework/data/redis/core/DefaultSetOperations.java#L149-L154, It would be nice if we can provide same operation for zset.

heka1024:
Resolves: #3253

Similar to `SINTERCARD`, It would be nice if we can provide same operation for zset.

- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/DefaultStringRedisConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/DefaultStringRedisConnection.java
@@ -1085,6 +1085,16 @@ public class DefaultStringRedisConnection implements StringRedisConnection, Deco
 	}
 
 	@Override
+	public Long zInterCard(byte[]... sets) {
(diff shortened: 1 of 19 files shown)


mp911de:
Care to add `ZINTERCARD` for the reactive API as well? Also, there should be also tests on the connection level to verify proper functionality of the `ZINTERCARD` on the Jedis/JedisCluster/Lettuce Connection-level.

heka1024:
@mp911de Thanks for the comment. I’ve addressed it. Could you review again?

Issue #3232, opened 2025-10-08
Feature Request: Support XDELEX, XACKEX and extend XADD and XTRIM options from Redis 8.2


viktoriya-kutsarova:
Hello!

Redis 8.2 supports 2 new advanced stream entry management commands:

1. XDELEX: https://redis.io/docs/latest/commands/xdelex/
2. XACKDEL: https://redis.io/docs/latest/commands/xackdel/

These new command are already supported in Lettuce (Since [Release 6.8.0](https://github.com/redis/lettuce/releases/tag/6.8.0.RELEASE)) and Jedis (Since [Release 6.1.0](https://github.com/redis/jedis/releases/tag/v6.1.0))

Feature request: Support these commands in Spring Data Redis's [stream operations](https://docs.spring.io/spring-data/redis/reference/api/java/org/springframework/data/redis/core/ReactiveStreamOperations.html), similar to the existing: acknowledge and delete methods.

In addition, expand (XADD)[https://redis.io/docs/latest/commands/xadd/] and (XTRIM)[https://redis.io/docs/latest/commands/xtrim/] commands to support new supported options since Redis 8.2:

- KEEPREF
- DELREF
- ACKED

Kiminni:
Hi! @viktoriya-kutsarova,
If this issue doesn’t have an assignee yet, would it be alright if I take it?

viktoriya-kutsarova:
Hi @Kiminni, thanks for your interest! I just submitted my PR today.

PR #3247, opened 2025-10-24
Add support for XDELEX and XACKDEL, and expand options for XADD and XTRIM

viktoriya-kutsarova:
This change introduces the [Support XDELEX, XACKEX and extend XADD and XTRIM options](https://github.com/spring-projects/spring-data-redis/issues/3232) feature as part of the features provided by the spring-data-redis project. Closes spring-projects/spring-data-redis#3232.

As part of the change the following commands would be made available:

- [x] [XDELEX](https://redis.io/docs/latest/commands/xdelex/) - deletes one or multiple entries from the stream at the specified key. XDELEX is an extension of the Redis Streams XDEL command that provides more control over how message entries are deleted concerning consumer groups.
 - [x] [XACKDEL](https://redis.io/docs/latest/commands/xackdel/) - acknowledges and conditionally deletes one or multiple entries (messages) for a stream consumer group at the specified key. XACKDEL combines the functionality of XACK and XDEL in Redis Streams. It acknowledges the specified entry IDs in the given consumer group and simultaneously attempts to delete the corresponding entries from the stream.
 
As part of the change the following command options would be made available for XADD and XTRIM:

 - [x] KEEPREF | DELREF | ACKED - Specifies how to handle consumer group references when trimming. Available since Redis 8.2. If no option is specified, KEEPREF is used by default. Unlike the XDELEX and XACKDEL commands where one of these options is required, here they are optional to maintain backward compatibility:
- [x] =: Exact trimming (default) - trims to the exact threshold
- [x] LIMIT count - Limits the number of entries to examine during trimming. Available since Redis 6.2.0. When not specified, Redis uses a default value of 100 * the number of entries in a macro node. Specifying 0 disables the limiting mechanism entirely.

The new commands are available as part of the following interfaces:

- RedisStreamCommands / ReactiveStreamCommands
- StreamOperations / ReactiveStreamOperations / BoundStreamOperations
- StringRedisConnection
- New enum StreamDeletionPolicy and StreamEntryDeletionResult inside RedisStreamCommands

This feature is available starting from [Redis OSS version 8.2.x](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.2-release-notes/) and later

---
- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
[...]

PR #3272, opened 2025-12-02
[upstream] Add support for XDELEX and XACKDEL, and expand options for XADD and XTRIM

onobc:
This change introduces the `Support XDELEX, XACKEX and extend XADD and XTRIM options` feature...

_See original pull request for more details._

Closes: #3232 
Original Pull Request: #3247

### DIFF
--- a/src/test/java/org/springframework/data/redis/core/DefaultStreamOperationsIntegrationTests.java
+++ b/src/test/java/org/springframework/data/redis/core/DefaultStreamOperationsIntegrationTests.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.List;
 
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedClass;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/data/redis/connection/jedis/StreamConverters.java
```diff
@@ -232,7 +232,7 @@ public static XAddParams toXAddParams(RecordId recordId, XAddOptions options) {
 
 		if (options.hasTrimOptions()) {
 			TrimOptions trim = options.getTrimOptions();
-			var strategy = trim.getTrimStrategy();
+			TrimStrategy strategy = trim.getTrimStrategy();
```
onobc, line 224: I'm an advocate for `var` and I may bad-influenced you by using `var` in my previous commit or in some pseudo code when we were discussing things. If so, my bad. In general you will not find it in the main source of many Spring projects (sometimes in the tests).

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
```diff
@@ -105,13 +106,251 @@ default RecordId xAdd(@NonNull MapRecord<byte[], byte[], byte[]> record) {
...
+		APPROXIMATE
+	}
+
+	@NullMarked
+	class TrimOptions {
+
+		private final TrimStrategy trimStrategy;
+		private final TrimOperator trimOperator;
+		private final @Nullable Long limit;
+		private final @Nullable StreamDeletionPolicy pendingReferences;
```
onobc: I do have a small concern about possible confusion caused from tbe`deletionPolicy/getPendingReferences` rename. I believe now that the benefit of a better high-level api name `pendingRefereneces` does not outweigh the inconsistency between the two names throughout. I am leaning towards consistency here and just leave it named `deletionPolicy` for now. I would like to get yours MP input on this topic.

onobc:
cc: @viktoriya-kutsarova @mp911de

src/test/java/org/springframework/data/redis/connection/jedis/JedisConvertersUnitTests.java
```diff
@@ -555,4 +555,6 @@ private void assertThatParamsHasExpiration(HSetExParams params, Protocol.Keyword
 			assertThat(params).extracting("expiration", "expirationValue").containsExactly(expirationType, expirationValue);
 		}
 	}
+
```
mp911de, line 558: Nit: Formatting

src/test/java/org/springframework/data/redis/connection/jedis/StreamConvertersUnitTest.java
```diff
@@ -41,4 +51,201 @@ void shouldConvertIdle() {
...
+					redis.clients.jedis.args.StreamDeletionPolicy.KEEP_REFERENCES);
+		}
+
+		@Test
+		void convertXAddOptionsWithRecordId() {
+
+			RecordId recordId = RecordId.of("1234567890-0");
+			XAddOptions options = XAddOptions.none();
+
+			XAddParams params = StreamConverters.toXAddParams(recordId, options);
```
mp911de, line 89: A testing approach of testing that a component is not doing something quickly leads to a lot of complexity as the subsequent lines unfold into several assertions. Instead, I suggest testing for the expectation that the component is doing something we expect, namely, not adding any command arguments.

`XAddParams.addParams(CommandArguments)` would be a neat way to utilize public API for testing which command arguments are added.

src/test/java/org/springframework/data/redis/connection/jedis/StreamConvertersUnitTest.java
```diff
@@ -41,4 +51,201 @@ void shouldConvertIdle() {
...
+
+			XTrimParams params = StreamConverters.toXTrimParams(options);
+
+			assertThat(params).hasFieldOrPropertyWithValue("trimMode",
+					redis.clients.jedis.args.StreamDeletionPolicy.KEEP_REFERENCES);
+		}
+	}
+
+	@Nested // GH-3232
+	class ToStreamDeletionPolicyShould {
```
mp911de: Instead of having each invariant represented by a test method, can we turn this into parametrized tests if you want to cover each code path in tests? The current arrangement is pretty verbose requiring a lot of mental energy to parse those tests. Using a parametrized test condenses the setup into an assertion and a set of input/expected arguments.

src/test/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionUnitTests.java
```diff
@@ -259,6 +263,134 @@ void xaddShouldHonorNoMkStream() {
 			assertThat(args.getValue()).extracting("nomkstream").isEqualTo(true);
 		}
 
+		@Test // GH-3232
+		void xaddShouldHonorMinId() {
```
mp911de, line 267: Testing individual arguments in the `xadd` command series deviates from the underlying code.

Clearly, previously, the conversion has happened in the `xadd` method itself. With this pull request, all command conversion has been refactored to a converter, so it would be sufficient to keep a single `xadd` test with options and remove all other test variants to align with the code under test.

src/test/java/org/springframework/data/redis/core/DefaultStreamOperationsIntegrationTests.java
```diff
@@ -600,4 +780,140 @@ void claimShouldReadMessageDetails() {
...
+	void acknowledgeAndDeleteUsingRecordShouldWork() {
+
+		K key = keyFactory.instance();
+		HK hashKey = hashKeyFactory.instance();
+		HV value = hashValueFactory.instance();
+
+		RecordId messageId = streamOps.add(key, Collections.singletonMap(hashKey, value));
+
+		streamOps.createGroup(key, ReadOffset.from("0-0"), "my-group");
+
```
mp911de, line 915: Nit: Formatting

src/main/java/org/springframework/data/redis/connection/jedis/StreamConverters.java
```diff
@@ -37,7 +40,17 @@
 
 import org.jspecify.annotations.Nullable;
 import org.springframework.data.domain.Range;
-import org.springframework.data.redis.connection.RedisStreamCommands;
+import org.springframework.data.redis.connection.RedisStreamCommands.XAddOptions;
```
mp911de: Something seems off with formatting, typically, we use star imports for a import count over 10.

src/main/java/org/springframework/data/redis/connection/lettuce/StreamConverters.java
```diff
@@ -169,4 +199,123 @@ public XClaimArgs convert(XClaimOptions source) {
 
 		}
 	}
+
+	/**
+	 * {@link Converter} to convert {@link XAddOptions} to Lettuce's {@link XAddArgs}.
+	 *
+	 * @since 4.1
+	 */
+	enum XAddOptionsToXAddArgsConverter implements Converter<XAddOptions, XAddArgs> {
```
mp911de: Are those converters actually used as `Converter`? In the more recent times we started inlining all converter code into methods as storing conversion code in a `Converter` has been merely an inversion adding another step before one can get to the actual code that is doing the conversion. Less code makes things easier to understand.

src/main/java/org/springframework/data/redis/connection/ReactiveStreamCommands.java
```diff
@@ -1661,6 +2044,14 @@ default Mono<Long> xTrim(ByteBuffer key, long count, boolean approximateTrimming
 				.map(NumericResponse::getOutput);
 	}
 
+	default Mono<Long> xTrim(ByteBuffer key, XTrimOptions options) {
```
mp911de, line 2056: This public API method is missing Javadoc.

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
```diff
@@ -105,13 +106,251 @@ default RecordId xAdd(@NonNull MapRecord<byte[], byte[], byte[]> record) {
...
-	 * @param options additional options (eg. {@literal MAXLEN}). Must not be {@literal null}, use
+	 * @param options additional options (e.g. {@literal MAXLEN}). Must not be {@literal null}, use
 	 *          {@link XAddOptions#none()} instead.
 	 * @return the {@link RecordId id} after save. {@literal null} when used in pipeline / transaction.
 	 * @since 2.3
 	 */
 	RecordId xAdd(MapRecord<byte[], byte[], byte[]> record, @NonNull XAddOptions options);
 
+
+	sealed interface TrimStrategy permits MaxLenTrimStrategy, MinIdTrimStrategy {
```
mp911de, line 118: The interface (and all inner classes) are public and we should provide Javadoc for public API elements.

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
mp911de: I suspect that I came up with the name of `pendingReferences` and `StreamDeletionPolicy` (not really sure about whether it was me or someone else). In any case, both terms describe concepts. Redis command documentation says:

> Specifies how to handle consumer group references when trimming. If there are no consumer groups, these arguments have no effect. Available since Redis 8.2.
> When trimming, removes entries … preserves existing references to these entries in all consumer groups' PEL (Pending Entries List).

Redis sometimes isn't great at making some concepts explicit, instead the underlying concept leaks through documentation. I think `StreamDeletionPolicy` is the shortest possible name to express a deletion policy for stream entry references from the pending element list. Agreed that the combination is not super-self explanatory, however it is something that when I encounter, I would want to read up more what this is. `PendingElementListDeletionPolicy` could be an alternative, yet its wordiness is much more annoying than the state we are currently at.

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
```diff
@@ -183,16 +424,26 @@ public static XAddOptions maxlen(long maxlen) {
...
 		/**
 		 * Apply efficient trimming for capped streams using the {@code ~} flag.
 		 *
 		 * @return new instance of {@link XAddOptions}.
+		 * @deprecated since 4.0: callers must specify a concrete trim strategy (MAXLEN or MINID)
+		 * via {@link TrimOptions}; do not use this method to only toggle approximate/exact.
+		 * Prefer {@code XAddOptions.trim(TrimOptions.maxLen(n).approximate())} or
+		 * {@code XAddOptions.trim(TrimOptions.minId(id).exact())}.
 		 */
+		@Deprecated(since = "4.0", forRemoval = false)
```
mp911de, line 460: I suggest avoiding `forRemoval = true`. It's a pretty harsh signal. Just the deprecation is a lighter form of signalling improper API usage, and that there is a better variant.

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
```diff
@@ -209,37 +460,52 @@ public boolean isNoMkStream() {
...
 		 */
 		public boolean hasMinId() {
-			return minId != null;
+			return trimOptions != null && trimOptions.getTrimStrategy() instanceof MinIdTrimStrategy;
+		}
+
+		public XAddOptions nomkstream(boolean nomkstream) {
+			return new XAddOptions(nomkstream, trimOptions);
+		}
+
```
mp911de, line 522: Nit: Formatting

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
```diff
@@ -250,28 +516,111 @@ public boolean equals(@Nullable Object o) {
...
+	 * @since 4.1
+	 */
+	enum StreamEntryDeletionResult {
+
+		UNKNOWN(-2L),
+		/**
+		 * The entry ID does not exist in the stream.
+		 */
+		NOT_FOUND(-1L),
+		/**
```
mp911de, line 611: Nit: Formatting (missing blank line)

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
```diff
@@ -250,28 +516,111 @@ public boolean equals(@Nullable Object o) {
...
 	}
 
+	/**
+	 * Deletion policy for stream entries.
+	 *
+	 * @author Viktoriya Kutsarova
+	 * @since 4.1
+	 */
+	enum StreamDeletionPolicy {
+		/**
```
mp911de, line 562: Nit: Formatting (missing blank lines)

src/main/java/org/springframework/data/redis/connection/jedis/StreamConverters.java
viktoriya-kutsarova: Sure, let's keep using the proper type.

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
viktoriya-kutsarova: It looks like I introduced some inconsistency with the rename, so I want to make sure I understand: are we discussing the naming of the class (`StreamDeletionPolicy`), the field (currently `pendingReferences`), or both?

Personally, I would stick with `streamDeletionPolicy` for both. Based on the Redis documentation, all three options (KEEPREF | DELREF | ACKED) represent different policies for how entries are removed from the stream. KEEPREF and DELREF delete entries strictly according to the trimming strategy (differing only in how PEL references are handled), while ACKED deletes only entries that were read and acknowledged by all consumer groups. In that sense, `streamDeletionPolicy` feels like the most concise name that still reflects the underlying concept.

As you noted, consistency across the API is important, so aligning on a single name and using it consistently seems like the best approach.

src/main/java/org/springframework/data/redis/connection/lettuce/StreamConverters.java
onobc: Inlining..

src/test/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionUnitTests.java
onobc: > so it would be sufficient to keep a single xadd test with options and remove all other test variants to align with the code under test.

There is no `StreamConverterUnitTests` for Lettuce like there is for Jedis. So we would need to create a Lettuce `StreamConverterUnitTests` to cover the various test cases. 

It is also worth pointing out that many of the other APIs under test also delegate to converters yet the tests are implemented in here (same case as xtrim / xadd). Those could also be "moved" into the `StreamConverterUnitTests` and hence more tests simplified. 

I do think this would be a good thing to do but I am hesitant to do it just for xadd / xtrim (focus of GH-3232) because we would end up w/ 3/4 of the tests being done one way and the other 1/4 another way. 
On the other hand, the concern of waiting to "do it all in a other PR" is that PR may never come to fruition.

Would you prefer to do this all at once or incrementally, starting w/ xtrim and xadd?

src/test/java/org/springframework/data/redis/core/DefaultStreamOperationsIntegrationTests.java
onobc: I am not seeing the violation?

src/test/java/org/springframework/data/redis/connection/jedis/StreamConvertersUnitTest.java
onobc: I reworked the test to use `@ParameterizedTest` throughout. It definitely makes it more concise. In order to keep the context of each test I added a name to it. The tests are also named and arranged intentionally to produce a readable result in the test report:

Screenshot 2025-12-06 at 8 21 25 PM

src/test/java/org/springframework/data/redis/core/DefaultStreamOperationsIntegrationTests.java
mp911de: I think it's an oversight on my end, apologies for the noise.

src/test/java/org/springframework/data/redis/connection/lettuce/LettuceConnectionUnitTests.java
mp911de: > I do think this would be a good thing to do but I am hesitant to do it just for xadd / xtrim (focus of https://github.com/spring-projects/spring-data-redis/issues/3232) because we would end up w/ 3/4 of the tests being done one way and the other 1/4 another way.
> On the other hand, the concern of waiting to "do it all in a other PR" is that PR may never come to fruition.

I think we should wrap up this effort at some point and so stretching polishing beyond a sane point is rather preventing a wrap-up. I suggest creating a follow-up ticket to migrate those tests. What do you think?

onobc: Done. https://github.com/spring-projects/spring-data-redis/issues/3276

src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
onobc: I went w/ the consistent name of `deletionPolicy`.

onobc:
Closing in favor of #3247

PR #3275, opened 2025-12-07
Extract precision API fallback with type-safe enum

### LINKED ISSUE #3274: Remove Duplicate Code and Hardcoded Strings in RedisTemplate
## Summary

  Extract duplicated precision command fallback logic from `RedisTemplate` into a reusable utility class with type-safe enum constants.

  ## Current Problem

  The same try-catch pattern is duplicated across 3 methods in `RedisTemplate`:

  ```java
  // This appears in expire(), expireAt(), and getExpire()
  try {
      return connection.pExpire(rawKey, rawTimeout);
  } catch (Exception ignore) {
      return connection.expire(rawKey, TimeoutUtils.toSeconds(timeout, unit));
  }

------- after -------

  Create a utility class PrecisionApiHelper with an enum for type-safe command names:

  // New approach - single reusable method
  return doWithKeys(connection -> PrecisionApiHelper.withPrecisionFallback(
      PrecisionCommand.PEXPIRE,
      () -> connection.pExpire(rawKey, rawTimeout),
      () -> connection.expire(rawKey, TimeoutUtils.toSeconds(timeout, unit))
  ));

youngsuk-kim:
## Refactor: Extract precision API fallback with type-safe enum

  Remove 66 lines of duplicate try-catch code from RedisTemplate.
  Replace hardcoded strings with PrecisionCommand enum.
  Add logging and specific exception handling.

  No functional changes, backward compatible.

  ### Changes
  - New: `PrecisionApiHelper.java` with `PrecisionCommand` enum
  - Modified: `RedisTemplate.java` - updated `expire()`, `expireAt()`, `getExpire()`

  ### Testing
  - All existing tests pass
  - No behavior changes

Closes #3274

  ---

  - [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don't submit any formatting related changes.
  - [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current
  year only).

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
+++ b/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
@@ -58,6 +58,8 @@ import org.springframework.data.redis.serializer.JdkSerializationRedisSerializer
 import org.springframework.data.redis.serializer.RedisSerializer;
 import org.springframework.data.redis.serializer.SerializationUtils;
 import org.springframework.data.redis.serializer.StringRedisSerializer;
+import org.springframework.data.redis.util.PrecisionApiHelper;
+import org.springframework.data.redis.util.PrecisionApiHelper.PrecisionCommand;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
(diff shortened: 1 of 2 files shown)


onobc:
Thank you for the code proposal @youngsuk-kim. I am a huge fan of the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and I ❤️  the de-duplication in the code proposal.

However, there are a couple of other factors to consider:

1. As noted in the previous [code comment](https://github.com/spring-projects/spring-data-redis/pull/3275/files#diff-94b73a347af23d52a16751da0255e2784ff530921967b81ff4a7d2568957fb7bL699) this fallback was only required for Redis `2.4` and our minimum version is now `2.8`. Ideally we should have removed this fallback quite some time ago.

2. Regardless of the above, we always use caution when introducing a public API because once you provide it, you can't take it away, you can only deprecate it slowly. This can (and usually does) turn into maintenance toil.  In other words, it would introduce less risk to make the util either package-protected OR just inline the helper method in `RedisTemplate` (which would solve the same duplication problem).

Ultimately though, the first point is a signal for us to **remove** this cruft from the code rather than **make it better**.

Again, thank you very much for the PR as it brought this fallback into the light. I am going to close this PR and will then follow up w/ a commit that removes the fallback.

Issue #3271, opened 2025-11-29
JacksonJsonRedisSerializer in Spring Data Redis 4.0.0 breaks deserialization due to use of unsafeSimpleType


duyvhh:
In **Spring Data Redis 3.5.x**, `Jackson2JsonRedisSerializer(ObjectMapper, Class<T>)` correctly deserializes JSON using:

```java
TypeFactory.defaultInstance().constructType(clazz);
```

In **4.0.0**, the new `JacksonJsonRedisSerializer(ObjectMapper, Class<T>)` instead uses:

```java
TypeFactory.unsafeSimpleType(clazz);
```

This produces an incomplete `JavaType` and causes deserialization to fail for simple DTOs that previously worked.

### **Suggestion**

Replace:

```java
TypeFactory.unsafeSimpleType(clazz)
```

with:

```java
TypeFactory.defaultInstance().constructType(clazz)
```

### **Environment**

* Spring Data Redis: **4.0.0**
* Jedis: **7.1.0**
* Jackson: **3.0.2**
* Java:  21
* RedisTemplate with `JacksonJsonRedisSerializer`

christophstrobl:
If you'd like us to spend some time investigating, please take the time to provide a complete [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) (something that we can unzip or git clone, build, and deploy) that reproduces the problem.

duyvhh:
I've setup a demo project here

https://github.com/duyvhh/sdr4issue

thanks

christophstrobl:
@duyvhh thank you for reporting and the reproducer 👍

PR #3277, opened 2025-12-11
Use Jackson default factory to create JavaType.

christophstrobl:
Should be back ported to 4.0.x

Closes: #3271

### DIFF
--- a/src/test/java/org/springframework/data/redis/serializer/JacksonJsonRedisSerializerTests.java
+++ b/src/test/java/org/springframework/data/redis/serializer/JacksonJsonRedisSerializerTests.java
@@ -23,7 +23,6 @@ import tools.jackson.databind.type.TypeFactory;
 
 import java.io.Serializable;
 import java.util.Arrays;
-import java.util.Objects;
 import java.util.UUID;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


src/test/java/org/springframework/data/redis/serializer/JacksonJsonRedisSerializerTests.java
```diff
@@ -0,0 +1,129 @@
...
+
+		JacksonJsonRedisSerializer<SessionToken> redisSerializer = new JacksonJsonRedisSerializer<>(SessionToken.class);
+
+		SessionToken source = new SessionToken(UUID.randomUUID().toString());
+
+		byte[] serialized = redisSerializer.serialize(source);
+		assertThat(redisSerializer.deserialize(serialized)).isEqualTo(source);
+	}
+
+	@Test // DATAREDIS-241
```
onobc, line 44: Why is this showing up as a new file but it has previous tests in it?

onobc: Nm - I see it was copied from the Jackson 2 variant.

src/test/java/org/springframework/data/redis/serializer/JacksonJsonRedisSerializerTests.java
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @author Thomas Darimont
+ * @author Christoph Strobl
+ * @author Mark Paluch
+ */
+class JacksonJsonRedisSerializerTests {
+
+	private JacksonJsonRedisSerializer<Person> serializer = new JacksonJsonRedisSerializer<>(Person.class);
+
+	@Test // GH-3271
```
onobc: NOTE: This test as-is does not reproduce the problem (i.e. If you revert the code change it still passes). I figured out the difference between this test and the reproducer...

It is how the POJO is defined (`SessionToken`). The code as written previously works when the POJO follows the standard `private field w/ getter and setter` pattern.  Like the `SessionToken` POJO you have created in the test:
```java
	static class SessionToken implements Serializable {

		private String userUuid;

		private SessionToken() {
			// why jackson?
		}

		public SessionToken(String userUuid) {
			this.userUuid = userUuid;
		}

		public String getUserUuid() {
			return userUuid;
		}

		public void setUserUuid(String userUuid) {
			this.userUuid = userUuid;
		}

		@Override
		public boolean equals(Object o) {
			if (o == this) {
				return true;
			}
			if (o == null || getClass() != o.getClass()) {
				return false;
			}
			SessionToken token = (SessionToken) o;
			return Objects.equals(userUuid, token.userUuid);
		}

		@Override
		public int hashCode() {
			return Objects.hash(userUuid);
		}

		@Override
		public String toString() {
			return "SessionToken{" + "userUuid='" + userUuid + '\'' + '}';
		}

	}
```
The above works as-is w/o the changes in the PR. It also works w/ the changes in the PR.

However, it does not work when the POJO has no getter/setter but has public field (as in the reproducer).
```java

	@JsonIgnoreProperties(ignoreUnknown = true)
	public static class SessionToken2 implements Serializable {

		public String userUuid;

		private SessionToken2() {
		}

		public SessionToken2(String userUuid) {
			this();
			this.userUuid = userUuid;
		}
	}
```
This variant fails w/o the old code but passes w/ the new code.

I am going to update the tests accordingly and keep the PR changes. Just passing on this info.

PR #3287, opened 2026-01-05
Remove unused generic from RedisKeyExpiredEvent #2106

### LINKED ISSUE #2106: RedisKeyExpiredEvent has an unused generic
Current behavior: [`RedisKeyExpiredEvent`](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/core/RedisKeyExpiredEvent.java#L33) has an unused generic argument `T`.

Expected behavior is one of the following:
* `RedisKeyExpiredEvent` does not have a generic argument.
* Method [`getValue()`](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/core/RedisKeyExpiredEvent.java#L100-L108) in `RedisKeyExpiredEvent` returns an object of the generic specified in `RedisKeyExpiredEvent`

gisu1102:
Closes #2106

### Motivation

`RedisKeyExpiredEvent<T>` declares a generic type parameter that is not used anywhere in the current API.
The presence of `<T>` suggests type safety that the event does not actually provide, which can be confusing for users.

### Changes

- Removed the unused generic type parameter from `RedisKeyExpiredEvent`
- Kept the existing runtime behavior unchanged (`getValue()` continues to return `Object`)

### Rationale

The original issue outlines two possible directions.
This PR intentionally takes the minimal approach by removing the unused generic parameter.

Implementing a typed `getValue(): T` would require additional design decisions around
value deserialization and type resolution, which felt out of scope for a small cleanup.
Removing the unused generic improves API clarity without introducing new semantics.

### Compatibility

This is a source-level incompatible change for code explicitly using `RedisKeyExpiredEvent<T>`.
However, since the generic parameter was unused and did not provide actual type safety,
the migration impact is minimal and runtime behavior remains unchanged.

### Tests

- Existing unit tests pass

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/RedisKeyExpiredEvent.java
+++ b/src/main/java/org/springframework/data/redis/core/RedisKeyExpiredEvent.java
@@ -30,7 +30,7 @@ import org.springframework.data.redis.core.convert.MappingRedisConverter.BinaryK
  * @author Mark Paluch
  * @since 1.7
  */
-public class RedisKeyExpiredEvent<T> extends RedisKeyspaceEvent {
+public class RedisKeyExpiredEvent extends RedisKeyspaceEvent {
 
 	/**
 	 * Use {@literal UTF-8} as default charset.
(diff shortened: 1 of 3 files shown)


mp911de:
That's an API change that we need to wait with, otherwise, existing application code will fail to compile.

gisu1102:
Thanks for the clarification. 
Understood — this is an API-breaking change and should be deferred to the next major release. Happy to revisit when work on the next major version starts.

PR #3302, opened 2026-01-23
Fix Duration validation to accept nanosecond fractions

### LINKED ISSUE #2975: ReactiveListOperations left/right pop operations fail with Duration with fractions of second
According to documentation, leftPop/rightPop methods expect to get Duration parameter ZERO or >= 1 second.

Failure user case: pass Duration with seconds >= 1 second, and nanos > 0.
Example:
```java
reactiveRedisTemplate.opsForList().leftPop(key, Duration.ofMillis(10100L));
```
Fails with exception
```
java.lang.IllegalArgumentException: Duration must be either zero or greater or equal to 1 second
```

The issue is caused by wrong implementation of [DefaultReactiveListOperations.isZeroOrGreaterOneSecond()](https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/core/DefaultReactiveListOperations.java#L335)

chanani:
Fixes #2975

## Summary
This PR fixes a bug in `ReactiveListOperations.leftPop()` and `rightPop()` methods where Duration values >= 1 second with nanosecond fractions (e.g., 1.5 seconds, Duration.ofMillis(1001)) were incorrectly rejected.

## Problem
The previous implementation checked if the nanosecond component was exactly zero:
```java
timeout.getNano() % TimeUnit.NANOSECONDS.convert(1, TimeUnit.SECONDS) == 0
```
This incorrectly rejected valid Duration values like:
- `Duration.ofMillis(1500)` (1.5 seconds)
- `Duration.ofSeconds(1).plusNanos(1)` (1.000000001 seconds)

## Solution
Changed the validation to compare the entire Duration value:
```java
timeout.compareTo(Duration.ofSeconds(1)) >= 0
```

This correctly:
- ✅ Accepts `Duration.ZERO`
- ✅ Accepts any Duration >= 1 second (including nanosecond fractions)
- ✅ Rejects Duration < 1 second

## Testing
Added comprehensive tests covering:
- Duration with nanoseconds (1.5 seconds)
- Duration created from millis (1001ms)
- Duration exactly 1 second
- Duration less than 1 second (still rejected as expected)

All existing tests pass.

## Notes
This is a bug fix only (Option 1 from maintainer feedback). The API requirement remains "Duration.ZERO or >= 1 second". 

Widening the API to accept all non-negative values (Option 2) will be addressed separately as it requires an API change.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/DefaultReactiveListOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/DefaultReactiveListOperations.java
@@ -337,7 +337,7 @@ class DefaultReactiveListOperations<K, V> implements ReactiveListOperations<K, V
 	}
 
 	private boolean isZeroOrGreaterOneSecond(Duration timeout) {
-		return timeout.isZero() || timeout.getNano() % TimeUnit.NANOSECONDS.convert(1, TimeUnit.SECONDS) == 0;
+		return timeout.isZero() || timeout.compareTo(Duration.ofSeconds(1)) >= 0;
 	}
 
 	private ByteBuffer rawKey(K key) {


onobc:
👋🏻 @chanani 
Thank you for a great contribution. Nice job! I added a small polishing commit here that moves as many tests as possible from integration to unit tests (https://github.com/spring-projects/spring-data-redis/commit/73441151e956d4c9c9071ef6cf5c7f000086c424). I am closing this via your commit https://github.com/spring-projects/spring-data-redis/commit/cce3fbda00cac587e2571e48bf7940eb8c180465

Issue #1004, opened 2015-08-31
Introduce annotation-driven Redis listener endpoint programming model


spring-projects-issues:
**[Greg Turnquist](https://jira.spring.io/secure/ViewProfile.jspa?name=gregturn)** opened **[DATAREDIS-428](https://jira.spring.io/browse/DATAREDIS-428?redirect=false)** and commented

Listener configuration in Spring Data Redis requires always to define a `MessageListener` because Spring Data Redis has no built-in support for an annotation-driven configuration like it is handled JMS or AMQP. Users have to obtain/configure the `RedisMessageListenerContainer` and to register their own `MessageListener`. `MessageListenerAdapter` leverages the requirement for a fixed method signature by allowing to pass a delegate to handle the message event but the infrastructure setup of the listener container is still necessary. Using the `MessageListener` itself puts a high burden on the developer to consume events because everything is in byte[]'s. Allowing to configure a listener endpoint like it is done with JMS would lighten the load on developers. A possible improvement could look like:

```
 
public class MyListenerClass {

    @RedisListener(id = "patternEndpoint", pattern = {"/topic/*", "__keyevent@*__:expired"})
    void handleOnPattern(String pattern, Message message) {
         // ...
    }

    @RedisListener(channel = "/channels/channel1")
    void handleOnChannel(String channel, Message message) {
        // ...
    }

}
```

---

**Issue Links:**
- [DATAREDIS-1026](https://jira.spring.io/browse/DATAREDIS-1026) Annotation-driven activation of RedisListeners for Pub/Sub
 (_**"is duplicated by"**_)
- [DATAREDIS-397](https://jira.spring.io/browse/DATAREDIS-397) Investigate creating `@RedisListener` for messaging
 (_**"supersedes"**_)

Muerti:
Is there any activity on this issue?

Ilyass-Bougati:
Thanks for consolidating the threads, @mp911de .

I completely agree that the challenge here is landing on the right model rather than just wiring things up.

To make this consistent with the rest of the Spring ecosystem, I think the design should mirror the patterns already present in Spring JMS and AMQP. Specifically, adopting the standard listener container factory pattern seems like the right move. It would provide that "approachable" configuration model while still allowing for granular control (like specific thread pools or serializers) per listener, rather than locking everyone into a global default.

If that architectural direction (specifically aligning with the standard listener factory model) sounds right to you, I’d be happy to put together a PR for the team to review.

[...]

PR #3303, opened 2026-01-23
Add support for annotation-driven Redis Pub/Sub Listeners

Ilyass-Bougati:
Relates to #1004

### Purpose
Following the discussion in #1004 and the suggestion by @onobc, this PR introduces the high-level architectural skeleton for the annotation-driven listener model.

The goal is to align on the core components and interfaces before filling in the implementation logic.

### Design Overview
To ensure consistency with the wider Spring ecosystem, this design strictly mirrors the **Listener Container Factory** pattern found in `spring-jms` and `spring-amqp`. This approach decouples the listener model from the execution, allowing for granular configuration (thread pools, serializers) per listener.

**Introduced Components:**
* **`RedisListenerContainerFactory`**: Strategy interface for creating `RedisMessageListenerContainer` instances.
* **`RedisListenerEndpoint`**: The model describing a listener configuration.
* **`RedisListenerEndpointRegistry`**: Lifecycle manager for the created containers.
* **`RedisListenerAnnotationBeanPostProcessor`**: Infrastructure bean that detects `@RedisListener` methods and registers endpoints.
* **`@EnableRedisListeners`**: Configuration annotation to activate the post-processor.

### Note on Testing
Per the discussion in #1004, this is a high-level design proposal containing only interfaces and skeleton classes. **No test cases are included in this PR** as there is no implementation logic yet. Comprehensive unit and integration tests will be added in the subsequent implementation PRs.

### DIFF
--- a/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
+++ b/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
@@ -409,6 +409,7 @@ public class RedisMessageListenerContainer implements InitializingBean, Disposab
 
 		while (!doSubscribe(backOffExecution)) {
 			// busy-loop, allow for synchronization against doUnsubscribe therefore we want to retry.
+			Thread.onSpinWait();
 			containerListenFuture = this.listenFuture;
 		}
 
@@ -500,6 +501,7 @@ public class RedisMessageListenerContainer implements InitializingBean, Disposab
 
 		while (!doUnsubscribe()) {
 			// busy-loop, allow for synchronization against doSubscribe therefore we want to retry.
+			Thread.onSpinWait();
 		}
 	}
 


mp911de:
Can you sketch out the declaration side of this change (i.e. how would application code look like) so we have an idea how this is going to be used? With such an approach, we can get an idea which components we need to build and how to approach certain aspects of invocation, exception handling and return value handling.

Ilyass-Bougati:
Sure, here is the sketch of the declaration side.

## Configuration
The user enables the infrastructure and defines a `RedisListenerContainerFactory`. This factory is the central place to configure serialization, thread pools, and error handling.
```java
@Configuration
@EnableRedisListeners
public class RedisListenerConfig {

    @Bean
    public RedisListenerContainerFactory<?> redisListenerContainerFactory(RedisConnectionFactory connectionFactory) {
        DefaultRedisListenerContainerFactory factory = new DefaultRedisListenerContainerFactory();
        factory.setConnectionFactory(connectionFactory);
        
        // Custom Serialization
        factory.setObjectMapper(new ObjectMapper()); 
        
        // Global Error Handling (e.g. logging or dead-lettering)
        factory.setErrorHandler(t -> System.err.println("Listener error: " + t.getMessage()));
        
        return factory;
    }
}
```

## Application Code (The Listener)
The developer can simply annotate methods. The infrastructure handles the argument resolution (converting `byte[]` to `String` or POJO) and dispatch.
```java
@Service
public class OrderService {

    // Simple String listener on a specific channel
    @RedisListener(topics = "orders.logs")
    public void logOrder(String logMessage) {
        System.out.println("Received: " + logMessage);
    }

    // POJO listener with automatic deserialization (via Jackson/Converter)
    @RedisListener(topics = "orders.created", containerFactory = "myCustomFactory")
    public void createOrder(OrderEvent event) {
        // 'event' is automatically deserialized from the Redis payload
        orderRepository.save(event);
    }

    // Pattern matching listener
    @RedisListener(topicPatterns = "orders.*")
    public void monitorOrders(Message message) {
        // Access raw Spring Data Redis Message object if needed
        System.out.println("Activity on channel: " + new String(message.getChannel()));
    }
}
```

## Handling Invocation & Exceptions
- **Invocation:** The `MessagingMessageListenerAdapter` will handle the reflection and argument resolution (supporting `Message`, `byte[]`, `String`, and POJOs).
- **Exceptions:** Exceptions thrown by the listener method are caught by the container. By default, they are logged, but the user can provide a standard `org.springframework.util.ErrorHandler` to the `ContainerFactory` to customize this behavior (e.g., sending to a dead letter topic).
- **Return Values:** For Pub/Sub, the return type is typically void. However, adopting the `MessagingMessageListenerAdapter` architecture allows us to support `@SendTo` in the future if we want to support a Request-Reply style pattern (publishing the return value to another topic).

mp911de:
I'm getting pretty much the impression talking to ChatGPT as the descriptions and comments are a wall of text sumarizing themselves discussing obvious details from the JMS listeners without considering Redis-specifics. 

The actually interesting questions are:

* How should a listener method look like for patterns and for channels
* Should the same method be used for both? Why if so? Why if not?
* What argument types can be injected into a method?
* How are these supposed to be deserialized? How is that going to be controlled?

If you would like us to spend some time collaborating with you, please spend some time and effort. We appreciate community contributions, but we do need them to meet a reasonable baseline for consideration. This means:

* Understanding the existing code structure and following established patterns
* Providing clear suggestions that match the feature idea problem space
* Ensuring the contribution aligns with Spring's design principles

Writing the code is not a problem to be solved, we can throw a few sentences into an LLM ourselves. Listening, understanding, and activiating creativity to address a genuine concern is what matters when developing libraries for a broader audience. 

When contributions require significant rework from our side, we have to deprioritize or even reject them in favor of changes that are ready for integration as we have to be mindful of our limited time, and we need to focus it where it has the most impact. If you would like us to spend some time working on this pull request, please spend some time making this a substantial and quality contribution.

Ilyass-Bougati:
Point taken. I appreciate the direct feedback regarding the contribution baseline; I intended to align on the high-level API first, but I see now that glossing over the runtime specifics made the proposal feel abstract.

To answer your technical constraints directly:

A single `@RedisListener` annotation should suffice for both fixed channels and patterns. The container can inspect the annotation to determine whether to issue a `SUBSCRIBE` or `PSUBSCRIBE` command, effectively abstracting that complexity away from the user. Consequently, the same method signature should support both use cases; the user typically cares only about the payload, not the underlying subscription mechanism. If they do need metadata, such as the source channel or the pattern matched, they can simply opt to inject the `org.springframework.data.redis.connection.Message` envelope.

On the data side, we need to support `byte[]`, `String`, and `Message` injection out of the box, but POJO support is the primary challenge due to the lack of headers in Redis. Unlike JMS or AMQP, we cannot rely on a Content-Type header to infer serialization. Therefore, smart inference is fragile here. We must rely on strict configuration: a default serializer at the `ContainerFactory` level for the common case, and an explicit override attribute on the annotation for specific listeners (e.g., handling Protobuf on one topic and JSON on another).

I will pivot the PR to focus entirely on this argument resolution and deserialization strategy, as that is clearly where the complexity lies.

mp911de:
Thanks for your suggestion.

So a single annotation to indicate pattern or channel subscriptions:

* `@RedisListener(topic = {"channel1", "channel-foo"})`
* `@RedisListener(topic = "my-pattern*")`
* Other naming ideas (see Redis Streams section at the end): `@RedisPubSubListener(topic = "my-pattern*")`, `@RedisChannelListener(topic = "my-channel")`, `@RedisPatternListener(topic = "my-pattern*")`)?

These could be also mixed and we would define once there are placeholders, we're using Pattern subscriptions.

Injectable argument types:

* `org.springframework.data.redis.connection.Message`
* Simple message body types: `byte[]` and `String`
* `org.springframework.data.redis.listener.Topic`

I think we should also consider `@RedisExceptionHandler` (similar to `@MessageExceptionHandler`) do define an error handler that is valid for exception handling of errors within the same class:

```java
@Controller
public class MyController {

	// ...

	@RedisExceptionHandler
	public void handleException(MyException exception) {
		// ...
	}
}
```

I think this arrangement is sufficient for the very initial step to get something going and already bears quite some complexity. 

Spring Framework provides a `@Payload` annotation in its messaging module to indicate payload/body details. It would make sense to postpone the serializer issue to a later step so that we don't overwhelm ourselves with too much complexity in the first step. Once we have a good setup for listener registration/de-registration and method invocation then we can evolve from there to a model for serializer selection. It's always easier to have some material for experiments.

Let's take another step back to see the greater picture. Right now, we're discussing Redis Pub/Sub. In that context, `@RedisListener` is unique and clearly describes its intent. Redis provides with Redis Streams another mechanism of messaging and it would make sense to spend a few cycles how an annotation-driven programming model for Redis Stream consumption could look like. The intent clearly is not to build such support with this pull request, but rather have this distinction in mind for naming components so that a later Redis Stream listener model would not suffer from poor naming.

Ilyass-Bougati:
That sounds like a solid plan. I agree that keeping the scope tight (focusing on the registration and invocation mechanics while sticking to simple types first) is the best way to stabilize the core before tackling the serialization ambiguity.

Regarding the naming, I think sticking with `@RedisListener` for Pub/Sub is the right call, especially to mirror `@JmsListener`. Redis Streams are distinct enough (with consumer groups, offsets, and acks) that they'll likely warrant a dedicated `@RedisStreamListener` down the road to handle those specific attributes without cluttering the Pub/Sub contract. That said, if you think the distinction needs to be explicit right now to prevent future ambiguity, I’m happy to refactor it to `@RedisPubSubListener`.

The addition of `@RedisExceptionHandler` also makes perfect sense and aligns well with the standard Spring method-handling patterns for local error resolution. I'll update the PR to implement the `RedisListenerAnnotationBeanPostProcessor` and the `MessagingMessageListenerAdapter` with support for `byte[]`, `String`, and `Message` types, leaving the complex content negotiation for the next iteration.

Ilyass-Bougati:
Hi @mp911de , just a gentle follow-up on this.

Per your earlier suggestion to postpone the complex serialization logic, I have pushed the core infrastructure implementation for the listener registration and invocation mechanics.

I wanted to get your "green light" on this foundational architecture before I proceed with adding serializers and error handlers.

mp911de:
I've been out for a week and this week, we're busy with release preparations. This change goes into an appropriate direction. We can initially simplify the arrangement by dropping `RedisListenerContainerFactory`. A `RedisMessageListenerContainer` should be provided ideally through the application context. Contrasting to JMS, in JMS a listener uses a single destination and a message selector while in Redis a container can subscribe to multiple channels/patterns.

Also, please rebase your pull request. It contains some intermediate changes that are already available on the `main` branch.

I suggest implementing some unit tests that do not require a running Redis instance to reach a point where the endpoint can be tested by notifying a listener through a message.

Ilyass-Bougati:
@mp911de Updates pushed.

I’ve rebased and refactored the infrastructure to align with the simplified design. The `RedisListenerContainerFactory` is gone; instead, the post-processor now wires endpoints directly to an existing `RedisMessageListenerContainer` bean (defaulting to `redisMessageListenerContainer`). I also updated the registry to track these container references for proper lifecycle management simply, and added the disconnected unit tests to verify the wiring without a running Redis instance.

src/main/java/org/springframework/data/redis/annotation/RedisListenerConfigurationSelector.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Selects the {@link RedisListenerBootstrapConfiguration} to register the necessary infrastructure beans for
+ * {@link RedisListener} annotation processing.
+ *
+ * @author Ilyass Bougati
+ * @see RedisListenerBootstrapConfiguration
+ * @see EnableRedisListeners
+ */
+public class RedisListenerConfigurationSelector implements DeferredImportSelector {
```
mp911de: We don't require a selector, importing the configuration instead of the selector is sufficient.

src/main/java/org/springframework/data/redis/annotation/MethodRedisListenerEndpoint.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * A {@link RedisListenerEndpoint} providing the method invocation mechanism for a specific
+ * bean method.
+ *
+ * @author Ilyass Bougati
+ * @see MessagingMessageListenerAdapter
+ */
+public class MethodRedisListenerEndpoint implements RedisListenerEndpoint, BeanFactoryAware {
+
+	private Object bean;
```
mp911de: Let's make the method and bean properties final. 

Also, we should make the endpoint aware of its listener.
This class should also be moved into the `config` package and made `SmartLifecycle`. The lifecycle controls registration (subscription) with its associated container.

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistry.java
```diff
@@ -0,0 +1,94 @@
...
+			if (!container.isRunning()) {
+				container.start();
+			}
+		}
+		this.running = true;
+	}
+
+	@Override
+	public void stop() {
+		for (RedisMessageListenerContainer container : this.containers) {
```
mp911de: We must not control the `RedisMessageListenerContainer` lifecycle as the container bean is expected to be managed externally.

In contrast to JMS, Redis does not have various flavors of resources (Spring, JCA, Jakarta), but we only have a single Redis variant that simplifies the overall design.

It would in any case make sense to make `RedisListenerEndpoint` a `SmartLifecycle` and propagate the lifecycle into these objects so that an endpoint can register and unregister itself with the container.

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistry.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	@Override
+	public boolean isRunning() {
+		return this.running;
+	}
+
+	@Override
+	public void destroy() throws Exception {
+		for (RedisMessageListenerContainer container : this.containers) {
```
mp911de: See above regarding lifecycle.

src/main/java/org/springframework/data/redis/annotation/MessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,80 @@
...
+ * <p>Supports argument resolution for:
+ * <ul>
+ * <li>{@link Message} (The raw Redis message)</li>
+ * <li>{@code byte[]} (The message body)</li>
+ * <li>{@link String} (The message body decoded as UTF-8)</li>
+ * </ul>
+ *
+ * @author Ilyass Bougati
+ */
+public class MessagingMessageListenerAdapter implements MessageListener {
```
mp911de: We might using Spring Messaging's `HandlerMethod` going forward as Spring Messaging provides quite some infrastructure for `HandlerMethodArgumentResolver` and `MessagePostProcessor` along with `Message`. 

Also, this class should go into the `listener.adapter` package.

src/main/java/org/springframework/data/redis/annotation/MethodRedisListenerEndpoint.java
```diff
@@ -0,0 +1,114 @@
...
+import org.springframework.util.StringValueResolver;
+
+/**
+ * A {@link RedisListenerEndpoint} providing the method invocation mechanism for a specific
+ * bean method.
+ *
+ * @author Ilyass Bougati
+ * @see MessagingMessageListenerAdapter
+ */
+public class MethodRedisListenerEndpoint implements RedisListenerEndpoint, BeanFactoryAware {
```
mp911de: This class sholud be in the `config` package.

src/main/java/org/springframework/data/redis/annotation/RedisListener.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * The topics for this listener.
+     */
+    String[] topics() default {};
+
+    /**
+     * The topic patterns for this listener.
+     */
+    String[] topicPatterns() default {};
```
mp911de: just `patterns`

src/main/java/org/springframework/data/redis/annotation/RedisListener.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * The bean name of the {@link org.springframework.data.redis.config.RedisListenerContainerFactory}
+     * to use to create the message listener container.
+     */
+    String containerFactory() default "";
+
+    /**
+     * The topics for this listener.
+     */
+    String[] topics() default {};
```
mp911de: just `channels`

src/main/java/org/springframework/data/redis/annotation/RedisListener.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * The unique identifier of the container managing this endpoint.
+     */
+    String id() default "";
+
+    /**
+     * The bean name of the {@link org.springframework.data.redis.config.RedisListenerContainerFactory}
+     * to use to create the message listener container.
+     */
+    String containerFactory() default "";
```
mp911de: We should refer to the container bean name here, not a container factory. If the bean name is empty, we would try to look up a default `RedisMessageListenerContainer` bean.

src/main/java/org/springframework/data/redis/annotation/MessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,80 @@
...
+
+	private Object bean;
+	private Method method;
+
+	public void setHandlerMethod(Object bean, Method method) {
+		this.bean = bean;
+		this.method = method;
+		ReflectionUtils.makeAccessible(this.method);
```
mp911de: FWIW, Spring's `AnnotatedMethod` already takes care of accessibility flags.

mp911de (changes requested):
Did a more review pass also on the design. It would make a lot of sense to depend on Spring Messaging as the infrastructure components there align nicely with our approach and we would be able to reuse several components and its general programming model along with its interfaces.

src/main/java/org/springframework/data/redis/annotation/RedisListenerConfigurationSelector.java
Ilyass-Bougati: Done.

src/main/java/org/springframework/data/redis/annotation/MethodRedisListenerEndpoint.java
Ilyass-Bougati: Done. Moved to `config` package, implemented `SmartLifecycle` for registration control, and made the core properties final.

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistry.java
Ilyass-Bougati: Updated. The registry now propagates lifecycle signals to endpoints only and no longer attempts to manage the externally-provided container's lifecycle.

src/main/java/org/springframework/data/redis/annotation/MethodRedisListenerEndpoint.java
Ilyass-Bougati: Done

src/main/java/org/springframework/data/redis/annotation/RedisListener.java
Ilyass-Bougati: Done

src/main/java/org/springframework/data/redis/annotation/RedisListener.java
Ilyass-Bougati: Repurposed `containerFactory` to `container`

src/main/java/org/springframework/data/redis/annotation/RedisListener.java
Ilyass-Bougati: Done

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistry.java
Ilyass-Bougati: I removed the `destroy` method and stopped implementing `DisposableBean`

src/test/java/org/springframework/data/redis/annotation/RedisListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.redis.config.RedisListenerEndpointRegistry;
+import org.springframework.data.redis.listener.RedisMessageListenerContainer;
+
+/**
+ * Integration test for {@link EnableRedisListeners} and {@link RedisListener}
+ *
+ * @author Ilyass Bougati
+ */
+class RedisListenerAnnotationBeanPostProcessorTests {
```
mp911de: We should have proper unit tests to test the bean post-processor in isolation. The integrative aspect is to come later on.

src/test/java/org/springframework/data/redis/annotation/MessagingMessageListenerAdapterTest.java
```diff
@@ -0,0 +1,96 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.redis.annotation;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.mockito.BDDMockito.given;
```
mp911de: Please use regular Mockito, not BDD Mockito.

Ilyass-Bougati:
@mp911de 

All set! I've addressed the final test feedback, squashed the commits, and force-pushed the clean history. Thank you for the incredibly thorough review and guidance throughout this PR, Mark. I learned a ton. Feel free to take it from here!

mp911de:
I added a commit on top of your changes. Using Spring Messaging turns out to be a good direction, we're getting a lot of out-of-the-box support such as injection of headers and conversion support for message payloads. I don't have yet a good idea for how to control which message converter to use, especially that Redis Pub/Sub does not have a concept of headers.

Going forward, Spring Message features exception handing and reply-to concepts (`@MessageExceptionHandler`, `@SendTo`). We could stretch `@RedisListener` towards `produces` and `consumes` and with the idea of having repeatable annotations, this could play nicely to support various encodings that come from different channels. In any case, the next step is a review from the rest of our team before we can merge the issue (also, we need documentation and I'll take care of this).

Once merged, are you interested in connecting the dots towards exception handling, serializer selection and reply to? It would make sense to introduce a `JmsListenerEndpointRegistrar` model for extended configurability.

mp911de:
I just messed up this PR with a bad force-push. You can find the code at https://github.com/spring-projects/spring-data-redis/tree/annotation-pubsub-listener

Ilyass-Bougati:
Haha, no worries at all @mp911de! It happens to the best of us.

And yes, I'm absolutely interested in continuing the work on this feature and connecting the dots. Let me know when the team finishes their review on this phase, and I'll start looking into the architecture for the registrar.

mp911de:
Thank you very much for your contribution. It was a lot of joy to see all the pieces falling into place. That's merged and polished now. Feel free to explore the above mentioned themes. Depending on the progress, we would be able to ship the feature along with some documentation (we will take care of the docs once all things are in place) with the next milestone in mid March.

Ilyass-Bougati:
Awesome, glad to see it officially merged!

I'll start digging into the `JmsListenerEndpointRegistrar` architecture and exception handling to help hit that mid-March target. I'll open a new PR once I have the foundation laid out.

mp911de:
Sounds good. `DefaultMessageHandlerMethodFactory` exposes quite a few configuration aspects such as `ConversionService`, `MessageConverter`, `Validator`, and `HandlerMethodArgumentResolver`. 

Especially for the converter I had to configure a specific converter. I think the difficulty is in finding a good balance between providing either a `MessageHandlerMethodFactory` or registering customizations that can contribute converters to `ConversionService`, `MessageConverter` and `HandlerMethodArgumentResolver`.

We would want to pre-configure `byte[]` converters with a specific `Charset` and `MessageConverter`s based on class path availability. In addition, we want to base our content type resolution on something that is provided through `@RedisListener(produces/consumes)` requiring a customized `ContentTypeResolver` to be provided to `AbstractMessageConverter` types.

Ilyass-Bougati:
That makes complete sense. I really like the approach of basing the content type resolution on `@RedisListener(consumes)` backed by a customized `ContentTypeResolver`. It keeps the API clean for the user while allowing us to fall back to smart classpath defaults for the converters.

I'll spend the weekend drafting this up along with the `RedisListenerEndpointRegistrar` integration and open a new PR. Have a great weekend!

PR #3304, opened 2026-01-26
Support `SET` CAS operations `IFEQ` and `IFNE`

Dgramada:
This PR introduces support for the Redis `SET` command's `IFEQ` and `IFNE` options, which were added in Redis 8.4 to enable atomic compare-and-swap (CAS) operations.

### Summary

Introduces a new `SetCondition` class in `RedisStringCommands` that provides factory methods for conditional SET operations:

- `upsert()` - Always set the value (default behavior)
- `ifAbsent()` - Set only if key does not exist (`NX`)
- `ifPresent()` - Set only if key exists (`XX`)
- `ifValueEqual(byte[])` - Set only if current value equals the specified value (`IFEQ`)
- `ifValueNotEqual(byte[])` - Set only if current value does not equal the specified value (`IFNE`)

### Changes

**Core API:**
- Added `SetCondition` class with immutable, cached instances for stateless conditions
- Extended `set()` and `setGet()` method overloads in `RedisStringCommands` to accept `SetCondition`
- Added support in `StringRedisConnection` and `DefaultStringRedisConnection`

**Driver Implementations:**
- Implemented Jedis support (`JedisStringCommands`, `JedisClusterStringCommands`, `JedisConverters`)
- Implemented Lettuce support (`LettuceStringCommands`, `LettuceConverters`)
- Implemented reactive support (`LettuceReactiveStringCommands`, `ReactiveStringCommands`)

**Template Operations:**
- Extended `ValueOperations` and `BoundValueOperations` with new `set()` and `setGet()` methods
- Extended `ReactiveValueOperations` with reactive equivalents
- Added Kotlin extensions for `ReactiveValueOperations`

**Tests:**
- Added comprehensive unit tests for converters
- Added integration tests for both standalone and cluster configurations
- Added reactive integration tests

### Compatibility

- Requires Redis 8.4+ for `IFEQ`/`IFNE` functionality
- Maintains backward compatibility with existing `SetOption` API
- All existing tests continue to pass


src/main/java/org/springframework/data/redis/connection/RedisStringCommands.java
```diff
@@ -392,48 +395,187 @@ default Long bitPos(byte @NonNull [] key, boolean bit) {
...
 	 *
 	 * @author Christoph Strobl
+	 * @see <a href="https://redis.io/commands/set">Redis SET command</a>
 	 * @since 1.7
 	 */
-	enum SetOption {
+	class SetOption {
+
+		// Cached instances for stateless conditions
+		private static final SetOption UPSERT_INSTANCE = new SetOption(Type.UPSERT, null);
```
mp911de: Let's keep these constants `public` to retain compile compatibility. No need to add `_INSTANCE` to the name.

src/main/java/org/springframework/data/redis/connection/RedisStringCommands.java
```diff
@@ -392,48 +395,187 @@ default Long bitPos(byte @NonNull [] key, boolean bit) {
...
+		private static final SetOption IF_ABSENT_INSTANCE = new SetOption(Type.SET_IF_ABSENT, null);
+		private static final SetOption IF_PRESENT_INSTANCE = new SetOption(Type.SET_IF_PRESENT, null);
+
+		private final @NonNull Type type;
```
mp911de: Please avoid adding `@NonNull`. Instead, please annotate the class with `@NullMarked` to opt in for non-null by default.

src/main/java/org/springframework/data/redis/connection/RedisStringCommands.java
```diff
@@ -392,48 +395,187 @@ default Long bitPos(byte @NonNull [] key, boolean bit) {
...
+		 * <p>
+		 * <b>Note:</b> Empty byte arrays are valid comparison values.
+		 *
+		 * @param value the expected current value to compare against; must not be {@literal null}.
+		 * @return a new {@link SetOption} instance for the {@code IFEQ} condition.
+		 * @throws IllegalArgumentException if {@code value} is {@literal null}.
 		 */
-		public static SetOption upsert() {
-			return UPSERT;
+		public static SetOption ifValueEqual(byte @NonNull [] value) {
```
mp911de: Maybe `ifEqual(…)` to shorten the name, maybe good for consideration.

src/main/java/org/springframework/data/redis/connection/RedisStringCommands.java
```diff
@@ -392,48 +395,187 @@ default Long bitPos(byte @NonNull [] key, boolean bit) {
...
+		 * <p>
+		 * <b>Note:</b> A defensive copy of the internal byte array is returned to preserve immutability.
+		 *
+		 * @return a copy of the comparison value, or {@literal null}
+		 */
+		public byte @Nullable [] getCompareValue() {
+			if (this.compareValue == null) {
+				return null;
+			}
+			return Arrays.copyOf(this.compareValue, this.compareValue.length);
```
mp911de: We typicall do not use copied instances. Depending on our Template API design, this property might be an internal piece anyway.

Also, looking at the calling code of this method, the method should be really non-nullable and throw an `IllegalStateException` if the value is null and the condition is not one of if-equal/if-not-equal for a clearer usage.

src/main/java/org/springframework/data/redis/connection/RedisStringCommands.java
```diff
@@ -392,48 +395,187 @@ default Long bitPos(byte @NonNull [] key, boolean bit) {
...
+		@Override
+		public String toString() {
+			return "%s{type=%s, compareValue=%s}".formatted(getClass().getSimpleName(), type, compareValue != null ? "*****" : "<none>");
+		}
+
+		/**
+		 * {@code SET} command options for {@code NX}, {@code XX}, {@code IFEQ}.
+		 *
+		 * @author Yordan Tsintsov
+		 * @since 4.1.0
```
mp911de: Nit: For `.0` versions, we would just use `4.1`. Less noise.

src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
```diff
@@ -186,6 +186,104 @@ default Boolean setIfPresent(@NonNull V value, @NonNull Duration timeout) {
...
+	 * Set the bound key to hold the string {@code value}, if and only if the current value
+	 * is equal to the {@code oldValue}.
+	 *
+	 * @param newValue must not be {@literal null}.
+	 * @param compareValue must not be {@literal null}.
+	 * @return {@literal null} when used in pipeline / transaction.
+	 * @since 4.1.0
+	 * @see <a href="https://redis.io/commands/setnx">Redis Documentation: SET</a>
+	 */
+	Boolean setIfEqual(@NonNull V newValue, @NonNull V compareValue);
```
mp911de, line 242: I wonder whether we should call these `compareAndSet`

src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
```diff
@@ -186,6 +186,104 @@ default Boolean setIfPresent(@NonNull V value, @NonNull Duration timeout) {
...
+	 *
+	 * @param newValue must not be {@literal null}.
+	 * @param compareValue must not be {@literal null}.
+	 * @param timeout the key expiration timeout.
+	 * @param unit must not be {@literal null}.
+	 * @return {@literal null} when used in pipeline / transaction.
+	 * @since 4.1.0
+	 * @see <a href="https://redis.io/commands/set">Redis Documentation: SET</a>
+	 */
+	Boolean setIfEqual(@NonNull V newValue, @NonNull V compareValue, long timeout, @NonNull TimeUnit unit);
```
mp911de: Let's use `Expiration` in new API to avoid repetitive overloads simplifying the overall usage experience.

As follow-up, it would make sense to retrofit existing methods accepting any durations/TimeUnit argument variants to evolve the API into a better model.

Since we have to do some cleanup, the `Expiration` revision would have to deprecate the other methods so that we evolve towards a more concise interface.

src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
```diff
@@ -186,6 +186,104 @@ default Boolean setIfPresent(@NonNull V value, @NonNull Duration timeout) {
...
+	 * Set the bound key to hold the string {@code value}, if and only if the current value
+	 * is not equal to the {@code oldValue}.
+	 *
+	 * @param newValue must not be {@literal null}.
+	 * @param compareValue must not be {@literal null}.
+	 * @return {@literal null} when used in pipeline / transaction.
+	 * @since 4.1.0
+	 * @see <a href="https://redis.io/commands/setnx">Redis Documentation: SET</a>
+	 */
+	Boolean setIfNotEqual(@NonNull V newValue, @NonNull V compareValue);
```
mp911de, line 286: Suggestion: `compareAndSetIfNot`

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -515,7 +515,7 @@ void doLock(String name, Object contextualKey, @Nullable Object contextualValue,
 		Expiration expiration = Expiration.from(this.lockTtl.getTimeToLive(contextualKey, contextualValue));
 		byte[] cacheLockKey = createCacheLockKey(name);
 
-		while (!ObjectUtils.nullSafeEquals(commands.set(cacheLockKey, new byte[0], expiration, SetOption.SET_IF_ABSENT),
+		while (!ObjectUtils.nullSafeEquals(commands.set(cacheLockKey, new byte[0], expiration, SetOption.ifAbsent()),
```
mp911de, line 518: These changes illustrate that preserving the enum constants in the refactored class is a good idea.

src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
Dgramada: Wouldn't it be better to keep `setIfNotEqual` because it:
- is consistent with existing Spring Data Redis naming: `setIfAbsent`, `setIfPresent`
- directly maps to Redis command options: IFEQ, IFNE

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
```diff
@@ -409,43 +403,53 @@ static GetExParams toGetExParams(Expiration expiration, GetExParams params) {
...
+	 * <dt>{@link SetOption.Type#SET_IF_VALUE_NOT_EQUAL}</dt>
+	 * <dd>{@code IFNE}</dd>
 	 * </dl>
 	 *
-	 * @param option must not be {@literal null}.
-	 * @since 2.2
+	 * @param condition must not be {@literal null}.
+	 * @since 4.1
 	 */
-	public static SetParams toSetCommandNxXxArgument(SetOption option) {
```
viktoriya-kutsarova: As this is a public method, we should probably first deprecate it and just add the new one.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
```diff
@@ -409,43 +403,53 @@ static GetExParams toGetExParams(Expiration expiration, GetExParams params) {
...
+	 * <dt>{@link SetOption.Type#SET_IF_VALUE_NOT_EQUAL}</dt>
+	 * <dd>{@code IFNE}</dd>
 	 * </dl>
 	 *
-	 * @param option must not be {@literal null}.
-	 * @since 2.2
+	 * @param condition must not be {@literal null}.
+	 * @since 4.1
 	 */
-	public static SetParams toSetCommandNxXxArgument(SetOption option, SetParams params) {
```
viktoriya-kutsarova: Leave this method, just deprecate it.

src/main/java/org/springframework/data/redis/connection/lettuce/LettuceConverters.java
```diff
@@ -557,32 +559,87 @@ private static Set<Flag> parseFlags(@Nullable Set<NodeFlag> source) {
...
+
+	/**
+	 * Converts a given {@link Expiration} and {@link SetOption} to the according {@link SetArgs}
+	 * with {@link ByteBuffer} compare value.<br />
+	 *
+	 * @param expiration can be {@literal null}.
+	 * @param option can be {@literal null}.
+	 * @since 4.1
+	 */
+	public static SetArgs toReactiveSetArgs(@Nullable Expiration expiration, @Nullable SetOption option) {
```
viktoriya-kutsarova: Can you check if the `LettuceReactiveStringCommands#set` also needs to use this method?

src/main/java/org/springframework/data/redis/connection/ReactiveStringCommands.java
```diff
@@ -64,8 +65,10 @@ class SetCommand extends KeyCommand {
 		private final @Nullable Expiration expiration;
 		private final @Nullable SetOption option;
 
-		private SetCommand(@Nullable ByteBuffer key, @Nullable ByteBuffer value, @Nullable Expiration expiration,
-				@Nullable SetOption option) {
+		private SetCommand(@Nullable ByteBuffer key,
```
viktoriya-kutsarova: nit: No need to put each parameter on a separate line.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
viktoriya-kutsarova: Please deprecate this method instead of deleting it to maintain backward compatibility.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
viktoriya-kutsarova: Please deprecate this method instead of deleting it to maintain backward compatibility.

src/test/java/org/springframework/data/redis/connection/AbstractConnectionIntegrationTests.java
```diff
@@ -3234,6 +3231,310 @@ void setWithoutExpirationAndPresentOptionShouldSetTtlWhenKeyDoesNotExist() {
...
+		actual.add(connection.get(key));
+
+		List<Object> result = getResults();
+		assertThat(result.get(0)).isEqualTo(Boolean.TRUE);
+		assertThat(result.get(1)).isEqualTo(Boolean.FALSE);
+		assertThat(result.get(2)).isEqualTo(Boolean.TRUE);
+		assertThat(result.get(3)).isEqualTo("foo");
+	}
+
+	@Test // GH-3254
```
viktoriya-kutsarova: Remove the unrelated GH-3254 reference.

src/test/java/org/springframework/data/redis/connection/AbstractConnectionIntegrationTests.java
```diff
@@ -3234,6 +3231,310 @@ void setWithoutExpirationAndPresentOptionShouldSetTtlWhenKeyDoesNotExist() {
...
+		actual.add(connection.get(key));
+
+		List<Object> result = getResults();
+		assertThat(result.get(0)).isEqualTo(Boolean.TRUE);
+		assertThat(result.get(1)).isEqualTo(Boolean.TRUE);
+		assertThat(result.get(2)).isEqualTo(Boolean.TRUE);
+		assertThat(result.get(3)).isEqualTo("bar-bar");
+	}
+
+	@Test // GH-3254
```
viktoriya-kutsarova: nit: Remove the unrelated GH-3254 reference.

src/main/java/org/springframework/data/redis/connection/RedisStringCommands.java
```diff
@@ -392,48 +395,184 @@ default Long bitPos(byte @NonNull [] key, boolean bit) {
...
+		}
+
+		@Override
+		public int hashCode() {
+			return Objects.hash(type, Arrays.hashCode(compareValue));
+		}
+
+		@Override
+		public String toString() {
+			return "%s{type=%s, compareValue=%s}".formatted(getClass().getSimpleName(), type, compareValue != null ? "*****" : "<none>");
```
viktoriya-kutsarova: Is compareValue sensitive here? If not, the ***** masking seems unnecessary - consider removing it from the output.

src/main/java/org/springframework/data/redis/connection/ReactiveStringCommands.java
viktoriya-kutsarova: nit: The parameters can stay on two lines.

viktoriya-kutsarova (commented):
To complete the Redis version update, also update Jenkinsfile and pipeline.properties, and move the Dockerfile to ci/openjdk25-redis-8.4/

mp911de:
Postponing to M2 (March milestone)

src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
Dgramada: Wouldn't it be better to keep `setIfNotEqual` because it:

- is consistent with existing Spring Data Redis naming: setIfAbsent, setIfPresent
- directly maps to Redis command options: IFEQ, IFNE

Dgramada:
I created a new clean PR for the CAS operations #3324

mp911de:
It has been a good idea to create a new PR given the comment history here. Closing in favor of #3324

PR #3305, opened 2026-02-03
Add Expiration-based API for SET command operations.

Dgramada:
## Summary

This PR introduces `Expiration`-based API methods for SET command operations, replacing the existing `TimeUnit` variants.

## Motivation

The current API using `TimeUnit` cannot express advanced Redis SET options:
- **KEEPTTL** - Preserve existing TTL when updating a value
- **Persistent** - Explicitly no expiration
- **Unix Timestamp** - Set expiration to absolute time (EXAT/PXAT)

The `Expiration` class already supports these features at the connection layer, but they were not accessible through the template API.

### New Methods (with `@since 4.1`)
- `ValueOperations`: `set`, `setGet`, `setIfAbsent`, `setIfPresent` with `Expiration`
- `BoundValueOperations`: `set`, `setGet`, `setIfAbsent`, `setIfPresent` with `Expiration`
- `ReactiveValueOperations`: `set`, `setGet`, `setIfAbsent`, `setIfPresent` with `Expiration`

### Deprecated Methods
- All `TimeUnit` variants marked with `@Deprecated(since = "4.1", forRemoval = true)`

### Bug Fix
- Fixed `LettuceReactiveStringCommands.setGet()` which was ignoring expiration and set options (pre-existing bug since 3.5)

### Kotlin Extensions
- Added coroutine extension functions for reactive `Expiration` variants

## Testing
- Added integration tests for all new methods
- Added unit tests for Kotlin extensions

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveStringCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveStringCommands.java
@@ -52,6 +52,7 @@ import org.springframework.util.Assert;
  * @author John Blum
  * @author Marcin Grzejszczak
  * @author Viktoriya Kutsarova
+ * @author Yordan Tsintsov
  * @since 2.0
  */
 class LettuceReactiveStringCommands implements ReactiveStringCommands {
--- a/src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
@@ -31,6 +31,7 @@ import org.springframework.util.Assert;
  * @author Jiahe Cai
  * @author Christoph Strobl
  * @author Marcin Grzejszczak
+ * @author Yordan Tsintsov
  */
 @NullUnmarked
 public interface BoundValueOperations<K, V> extends BoundKeyOperations<K> {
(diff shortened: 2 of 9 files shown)


src/main/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveStringCommands.java
```diff
@@ -90,15 +91,7 @@ public Flux<BooleanResponse<SetCommand>> set(Publisher<SetCommand> commands) {
 			Assert.notNull(command.getKey(), "Key must not be null");
 			Assert.notNull(command.getValue(), "Value must not be null");
 
-			SetArgs args = null;
```
onobc, line 93: > Fixed LettuceReactiveStringCommands.setGet() which was ignoring expiration and set options (pre-existing bug since 3.5)

Is there a bug to associate this with? If there is a bug tracking this let's put it in a separate commit (in this PR is fine) so we can attribute the commit to the issue.

Otherwise it is fine in this commit.

Dgramada: I could not find an issue concerning the `setGet()` bug. I found it accidentally while writing tests for `setGet()` and would assume that it was not know that the method would not enforce expirations.

onobc:
Thank for another excellent contribution @Dgramada !

PR #3309, opened 2026-02-10
Consider DefaultTyping configuration in Jackson Serializers

### LINKED ISSUE #3306: GenericJacksonJsonRedisSerializer ignores DefaultTyping.NON_FINAL_AND_ENUMS for enum types
When using `GenericJacksonJsonRedisSerializer`, enum values are always serialized without type metadata, even when `DefaultTyping.NON_FINAL_AND_ENUMS` is explicitly configured. This causes `ClassCastException` during deserialization.

### Example
```java
public enum Status {
    ACTIVE, INACTIVE
}

@Cacheable(cacheNames = "status", cacheManager = "redisCacheManager")
public Status getStatus(String id) {
    return Status.ACTIVE;
}
```

**Serialized value in Redis:**
```json
"ACTIVE"
```

**Exception on deserialization:**
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class Status
```

### Configuration Attempt
```java
GenericJacksonJsonRedisSerializer.create(b -> {
    b.enableDefaultTyping(ptv);
    b.customize(mb -> 
        mb.activateDefaultTypingAsProperty(
            ptv,
            DefaultTyping.NON_FINAL_AND_ENUMS,
            "@class"
        )
    );
});
```

Using `customize()` to override with Jackson's default `DefaultTypeResolverBuilder` does enable enum type metadata, but it **loses Spring's Kotlin-specific handling** for final classes:
```java
// This Kotlin support in Spring's TypeResolverBuilder is lost:
if (javaType.isFinal() && !KotlinDetector.isKotlinType(javaType.getRawClass())
        && javaType.getRawClass().getPackageName().startsWith("java")) {
    return false;
}
```

This forces users to choose between enum support and Kotlin data class support.

### Root Cause

Spring's `TypeResolverBuilder` unconditionally excludes enum types regardless of the `DefaultTyping` setting:

https://github.com/spring-projects/spring-data-redis/blob/main/src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java#L631-L633
```java
if (javaType.isEnumType() || ClassUtils.isPrimitiveOrWrapper(javaType.getRawClass())) {
    return false;  // Always excluded
}
```

### Suggested Fix

Store the `DefaultTyping` parameter and use it in `useForType()`:
```java
private static class TypeResolverBuilder extends DefaultTypeResolverBuilder {

    private final DefaultTyping defaultTyping;

    public TypeResolverBuilder(PolymorphicTypeValidator subtypeValidator, DefaultTyping defaultTyping, 
            JsonTypeInfo.As includeAs, JsonTypeInfo.Id idType, @Nullable String propertyName) {
        super(subtypeValidator, defaultTyping, includeAs, idType, propertyName);
        this.defaultTyping = defaultTyping;
    }

    @Override
    public boolean useForType(JavaType javaType) {
        if (javaType.isJavaLangObject()) {
            return true;
        }

        javaType = resolveArrayOrWrapper(javaType);

        // Respect DefaultTyping.NON_FINAL_AND_ENUMS for enum types
        if (javaType.isEnumType()) {
            return defaultTyping == DefaultTyping.NON_FINAL_AND_ENUMS;
        }

        if (ClassUtils.isPrimitiveOrWrapper(javaType.getRawClass())) {
            return false;
        }

        if (javaType.isFinal() && !KotlinDetector.isKotlinType(javaType.getRawClass())
                && javaType.getRawClass().getPackageName().startsWith("java")) {
            return false;
        }

        return !TreeNode.class.isAssignableFrom(javaType.getRawClass());
    }

    // ... resolveArrayOrWrapper() unchanged
}
```

This preserves backward compatibility (enums excluded by default with `NON_FINAL`) while allowing enum type metadata when explicitly requested via `NON_FINAL_AND_ENUMS`.

This preserves backward compatibility while allowing enum type metadata when explicitly requested.

### Current Workarounds

1. Wrap enums in a generic container class
2. Copy and modify Spring's `TypeResolverBuilder` (not maintainable)

### Environment

- Spring Data Redis: 4.0.x
- Jackson: 3.x

mp911de:
We're exploring whether it makes sense to accept a `DefaultTyping` config option for Jackson serializer configuration.

Likely, accepting `DefaultTyping` causes several downstream effects that we actually do not like (e.g. removing typehints for final types such as records likely breaking their support).


src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
```diff
@@ -479,7 +478,7 @@ public static class GenericJackson2JsonRedisSerializerBuilder {
 
 		private @Nullable Boolean defaultTypingEnabled;
 
-		private @Nullable DefaultTyping defaultTyping;
+		private @Nullable Predicate<JavaType> defaultTyping;
```
mp911de: That's a direction I ended up with, too. A single `Predicate` is likely not going to be sufficient. The code to determine whether to apply default typing contains opinions of ours. We would want to achieve an approach that allows folks to override our opinion or append theirs (that is, take some of our opinions but do something different for a type that we handle or apply/avoid typing for types we don't handle).

What if we introduced a `DefaultTypingPredicate` interface returning something like `YES`, `NO`, `I_DONT_HAVE_AN_OPINION` allowing functional composition?


```java
DefaultTyping.defaults()
   .and(DefaultTyping.apply((Class<?> it) -> it.isEnum())
   .and(DefaultTyping.skip((Class<?> it) -> it.isPrimitive())
   .and((Class<?> it) -> it.isRecord() ? DefaultTyping.apply() : DefaultTyping.continue());
```

or 

```java
DefaultTyping.defaults()
   .and((Class<?> it) -> it.isRecord() ? DefaultTyping.include() : DefaultTyping.passThrough())
   .and(DefaultTyping.include((Class<?> it) -> it.isEnum()))
//    .andInclude ((Class<?> it) -> it.isEnum())
   .and(DefaultTyping.exclude((Class<?> it) -> it.isPrimitive()))
//   .andExclude((Class<?> it) -> it.isPrimitive())
   .and(MyApplicationSpecificTyping.INSTANCE);
```

We'd need an idea on naming, though.

src/main/java/org/springframework/data/redis/serializer/DefaultTypingPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.util.ClassUtils;
+
+/**
+ * @author Chris Bono
+ * @since 4.1
+ */
+@FunctionalInterface
+public interface DefaultTypingPredicate {
+
+	enum Action {
```
mp911de: How about `ConditionOutcome`? Or just `Outcome`, like in Spring Boot conditions?

src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
```diff
@@ -322,46 +320,46 @@ public GenericJacksonJsonRedisSerializerBuilder<B> enableSpringCacheNullValueSup
 		 * @see <a href=
 		 *      "https://owasp.org/www-community/vulnerabilities/Deserialization_of_untrusted_data">https://owasp.org/www-community/vulnerabilities/Deserialization_of_untrusted_data</a>
 		 */
-		@Contract("-> this")
+		@Contract("_ -> this")
```
mp911de: There's no argument given to this method.

PR #3314, opened 2026-02-17
Reset Caches by flushing DB

christophstrobl:


src/main/java/org/springframework/data/redis/cache/CacheWriterOperation.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package org.springframework.data.redis.cache;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * @author Christoph Strobl
+ */
+interface CacheWriterOperation<T> {
```
onobc: Can be a `@FunctionalInterface`.

src/main/java/org/springframework/data/redis/cache/ResetCachesStrategies.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+package org.springframework.data.redis.cache;
+
+import org.springframework.data.redis.connection.RedisServerCommands.FlushOption;
+
+/**
+ * @author Christoph Strobl
+ */
+public abstract class ResetCachesStrategies {
```
onobc: I think this may read better as `CacheResetStrategy/ies` - wdyt?

Uggh, naming... always hard.

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -535,10 +537,17 @@ Long doUnlock(String name, RedisConnection connection) {
 		return connection.keyCommands().del(createCacheLockKey(name));
 	}
 
-	private <T> T execute(String name, Function<RedisConnection, T> callback) {
+	@Override
+	public <T> T execute(Function<RedisConnection, T> callback) {
```
onobc, line 541: I am sure this will clear up w/ javadocs once we get firm on the direction, but it is currently not clear that the newly added execute method does not do the potential wait for unlocked (i.e. `checkAndPotentiallyWaitUntilUnlocked`).

src/main/java/org/springframework/data/redis/cache/DefaultRedisCacheWriter.java
```diff
@@ -535,10 +537,17 @@ Long doUnlock(String name, RedisConnection connection) {
...
+		return execute(null, callback);
+	}
+
+	private <T> T execute(@Nullable String name, Function<RedisConnection, T> callback) {
```
onobc, line 545: We could just call `executeLockFree` instead of adding the `@Nullable` here.

src/main/java/org/springframework/data/redis/cache/RedisCacheManager.java
```diff
@@ -312,6 +331,16 @@ public boolean isAllowRuntimeCacheCreation() {
 		return this.allowRuntimeCacheCreation;
 	}
 
+	@Override
+	public void resetCaches() {
+
+		if(resetCachesStrategy instanceof CacheWriterOperation<?> operation) {
```
onobc: My first thought was to abstract the impl down into the strategy via a `resetCaches` method on the strategy where the strategy could take in a cache writer and a cache manager. The flushing variant could use the cache writer and the default variant could use the CM resetCaches impl. However, the latter would call back into `CacheManager.resetCaches` and that would infinite recurse (i.e. not get to `super.resetCaches`). 

Just putting my thoughts -  I am not sure which direction to go but I do like the thought of the strategy handling the work.

src/main/java/org/springframework/data/redis/cache/RedisCacheManager.java
onobc, line 1: I am sure I am missing something, but it seems that we are overlapping w/ current cache cleaning impl in BatchStrategy (`keys` and `scan`). Seems like we could just add another strategy for flushAsync().

src/main/java/org/springframework/data/redis/cache/ResetCachesStrategies.java
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	enum FlushingResetStrategy implements ResetCachesStrategy, CacheWriterOperation<String> {
+
+		INSTANCE;
+
+		@Override
+		public String doWithCacheWriter(RedisCacheWriter cacheWriter) {
+			return cacheWriter.execute(connection -> {
+				connection.serverCommands().flushDb(FlushOption.ASYNC);
```
onobc: Is there any other cleanup / stats we are missing that the `super.resetCaches` may be doing that we should do here as well?

onobc (commented):
Thanks @christophstrobl . I have added some comments / questions.

src/main/java/org/springframework/data/redis/cache/ResetCachesStrategies.java
mp911de: How about just `ResetStrategy`? `Cache` defines `clear` and `evict` methods and `reset` is distinct from these (thus it ranges semantically in a similar space).

Design-wise, it could be an interface with two static methods, `sequential()` (the default) and `flushDb()`. I like the `CacheWriterOperation` approach as it repeats how we approach functional composition in other parts of the framework.

PR #3318, opened 2026-02-23
Added support for Redis 8.4 DELEX command

Dgramada:
This change introduces the [DELEX](https://redis.io/docs/latest/commands/delex/) string command feature as part of the features provided by the spring-data-redis project.

This feature is available starting from [Redis OSS version 8.4](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.4-release-notes/) and later

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/ValueOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/ValueOperations.java
@@ -19,12 +19,14 @@ import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 import org.springframework.data.redis.connection.BitFieldSubCommands;
+import org.springframework.data.redis.connection.RedisStringCommands.DeleteOption;
 import org.springframework.data.redis.core.types.Expiration;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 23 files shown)


src/test/java/org/springframework/data/redis/connection/jedis/JedisClusterConnectionTests.java
```diff
@@ -3530,4 +3532,73 @@ void shouldUseCachedTopology() {
...
+	void delexWithValueEqualOptionShouldNotDeleteKeyWhenKeyDoesNotExist() {
+
+		Boolean result = clusterConnection.delex(KEY_1_BYTES, DeleteOption.ifEqual(), VALUE_1_BYTES);
+
+		assertThat(result).isFalse();
+		assertThat(nativeConnection.exists(KEY_1)).isFalse();
+	}
+
+	@Test
+	@EnabledOnRedisVersion("8.4")
```
mp911de: We prefer `@EnabledOnCommand("DELEX")` gurards

src/main/java/org/springframework/data/redis/core/ValueOperations.java
```diff
@@ -349,6 +349,28 @@ default Boolean setIfPresent(@NonNull K key, @NonNull V value, @NonNull Duration
...
+	/**
+	 * Delete the key if the value is not equal to the current value.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param value must not be {@literal null}.
+	 * @return {@literal null} when used in pipeline / transaction.
+	 * @see <a href="https://redis.io/commands/delex">Redis Documentation: DELEX</a>
+	 * @since 4.2
+	 */
+	@NonNull Boolean deleteIfNotEqual(@NonNull K key, @NonNull V value);
```
mp911de: `@NonNull Boolean` isn't right here as the return value can be `null` when used in transactions. Just `Boolean` is fine.

Dgramada:
I should mention that CAS/CAD operations in Redis also support two more options - **IF_DIGEST_EQUAL** and **IF_DIGEST_NOT_EQUAL**, which could be added at some point as well.

Maybe we should take this into consideration, while naming the current template CAS/CAD methods.

We could then name the methods something like:

- `compareAndDelete` (instead of `deleteIfEqual`)
- `compareAndDeleteIfDifferent` (instead of `deleteIfNotEqual`)
- `compareDigestAndDelete` (potential name if digest logic is added at some point in the future)
- `compareDigestAndDeleteIfDifferent` (potential name if digest logic is added at some point in the future)

mp911de:
We want to avoid uncontrolled growth in method (name) overload variants. Our current thinking is at `compareAndDelete(K key, V oldValue, V newValue)` and `compareAndDelete(K key, CasOperator<V> oldValue, V newValue)` where `CasOperator` (or a better name) can be `CasOperator.ifEqual(…)`, `CasOperator.ifNotEqual(…)` and so on.

mp911de:
I polished the change and pushed it onto a branch https://github.com/spring-projects/spring-data-redis/tree/delex

mp911de:

PR #3321, opened 2026-03-01
Add `RedisListenerEndpointRegistrar` and content negotiation

Ilyass-Bougati:
Hey @mp911de,

As promised in #3303, here is the follow-up PR tackling the next phase of the `@RedisListener` infrastructure.

This PR introduces the `RedisListenerEndpointRegistrar` to handle programmatic registration and extended configurability. 

To address the content negotiation approach we discussed, I've added the `consumes` attribute to `@RedisListener` and wired up a custom `ContentTypeResolver`. It hooks into `AbstractMessageConverter` to drive deserialization explicitly from the annotation, falling back to smart classpath defaults (and `byte[]` converters) when a specific type isn't provided.

Let me know if the way the converters and resolvers are registered here aligns with what you had in mind!

## Relates to
* #1004
* Follow-up to #3303

### DIFF
--- a/src/main/antora/modules/ROOT/pages/redis/pubsub-annotated.adoc
+++ b/src/main/antora/modules/ROOT/pages/redis/pubsub-annotated.adoc
@@ -43,7 +43,7 @@ public class RedisConfiguration {
     RedisMessageListenerContainer factory = new RedisMessageListenerContainer();
     factory.setConnectionFactory(connectionFactory);
     return factory;
-	}
+  }
 }
 ----
 
--- a/src/main/java/org/springframework/data/redis/annotation/RedisListener.java
+++ b/src/main/java/org/springframework/data/redis/annotation/RedisListener.java
@@ -99,9 +99,9 @@ public @interface RedisListener {
 	String topic() default "";
 
 	/**
-	 * Configure the primary mapping by media type that is consumed by the handler method. Setting {@code consumes}
-	 * defines the message content type as Redis Pub/Sub messages do not feature headers to indicate a content type
-	 * towards message converter selection. Examples:
+	 * Configure the primary mapping by MIME type that is consumed by the handler method. Setting {@code consumes} defines
+	 * the message content type as Redis Pub/Sub messages do not feature headers to indicate a content type towards
+	 * message converter selection. Examples:
 	 *
 	 * <pre class="code">
 	 * consumes = "text/plain"


src/main/java/org/springframework/data/redis/annotation/RedisContentTypeResolver.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jspecify.annotations.Nullable;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.converter.ContentTypeResolver;
+import org.springframework.util.InvalidMimeTypeException;
+import org.springframework.util.MimeType;
+
+/**
+ * @author Ilyass Bougati
+ */
+public class RedisContentTypeResolver implements ContentTypeResolver {
```
mp911de: I had improper expectations towards the `ContentTypeResolver` interface. I somehow assumes access to the message object but we only have access to `MessageHeaders`.

With the lack of further context, it would make sense to use `DefaultContentTypeResolver` instead with the idea of configuring a default `MimeType` via a `RedisListenerEndpointRegistrar` setting. 

Additionally, we need to propagate `MessageHeaders.CONTENT_TYPE` from `@RedisListener` to the message headers from within `HandlerMethodMessageListenerAdapter`.

src/main/java/org/springframework/data/redis/annotation/RedisListenerConfigurer.java
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.redis.annotation;
+
+import org.springframework.data.redis.config.RedisListenerEndpointRegistrar;
+
+@FunctionalInterface
+public interface RedisListenerConfigurer {
```
mp911de: I suggest removing `@FunctionalInterface` and evolve this configurer into a `WebMvcConfigurer`-style configurer with the follwing methods:

```
	default void addConverters(ConverterRegistry registry) {
	}

	default void addArgumentResolvers(List<HandlerMethodArgumentResolver> resolvers) {
	}

	default @Nullable Validator getValidator() {
		return null;
	}

	default void configureMessageConverters(RedisMessageConverters.Builder builder) {
	}
```

Over time, we will have multiple aspects to be configured, looking forward to message sending, we can have `HandlerMethodReturnValueHandler`, `HandlerExceptionResolver` and others.

`RedisMessageConverters.Builder` should attempt initializing converters similar to the HTTP counterpart `HttpMessageConverters.ServerBuilder`. 

I would want to start the `RedisMessageConverters.Builder` interface with `registerDefaults(boolean)`, `withStringConverter(MessageConverter stringMessageConverter);`, `withStringConverter(CharSet charset);` and `addCustomConverter(…)`. If the builder is left unconfigured, then we'd default to an UTF-8 converter.

Zooming out, we can configure all these in some way on `DefaultMessageHandlerMethodFactory` but this is inconvenient.

src/main/java/org/springframework/data/redis/annotation/RedisListenerConfigurer.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Callback allowing a {@link org.springframework.data.redis.config.RedisListenerEndpointRegistry}
+     * and specific {@link org.springframework.data.redis.config.RedisListenerEndpoint} instances to be registered
+     * against the given {@link org.springframework.data.redis.config.RedisListenerEndpointRegistrar}.
+     *
+     * @param registrar the registrar to be configured
+     */
+    void configureRedisListeners(RedisListenerEndpointRegistrar registrar);
```
mp911de: Rename to `configureRegistrar(…)` and let's make it a default method. This method serves as escape-hatch and it would be called after us applying all of our configuration processing. The escape-hatch serves as generic hook for configuring aspects that cannot be configured through methods we expose through the interface above.

Ilyass-Bougati:
Thanks for the feedback, @mp911de ! I've refactored the configuration as suggested:
- `RedisListenerConfigurer` is no longer a functional interface and now includes default hooks for converters, argument resolvers, and validators.
- Renamed the original callback to `configureRegistrar` as an 'escape hatch'.
- Introduced `RedisMessageConverters` with a fluent Builder to manage default and custom converters.
- Updated the `RedisListenerAnnotationBeanPostProcessor` and `Registrar` to support these new customization points.

src/main/java/org/springframework/data/redis/annotation/RedisContentTypeResolver.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.converter.ContentTypeResolver;
+import org.springframework.messaging.converter.DefaultContentTypeResolver;
+import org.springframework.util.InvalidMimeTypeException;
+import org.springframework.util.MimeType;
+
+/**
+ * @author Ilyass Bougati
+ */
+public class RedisContentTypeResolver extends DefaultContentTypeResolver {
```
mp911de: This class can be removed, it is doing what its superclass is doing.

src/main/java/org/springframework/data/redis/annotation/RedisMessageConverters.java
```diff
@@ -0,0 +1,95 @@
...
+    private RedisMessageConverters() {
+    }
+
+    public interface Builder {
+
+        Builder registerDefaults(boolean registerDefaults);
+
+        Builder withStringConverter(MessageConverter stringMessageConverter);
+
+        Builder withStringConverter(Charset charset);
```
mp911de: Good candidate for 

```
        default Builder withStringConverter(Charset charset){
            return withStringConverter(new StringMessageConverter(charset));
        }
```

src/main/java/org/springframework/data/redis/annotation/RedisMessageConverters.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.converter.StringMessageConverter;
+import org.springframework.messaging.converter.CompositeMessageConverter;
+
+/**
+ * Represents a builder for {@link MessageConverter}s to be used with Redis listeners.
+ *
+ * @author Ilyass Bougati
+ */
+public final class RedisMessageConverters {
```
mp911de: Take a look at `DefaultHttpMessageConverters.detectMessageConverters` to provide a detection mechanism for default converters.

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistrar.java
```diff
@@ -0,0 +1,196 @@
...
+
+        if (this.customArgumentResolvers != null && !this.customArgumentResolvers.isEmpty()) {
+            defaultFactory.setCustomArgumentResolvers(this.customArgumentResolvers);
+        }
+
+        defaultFactory.afterPropertiesSet();
+        return defaultFactory;
+    }
+
+    private MessageConverter buildSmartDefaultMessageConverter() {
```
mp911de: MessageConverters are build through `RedisMessageConverters` and so the Jackson part isn't required.

src/main/java/org/springframework/data/redis/annotation/RedisListenerAnnotationBeanPostProcessor.java
```diff
@@ -191,49 +143,23 @@ public Object postProcessAfterInitialization(Object bean, String beanName) {
...
 	 */
 	protected void processRedisListener(RedisListener redisListener, Method method, Object bean) {
-
 		MethodRedisListenerEndpoint endpoint = createEndpoint(redisListener, method, bean);
-
-		RedisMessageListenerContainer container = null;
-		String containerName = resolve(redisListener.container());
-		Assert.state(this.beanFactory != null, "BeanFactory must be set to obtain container container by bean name");
-
-		if (StringUtils.hasText(containerName)) {
```
mp911de, line 201: This removal doesn't look right as its been replaced with a default container later on.

src/main/java/org/springframework/data/redis/config/MethodRedisListenerEndpoint.java
```diff
@@ -161,8 +167,13 @@ public HandlerMethodMessageListenerAdapter createListener() {
 		InvocableHandlerMethod invocableHandlerMethod = this.messageHandlerMethodFactory
 				.createInvocableHandlerMethod(this.bean, this.method);
 
-		// Endpoint is now aware of its listener
-		return new HandlerMethodMessageListenerAdapter(invocableHandlerMethod);
+		HandlerMethodMessageListenerAdapter listener = new HandlerMethodMessageListenerAdapter(invocableHandlerMethod);
+
+		// Pass the properties down from the endpoint to the adapter
+        listener.setConsumes(this.consumes);
```
mp911de: this should be a constructor argument to HandlerMethodMessageListenerAdapter.

src/main/java/org/springframework/data/redis/annotation/RedisListenerAnnotationBeanPostProcessor.java
```diff
@@ -242,55 +168,56 @@ private String getEndpointId(RedisListener redisListener) {
...
 
-		private MessageHandlerMethodFactory createDefaultRedisHandlerMethodFactory() {
+		if (this.beanFactory instanceof ListableBeanFactory lbf) {
+			Map<String, RedisListenerConfigurer> configurers = lbf.getBeansOfType(RedisListenerConfigurer.class);
 
-			DefaultFormattingConversionService conversionService = new DefaultFormattingConversionService();
-			conversionService.addConverter(ByteArrayToStringConverter.INSTANCE);
-			DefaultMessageHandlerMethodFactory defaultFactory = new DefaultMessageHandlerMethodFactory();
-			defaultFactory.setConversionService(conversionService);
+			if (!configurers.isEmpty()) {
```
mp911de: This is duplicated by the code in `RedisListenerBootstrapConfiguration.redisListenerEndpointRegistrar`

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistrar.java
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.redis.config;
+
+import jakarta.annotation.Nullable;
```
mp911de: Nit: Use JSpecify.

src/main/java/org/springframework/data/redis/annotation/RedisListenerAnnotationBeanPostProcessor.java
```diff
@@ -68,14 +59,13 @@
...
- * @see RedisListenerEndpointRegistry
  */
 public class RedisListenerAnnotationBeanPostProcessor
-		implements BeanPostProcessor, InitializingBean, BeanFactoryAware, Ordered, SmartInitializingSingleton {
+		implements BeanPostProcessor, BeanFactoryAware, Ordered, SmartInitializingSingleton {
 
 	protected final Log logger = LogFactory.getLog(getClass());
 
-	private final MessageHandlerMethodFactoryAdapter messageHandlerMethodFactory = new MessageHandlerMethodFactoryAdapter();
+	private final RedisListenerEndpointRegistrar registrar = new RedisListenerEndpointRegistrar();
```
mp911de, line 80: There's a conceptual conflict withd defining `RedisListenerEndpointRegistrar` as bean and here as final local instance.

I suggest dropping the bean idea and instead, keeping this instance here around.

src/main/java/org/springframework/data/redis/annotation/RedisListenerConfigurer.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.core.convert.converter.ConverterRegistry;
+import org.springframework.data.redis.config.RedisListenerEndpointRegistrar;
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
+import org.springframework.validation.Validator;
+
+public interface RedisListenerConfigurer {
```
mp911de: This is a candidate for the `config` package.

src/main/java/org/springframework/data/redis/annotation/RedisMessageConverters.java
```diff
@@ -0,0 +1,95 @@
+package org.springframework.data.redis.annotation;
```
mp911de: This is a candidate for the `config` package.

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistrar.java
```diff
@@ -0,0 +1,196 @@
...
+            this.beanFactory = configurableBeanFactory;
+        }
+    }
+
+    @Override
+    public void afterPropertiesSet() {
+        resolveRegistry();
+
+        if (this.methodFactory == null) {
+            this.methodFactory = createDefaultMessageHandlerMethodFactory();
```
mp911de: This is too late, instead, let's introduce a getter for `MessageHandlerMethodFactory` that initializes the `MessageHandlerMethodFactory` based on the configuration that has been applied so far.

mp911de (changes requested):
Did another pass and left a few comments. Please mind using tab indents instead of space indents.

Ilyass-Bougati:
Thanks for the feedback! I've pushed a new commit trying to address all of your latest review comments. Let me know how it looks or if anything else needs adjusting.

src/main/java/org/springframework/data/redis/config/RedisMessageConverters.java
```diff
@@ -0,0 +1,120 @@
...
+	private RedisMessageConverters() {}
+
+	static List<MessageConverter> detectMessageConverters() {
+		List<MessageConverter> converters = new ArrayList<>();
+
+		if (jackson2Present) {
+			converters.add(new JacksonJsonMessageConverter());
+		}
+
+		converters.add(new StringMessageConverter(StandardCharsets.UTF_8));
```
mp911de: Registering `StringMessageConverter` conflicts with the configuration down in the builder when a `stringMessageConverter` is configured. The code in this class is a bit of a mess with multiple times creating `StringMessageConverter`.

Auto-detection should register `GsonMessageConverter`, `JsonbMessageConverter`, and `KotlinSerializationJsonMessageConverter` if their respective trigger types are present. Refer to `AbstractMessageBrokerConfiguration`.

src/main/java/org/springframework/data/redis/listener/adapter/HandlerMethodMessageListenerAdapter.java
```diff
@@ -40,12 +41,18 @@ public class HandlerMethodMessageListenerAdapter implements MessageListener {
 	private final Log logger = LogFactory.getLog(getClass());
 
 	private final InvocableHandlerMethod handlerMethod;
+	private String consumes;
```
mp911de: Let's make this one `final`.

src/test/java/org/springframework/data/redis/annotation/RedisListenerIntegrationTests.java
```diff
@@ -91,7 +90,7 @@ void shouldListenForMessage() throws InterruptedException {
 		MyListener bean = context.getBean(MyListener.class);
 		bean.message.clear();
 
-		template.convertAndSend("my-channel-listener", "Hello Redis!");
+		template.convertAndSend("my-channel-listener", "\"Hello Redis!\"");
```
mp911de: That's not how it is supposed to work. A string message must remain a string message.

src/main/java/org/springframework/data/redis/config/RedisListenerEndpointRegistrar.java
```diff
@@ -0,0 +1,166 @@
...
+	@Override
+	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
+		if (beanFactory instanceof ConfigurableBeanFactory configurableBeanFactory) {
+			this.beanFactory = configurableBeanFactory;
+		}
+	}
+
+	@Override
+	public void afterPropertiesSet() {
+		resolveRegistry();
```
mp911de: The class should not resolve a registry. Instead, it should only verify if a registry has been configured.

mp911de:
I applied some polishing and documented how to use the configurer along with a series of other aspects. I'll keep this PR open for team-review before merging it. The final step (after merge) would be to add reply and exception handling behavior.

Ilyass-Bougati:
Thank you @mp911de!

Regarding the reply and exception handling behavior, do you have a preferred approach or a specific Spring messaging pattern (like `@SendTo` or specific error handler interfaces) you'd like me to mirror for that? Let me know your thoughts, and I can start looking into the implementation once this is merged.

mp911de:
Thank you for your massive contribution. It adds a long-envisioned feature and we're pretty much on track to get full support. That's merged and polished now.

mp911de:
@Ilyass-Bougati would you like to contribute the missing bits for `@SendTo` responses and annotation-driven exception handling methods (`@MessageExceptionHandler`)?

Ilyass-Bougati:
Hi @mp911de, thank you for reaching out! I was away for the end of Ramadan and the Eid holiday, but I am back and would absolutely love to contribute the missing bits for `@SendTo` and `@MessageExceptionHandler`.

I'm going to start looking into how these are implemented in other Spring messaging modules to make sure I align with the framework's established patterns. Are there any specific classes or modules (like `spring-jms` or `spring-kafka`) that you'd recommend I use as the primary blueprint for the Redis implementation?

mp911de:
Eid Mubarak. I suggest to have a look on `SendToMethodReturnValueHandler` and `AbstractMethodMessageHandler`. Both are good entrypoints for return value handling. `AbstractMethodMessageHandler` contains also handling for exceptions. `SimpAnnotationMethodMessageHandler` contains to some extent what `HandlerMethodMessageListenerAdapter` is doing. I think that we will have to update our adapter to not invoke directly the handler method but rather dispatch the message to `AbstractMethodMessageHandler.handleMessage(…)`. I'm not sure whether we need the mapping information `<T>` (as in `AbstractMethodMessageHandler<T>`) but that's something to find out along the way.

PR #3327, opened 2026-03-11
Draft API for connection and template layers for Redis JSON commands

Dgramada:
Provides a draft version of how the JSON template and connection layer API should look like. The purpose of this PR is to receive feedback and to get a better feeling of what should Spring Data Redis provide for users that would like to use Redis JSON.


src/main/java/org/springframework/data/redis/core/json/JsonArrayRange.java
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Value object representing a range of array elements.
+ *
+ * @author Yordan Tsintsov
+ * @since 4.3
+ * @see <a href="https://redis.io/docs/latest/commands/json.arrindex/">Redis JSON.ARRINDEX</a>
+ */
+@NullMarked
+public final class JsonArrayRange {
```
onobc: We do have a Spring Data Range object `org.springframework.data.domain.Range<T>`. I wonder if we could that instead?

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,489 @@
...
+	 * Append the JSON values into the array at path after the last element in it.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param path must not be {@literal null}.
+	 * @param values must not be {@literal null}. {@literal null} values should be represented as JSON "null" values.
+	 * @return a list where each element contains the number of elements added to the array or {@literal null} if path does not exist.
+	 * @see <a href="https://redis.io/docs/latest/commands/json.arrappend/">Redis Documentation: JSON.ARRAPPEND</a>
+	 * @since 4.3
+	 */
+	List<@Nullable Long> jsonArrAppend(byte[] key, String path, String... values);
```
onobc: I notice the commands use `string` but the operations use `JsonPath`. Even if JsonPath is just a wrapper around a string it may insulate us and give us a bit more flexibility for later extension. Wdyt?

src/main/java/org/springframework/data/redis/core/json/JsonPath.java
```diff
@@ -0,0 +1,311 @@
...
+ * JsonPath.root().child("items")
+ *     .filter(FilterExpression.of("@.price &lt; 10")); // $.items[?(@.price &lt; 10)]
+ * </pre>
+ *
+ * @author Yordan Tsintsov
+ * @see <a href="https://redis.io/docs/latest/develop/data-types/json/path/">Redis JSONPath Documentation</a>
+ * @since 4.3
+ */
+@NullMarked
+public final class JsonPath {
```
onobc: I understand why you went w/ this DSL - to help build the JSON path. However, I think that providing the full JSON path string may be more readable than using the DSL.  

For example:
```
JsonPath.root().child("address").child("city");
```
vs.
```
JsonPath.of("$.address.city")
```
Also, JSON path users are accustomed to providing the full path string. The full path string is also what is returned when users Google search  "How to create a JSON path string to navigate to the ....".

This would also reduce quite a bit of API surface area to maintain.

Another concern is w/ the string concatenation - I am not sure how it would surface, but it reminds me of building up a SQL query via string concatenation that could lead to a CVE (not sure exactly how, but it just made me raise my eyebrow).

Wdyt about just providing a `JsonPath.of(String)` for V1?

onobc (changes requested):
👋🏻 @Dgramada ,

Thanks for the PR - it is great to see the feature being added. I have left a few comments to address / points to discuss.

src/main/java/org/springframework/data/redis/core/json/JsonPath.java
Dgramada: Yes you are right, this is more reasonable. I will keep it simple.

Dgramada:
If the interfaces are acceptable, I could close this issue and create a separate one that introduces them alongside their implementations and testing. Would you prefer this approach or would you like me to extend this PR?

src/main/java/org/springframework/data/redis/core/json/JsonPath.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Immutable value object representing a JSONPath expression for use with Redis JSON operations.
+ *
+ * @author Yordan Tsintsov
+ * @see <a href="https://redis.io/docs/latest/develop/data-types/json/path/">Redis JSONPath Documentation</a>
+ * @since 4.3
+ */
+@NullMarked
+public final class JsonPath {
```
onobc: @mp911de what are your thoughts on using a record here instead? The `eq/hc/ts` cruft goes away.

src/main/java/org/springframework/data/redis/core/json/JsonArrayRange.java
onobc: When you get into the impl PTAL at this `Range` and see if it something we would want to re-use.

onobc:
> If the interfaces are acceptable, 
I do - they are at a nice simple starting point.

> I could close this issue and create a separate one that introduces them alongside their implementations and testing. 
> Would you prefer this approach or would you like me to extend this PR? 

I would just rock this PR (continue on from this point).

Thank you for the quick turnaround - I was on PTO - sorry for the delay.

src/main/java/org/springframework/data/redis/core/json/JsonPath.java
mp911de: I suggest that we back off from any JSON path handling. If we feature a JsonPath type, then this suggest that we handle all sorts of escaping which isn't the case and we're not in the business of JSON escaping.

Dgramada: So should I go with raw JSON path strings and avoid the JsonPath wrapper object as a whole?

mp911de: Yes, I think that's what makes currently most sense. We can introduce more functionality always later on if we learn how to improve existing API.

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,493 @@
...
+	 * @param path must not be {@literal null}.
+	 * @param value must not be {@literal null}. {@literal null} values should be represented as JSON "null" values.
+	 * @param start index to start searching from.
+	 * @return a list where each element contains the index of the first occurrence of the value in the array or {@literal null} if path does not exist.
+	 * @see <a href="https://redis.io/docs/latest/commands/json.arrindex/">Redis Documentation: JSON.ARRINDEX</a>
+	 * @since 4.2
+	 */
+	default List<@Nullable Long> jsonArrIndex(byte[] key, String path, String value, long start) {
+
+		Assert.notNull(key, "Key must not be null");
```
onobc: Since the default methods just ultimately chain to the non-default method, this validation will be duplicated because the non-default impl will have to implement the same checks.

> [!NOTE]
> This comment applies to the PR

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,493 @@
...
+
+	/**
+	 * Search for the first occurrence of a JSON value in an array.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param path must not be {@literal null}.
+	 * @param value must not be {@literal null}. {@literal null} values should be represented as JSON "null" values.
+	 * @param start index to start searching from.
+	 * @param stop index to stop searching at.
+	 * @return a list where each element contains the index of the first occurrence of the value in the array or {@literal null} if path does not exist.
```
onobc: Looking at [the command](https://redis.io/docs/latest/commands/json.arrindex/) docs I don't see where the list of return values is coming from. Should it just be a single value?

Dgramada: An example would be:

```
JSON.SET user:1 $ '{
  "groups": [
    {
      "name": "admins",
      "roles": ["read", "write", "delete"]
    },
    {
      "name": "editors",
      "roles": ["read", "write"]
    },
    {
      "name": "viewers",
      "roles": ["read"]
    }
  ]
}'

```
If you call this JSON with the following command:

`JSON.ARRINDEX user:1 $.groups[*].roles "write"`

You would receive the following output:

`[1, 1, -1]`

Because of this, a lot of the methods return a list.

onobc: Ahh I see. It may be helpful to mention this in the class level javadocs w/ an example as this really clarifies it for me.

src/main/java/org/springframework/data/redis/core/JsonOperations.java
```diff
@@ -0,0 +1,608 @@
...
+ * array manipulation, string operations, and numeric operations at specific JSON paths.
+ * <p>
+ * Path arguments are JSON path strings, where {@code $} represents the root element.
+ *
+ * @author Yordan Tsintsov
+ * @see <a href="https://redis.io/docs/latest/develop/data-types/json/">Redis JSON Documentation</a>
+ * @since 4.2
+ */
+@NullUnmarked
+public interface JsonOperations<K> {
```
onobc: RedisOperations is set up by having key and value type info (i.e. `RedisOperations<K, V>`). This menas that value types are always `V`.

The JsonOperations, in contrast requires type info in its API calls and may deviate from the V at RedisOperations<K, V>. 

Imagine `RedisOperations<String, Person>` and now you would want to store `redisOperations.valueOps().set(key, new Order())` because you're doing `jsonOps().get(key, Order.class)`. These concepts don't align.

That being said, the JSON API seems to be able to operate on any types. Me and Mark were chatting about this and will continue the chat tomorrow. Will follow up w/ the outcome here to see if we want to type the `JsonOperations` w/ `K,V` or as you have done it w/ `K`.

Dgramada: My concern with `JsonOperations<K, V>` would be when used with lets say some data structure like `List<E>`.
Then we would have something like `JsonOperations<String, List>` and we would lose the list element type and would have to have to provide it as a parameter in a method overload.

tishun: Hey @onobc ,

me and @Dgramada discussed this question and I realized I don't fully understand it. Let me make sure we are on the same page first:
- you can only call `JSON.GET` on a key that was saved as a JSON with `JSON.SET`. If you call `set my-key my-value` and then `json.get my-key` you will get `(error) Existing key has wrong Redis type`.
- the `V` generic is relatively ambiguous in this case, as it might have or might not have the same semantics of a typical value one stores in their Redis instance

Overall I had the same kind of conundrum when I was designing the Lettuce APIs : whether the payload is best represented by a simple type (`String`, `byte[]`); an complex type (`Person.class` or `V`); or an datatype abstraction (`JsonObject`). IMHO there is no clear and obvious right choice and it all depends on the use case.

What I can say from experience is that the community quickly [provided the feedback](https://github.com/redis/lettuce/issues/3365) that they were actually expecting the simplest solution (`String`) so we had to overload our classes with it.

NB! In Lettuce this also means that we are not able to benefit from early serialization / late deserialization on the client thread and we end up congesting the thread pool that processes IO (since String is final we have no good way to wrap it with this logic).

onobc: Thanks for those details @tishun and @Dgramada . We (team) discussed this and what we would like to do is proceed w/ this PR but focus only on the command layer. This can go into main and be available w/ the 4.1 release. The operations aspect of the PR we will move into a branch or gist for safekeeping and exploration - ultimately targeting 4.2 for the operations / API layer in November.

This approach will give us some time to design the API/operations layer while making the commands available in 4.1.

Sound good?

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,481 @@
...
+		return jsonArrIndex(key, path, value, start, 0);
+	}
+
+	/**
+	 * Search for the first occurrence of a JSON value in an array.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param path must not be {@literal null}.
+	 * @param value must not be {@literal null}. {@literal null} values should be represented as JSON "null" values.
+	 * @param start index to start searching from.
```
onobc: May be helpful to add `inclusive` for `start` and `exclusive` for `stop`.

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,481 @@
...
+	 * @since 4.1
+	 */
+	List<@Nullable String> jsonArrPop(byte[] key, String path, int index);
+
+	/**
+	 * Trim an array so that it contains only the specified inclusive range of elements.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param path must not be {@literal null}.
+	 * @param start index to start trimming from.
```
onobc: Ditto https://github.com/spring-projects/spring-data-redis/pull/3327/changes#r3002905067

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,481 @@
...
+
+	/**
+	 * Arguments for {@code JSON.MSET} command.
+	 *
+	 * @param key the key, must not be {@literal null}.
+	 * @param path the JSON path, must not be {@literal null}.
+	 * @param value the value to set.
+	 * @since 4.1
+	 */
+	record JsonMSetArgs(byte[] key, String path, String value) {
```
onobc: Can we not rely on `@NonNull` in record init params?

src/test/java/org/springframework/data/redis/connection/AbstractConnectionIntegrationTests.java
```diff
@@ -5163,6 +5165,305 @@ void zRangeStoreRevByLexStoresKeys() {
 		assertThat((LinkedHashSet<Object>) result.get(5)).containsSequence(VALUE_3, VALUE_4);
 	}
 
+	@Test // GH-3327
+	@EnabledOnCommand("JSON.ARRAPPEND")
+	void jsoNArrAppend() {
```
onobc: I noticed the tests are not running because there is no `"JSON.*"` command returned from the conditions on the server (i.e. the tests never run).
I believe we need to guard this on version `@EnabledOnRedisVersion(8.2)` (is `8.2` correct version?).


When they do run, you will find that they run into a StackOverflow due to the `DefaultedRedisConnection` not implementing the new methods by invoking the delegate. For example, this fixes the recursion on the `jsonSet`:

```java
	@Override
	public Boolean jsonSet(byte[] key, String value) {
		return this.delegate.jsonSet(key, value);
	}

	@Override
	public Boolean jsonSet(byte[] key, String path, String value, JsonSetOption option) {
		return this.delegate.jsonSet(key, path, value, option);
	}
```

onobc (changes requested):
👋🏻 @Dgramada I left a couple comments to address. Thanks.

src/main/java/org/springframework/data/redis/core/JsonOperations.java
tishun: Makes a lot of sense to me personally, @Dgramada what do you think?

src/test/java/org/springframework/data/redis/connection/AbstractConnectionIntegrationTests.java
Dgramada: The current testing setup that downloads Redis from source shouldn't have the JSON commands. Redis JSON is a separate module that was included in [Redis 8.0.0](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.0-release-notes/#redis-open-source-800-may-2025). 

However the current Makefile setup doesn't turn on the necessary flag to include the modules like JSON, Search, Bloom Filter, etc. Furthermore if we decide to test JSON through source this would require adding additional dependencies for the JSON module to work, which is not ideal. You can find the module flag and all the necessary dependencies for everything to work [here](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.0-release-notes/#redis-open-source-800-may-2025).

I have an [ongoing pull request](https://github.com/spring-projects/spring-data-redis/pull/3336) that migrates the current Makefile setup to use Docker containers instead of downloading Redis from source. Images of Redis 8.0.0 and onwards have all the Redis modules baked in.

Dgramada: I would like to mention that this PR is missing the Jedis implementation because it depends on [#3315](https://github.com/spring-projects/spring-data-redis/pull/3315).

Also there is currently 1 test failing which is the `jsonArrAppend()` test. This might be a bug on the Lettuce side, but I must verify this. If the problem is in Lettuce, we could drop this method and introduce it in the future once it is fixed.

onobc: > **I would like to mention that this PR is missing the Jedis implementation because it depends on https://github.com/spring-projects/spring-data-redis/pull/3315.

Looks like @mp911de just merged that a few hours ago.

> Also there is currently 1 test failing which is the jsonArrAppend() test. This might be a bug on the Lettuce side, but I must verify this. If the problem is in Lettuce, we could drop this method and introduce it in the future once it is fixed.

Good plan - lmk what you find.

onobc: > The current testing setup that downloads Redis from source shouldn't have the JSON commands. Redis JSON is a separate module that was included in [Redis 8.0.0](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.0-release-notes/#redis-open-source-800-may-2025).
> 
> However the current Makefile setup doesn't turn on the necessary flag to include the modules like JSON, Search, Bloom Filter, etc. Furthermore if we decide to test JSON through source this would require adding additional dependencies for the JSON module to work, which is not ideal. You can find the module flag and all the necessary dependencies for everything to work [here](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.0-release-notes/#redis-open-source-800-may-2025).
> 
> I have an [ongoing pull request](https://github.com/spring-projects/spring-data-redis/pull/3336) that migrates the current Makefile setup to use Docker containers instead of downloading Redis from source. Images of Redis 8.0.0 and onwards have all the Redis modules baked in.

 Ahh ok, this makes sense now. Hopefully the docker-compose PR gets approved and merged.

Dgramada: Hey @onobc,
A quick update from me. jsonArrAppend() is broken in Lettuce so I will drop the method from the PR.

src/main/java/org/springframework/data/redis/connection/jedis/JedisJsonCommands.java
```diff
@@ -0,0 +1,210 @@
...
+
+	@Override
+	public List<Long> jsonArrAppend(byte @NonNull [] key, @NonNull String path, @NonNull String @NonNull... values) {
+
+		Assert.notNull(key, "Key must not be null");
+		Assert.notNull(path, "Path must not be null");
+		Assert.notEmpty(values, "Values must not be empty");
+		Assert.noNullElements(values, "Values must not be null");
+
+		return connection.invoke().just(UnifiedJedis::jsonArrAppend, RedisJsonPipelineCommands::jsonArrAppend, new String(key), Path2.of(path), values);
```
viktoriya-kutsarova: Suggest using JedisConverters.toString(key) instead of new String(key).

new String(byte[]) decodes with the JVM default charset, which is only guaranteed UTF-8 on Java 18+ (JEP 400) absent -Dfile.encoding overrides. Jedis re-encodes the String key back to bytes via SafeEncoder.encode (UTF-8) before writing to the wire — chain is RawableFactory.from(String) → new RawString(str) → SafeEncoder.encode(str) → getBytes(UTF_8). For any non-ASCII key bytes the wire representation can differ from the caller's byte[], so a key written via jsonSet won't round-trip against the same bytes via stringCommands().get(...).

JedisConverters.toString wraps SafeEncoder.encode (UTF-8) and matches what Jedis uses on the other side. Same precedent already used at JedisClusterConnection.java:589 and JedisStreamCommands.java:99,116.

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,389 @@
...
+	 * Append the JSON values into the array at path after the last element in it.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param path must not be {@literal null}.
+	 * @param values must not be {@literal null}. {@literal null} values should be represented as JSON "null" values.
+	 * @return a list where each element contains the new length of the array or {@literal null} if path does not exist.
+	 * @see <a href="https://redis.io/docs/latest/commands/json.arrappend/">Redis Documentation: JSON.ARRAPPEND</a>
+	 * @since 4.2
+	 */
+	List<Long> jsonArrAppend(byte @NonNull [] key, @NonNull String path, @NonNull String @NonNull... values);
```
viktoriya-kutsarova: Let's review if the `@NonNull String @NonNull... values` argument should be of type String or a different approach.

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,389 @@
...
+	default String jsonGet(byte @NonNull [] key) {
+		return jsonGet(key, ROOT_PATH);
+	}
+
+	/**
+	 * Get the JSON values at the given key and paths.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @param paths must not be {@literal null}.
+	 * @return list where each element is a JSON value or {@literal null} if path does not exist.
```
viktoriya-kutsarova: Should we maybe change the javadoc for the return type to something like:
```
Returns a JSON-serialized string. When a single path is given, returns the value at that path. When multiple paths are given, returns a JSON object with each path as a key. Returns {@code null} if the key does not exist.
```

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,389 @@
...
+	 * @return a list where each element is the type at the given path.
+	 * @see <a href="https://redis.io/docs/latest/commands/json.type/">Redis Documentation: JSON.TYPE</a>
+	 * @since 4.2
+	 */
+	List<JsonType> jsonType(byte @NonNull [] key, @NonNull String path);
+
+	/**
+	 * {@code JSON.SET} command arguments for {@code NX}, {@code XX}.
+	 */
+	enum JsonSetOption {
```
viktoriya-kutsarova: Let's expand the JsonSetOption enum to a class, similar to `SetCondition`.

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
```diff
@@ -0,0 +1,389 @@
...
+		 *
+		 * @return {@link JsonSetOption#IF_PATH_EXISTS}
+		 */
+		public static JsonSetOption ifPathExists() {
+			return IF_PATH_EXISTS;
+		}
+
+	}
+
+	enum JsonType {
```
viktoriya-kutsarova, line 320: Should we add a NULL type in case we have a null value set on purpose. For example, do we need such a distinction:

```
if (jsonType(key, "$.email").get(0) == JsonType.NULL) {
    // user explicitly set email to null — show "no email"
} else {
    // path missing OR has a value
}
```

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
```diff
@@ -876,6 +868,42 @@ public static HostAndPort toHostAndPort(RedisNode node) {
...
 
+	public static JsonSetParams toJsonSetParams(RedisJsonCommands.JsonSetOption args) {
+		return switch (args) {
+			case IF_PATH_NOT_EXISTS -> new JsonSetParams().nx();
+			case IF_PATH_EXISTS -> new JsonSetParams().xx();
+			default -> new JsonSetParams();
+		};
+	}
+
+	public static List<Number> fromJsonNumberListObject(Object object) {
```
viktoriya-kutsarova: Could you please provide tests for decimal numbers - in case you have null in the response and if you run the command multiple times with a decimal number.

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
Dgramada: Currently we have 3 cases:

1.  JSON value is null -> Lettuce returns `UNKNOWN`, Jedis returns `null`
2. JSON path doesn't exist -> Both drivers return empty list
3. Key doesn't exist -> Lettuce throws exception (likely a driver bug), Jedis returns `null`

Jedis doesn't make a distinction between null JSON values and missing keys, while Lettuce does. Furthermore, there are plans to make improvements in the Jedis API as it currently has multiple issues and is quite different from the Lettuce API.

I would personally remove UNKNOWN for now and just return `null` to keep the Spring Data Redis API minimal and make it more flexible for future extensions once Jedis and Lettuce are more aligned.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
Dgramada: I decided to drop the JSON.NUMINCRBY command. Currently Jedis returns a double list when using RESP 2 and a JSONArray object which converts numbers to BigDecimal when using RESP 3. So in other words Jedis always modifies the type to Double or BigDecimal which would lead to writing loose testing. I would prefer to leave this method out and potentially add it in the future once Jedis addresses these issues - from modifying types to returning external library types (JSONArray is an external dependency type).

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
Dgramada: Lettuce exposes String and JsonValue overloads. Jedis is typed as Object but, at runtime, is invoked with a String and dispatched to its String branch. Both drivers convert values to UTF‑8 bytes and send as is, without Jackson/Gson serialization. Validation happens server-side in the RedisJSON module and the call succeeds if the bytes form a valid JSON document.

Passing raw null as a parameter value throws an exception in both drivers once they start encoding the value.
POJOs would also fail as the class wouldn't pass through an actual JSON serialization.

In my opinion, the best fit would be to accept a String parameter which represents a JSON value.

Example:

```
"5" -> JSON number

""5"" -> JSON string

"true" -> JSON boolean

""true"" -> JSON string

"{
   "firstName": "Rand"
   "lastName": "Al'Thor",
   "address": {
      "name": "Two Rivers"
   }
}" -> JSON object where address points to another JSON object value

"{
   "firstName": "Rand"
   "lastName": "Al'Thor",
   "address": "{
      "name": "Two Rivers"
   }"
}" -> JSON object where address points to a JSON string value
```

mp911de: A plain `String` doesn't carry sufficient semantics here, I'm afraid. Every caller arriving at this API would end up asking themselves the very same question, having to dive into the driver details in order to reconstruct the actual contract and that's a recurring tax we'd rather not impose in the first place.

This may feel counterintuitive at present, but I tend towards generalizing the `JsonValue` notion instead (as in `JsonOperations.JsonValue`). We've effectively arrived at such an abstraction on the template API already; the natural next step is to push the core abstraction down to the connection level quite along the lines of the stream support in `org.springframework.data.redis.connection.stream`. As things stand, it's really only the `JsonValue.as` methods that betray a template-API nature.

Revisiting the template API design with our present insights, I'd refine `JsonValue` towards a `JsonResult extends ….connection.json.JsonValue` arrangement on the template side, moving the remaining pieces into the `connection.json` package proper.

`JsonValue` itself would then define static factory methods: `JsonValue.literal(Number)` along with `int`/`long`/`double`/`String` overloads, `JsonValue.nullValue()`, and `JsonValue.ofJson(String)` for verbatim JSON. All connection-level methods would accept `JsonValue` rather than a raw `String` accordingly.

Let me know what you think; happy to iterate on the specifics.

mp911de:
I think we're closing in on a first solid design here. I'd suggest drawing some inspiration from 28f5665 (on the Json branch in this repository), particularly around `JsonValue`. There ought to be a `JsonType` (analogous to `DataType`) residing in the `connection.json` package, alongside `JsonValue` itself. We may well reuse `JsonType` later on within `JsonResult`.

I'd propose finalizing the `JsonCommands` parts first, so that we can then approach `JsonOperations` and `JsonTemplate` as a separate, second step. Splitting the effort into two pieces keeps the complexity manageable on both ends.

src/main/java/org/springframework/data/redis/connection/RedisJsonCommands.java
Dgramada: Just to clarify, JsonValue will act as a wrapper and won't really do any parsing right? This is a good approach as it gives more leeway for the API if it needs any modification in the future. 

Maybe we should use this approach for the JSONPath as well? We could wrap the string path in a JsonPath similarly to how Jedis and Lettuce do.

src/main/java/org/springframework/data/redis/connection/json/JsonValue.java
```diff
@@ -0,0 +1,112 @@
...
+	 */
+	static JsonValue ofJson(String rawJson) {
+		return new LiteralJsonValue(rawJson);
+	}
+
+	/**
+	 * @return the canonical JSON text of this value.
+	 */
+	@Override
+	String toString();
```
viktoriya-kutsarova: Consider renaming this method so that it is not confused with the default toString().

src/main/java/org/springframework/data/redis/connection/json/LiteralJsonValue.java
```diff
@@ -0,0 +1,42 @@
...
+
+	static final LiteralJsonValue NULL = new LiteralJsonValue("null");
+
+	LiteralJsonValue {
+		Assert.notNull(value, "Value must not be null");
+	}
+
+	static String quote(String value) {
+		Assert.notNull(value, "Value must not be null");
+		return "\"" + value + "\"";
```
viktoriya-kutsarova, line 34: Please revisit this quote method — since the output is parsed as JSON, an unescaped ", \, or control character in value lets a caller break out of the string literal (JSON injection), so we shouldn't be concatenating raw input.

src/main/java/org/springframework/data/redis/connection/json/JsonValue.java
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * JSON value from a JSON document. The supplied text is forwarded unchanged; no validation or escaping is
+	 * performed.
+	 *
+	 * @param rawJson a valid JSON document. Must not be {@literal null}.
+	 * @return a {@link JsonValue} carrying the JSON.
+	 */
+	static JsonValue ofJson(String rawJson) {
```
viktoriya-kutsarova, line 102: Let's revisit this method's name - maybe it show become `raw(String json)`?

src/main/java/org/springframework/data/redis/connection/json/JsonValue.java
```diff
@@ -0,0 +1,111 @@
...
+		return new LiteralJsonValue(Double.toString(number));
+	}
+
+	/**
+	 * JSON string from a Java {@link String}.
+	 *
+	 * @param value must not be {@literal null}.
+	 * @return {@link JsonValue} representing JSON string.
+	 */
+	static JsonValue literal(String value) {
```
viktoriya-kutsarova, line 91: Let's revisit this method's name.

Dgramada: If we change the naming of this method to something like:

`static JsonValue string(String value)`

I would suggest we change the other `literal` methods to `number(long value)`, `bool(boolean value)` and so on so that the string method isn't the odd one out.

mp911de: `JsonValue.of(…)` is probably sufficient. `JsonValue.raw(…)` signals already that it wants to be used with some care.

I would treat this as something to revisit once we are done rather than a blocker for moving forward.

Dgramada:
Both the branch and pull request have become quite bloated. I created a new clean branch with all the current code plus the template layer built on top of it. We can close this pull request and continue with the new clean pull request.

mp911de:
Closing as duplicate of #3390

PR #3333, opened 2026-03-17
Added Redis DIGEST command support for connection and template layer.

Dgramada:
This PR adds support for the new Redis [DIGEST](https://redis.io/docs/latest/commands/digest/) command, available since Redis 8.4.0. The DIGEST command returns the XXH3 hash digest of a string value stored at a key, represented as a 16-character hexadecimal string.

The hash digest can be used for efficient comparison operations without needing to compare full string content. It integrates with Redis's SET command options (IFDEQ/IFDNE) and DELEX command for conditional operations based on value digests.

The PR adds methods for both reactive and synchronous template and connection layers.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
+++ b/src/main/java/org/springframework/data/redis/core/RedisTemplate.java
@@ -95,6 +95,7 @@ import org.springframework.util.CollectionUtils;
  * @author Chen Li
  * @author Vedran Pavic
  * @author Chris Bono
+ * @author Yordan Tsintsov
  * @param <K> the Redis key type against which the template works (usually a String)
  * @param <V> the Redis value type against which the template works
  * @see StringRedisTemplate
@@ -577,7 +578,7 @@ public class RedisTemplate<K, V> extends RedisAccessor implements RedisOperation
 	}
 
 	@Override
-	public @Nullable String getDigest(@NonNull K key) {
+	public String getDigest(@NonNull K key) {
 
 		byte[] rawKey = rawKey(key);
 
--- a/src/test/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveKeyCommandsIntegrationTests.java
+++ b/src/test/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveKeyCommandsIntegrationTests.java
@@ -54,6 +54,7 @@ import org.springframework.data.redis.test.condition.EnabledOnRedisVersion;
  * @author Christoph Strobl
  * @author Mark Paluch
  * @author Dahye Anne Lee
+ * @author Yordan Tsintsov
  */
 @ParameterizedClass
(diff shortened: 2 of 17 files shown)


src/main/java/org/springframework/data/redis/core/RedisTemplate.java
```diff
@@ -576,6 +576,14 @@ public Boolean copy(K source, K target, boolean replace) {
 		return doWithKeys(connection -> connection.copy(sourceKey, targetKey, replace));
 	}
 
+	@Override
+	public @Nullable String getDigest(@NonNull K key) {
```
onobc: We can remove `@Nullable` from return type to be consistent w/ other methods since the class is `@NullUnmarked`. 

> [!NOTE]
> This applies in multiple places in this PR - but just adding this one comment to cover them all

src/test/java/org/springframework/data/redis/connection/lettuce/LettuceReactiveKeyCommandsIntegrationTests.java
```diff
@@ -92,6 +92,45 @@ void existsKeyReturnsZeroWhenKeysDoNotExist() {
 				.expectNext(0L).verifyComplete();
 	}
 
+	@Test
```
onobc: Please add the issue/pr number comment to all new tests:
```
// GH-3333
```
> [!NOTE]
> This applies to all tests in this PR - but just adding this one comment to cover them all

onobc (changes requested):
Thanks for adding this @Dgramada ! I have left some minor comments to address and also can you add yourself as `@author` in all touched files? Other than that, looks good.

src/main/java/org/springframework/data/redis/core/RedisOperations.java
```diff
@@ -196,6 +196,17 @@ public interface RedisOperations<K, V> {
...
+	/**
+	 * Get the hash digest for the value stored in the specified key as a hexadecimal string. This command is intended to
+	 * be used with string values only.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @return the digest of the key. {@literal null} when used in pipeline / transaction.
+	 * @see <a href="https://redis.io/commands/digest">Redis Documentation: DIGEST</a>
+	 * @since 4.1
+	 */
+	String getDigest(@NonNull K key);
```
onobc, line 208: In the case where the key does not exist it returns a null bulk string. Does that get turned into a `null` or is it an actual [null bulk string](https://redis.io/docs/latest/develop/reference/protocol-spec/#null-bulk-strings)?

onobc:
Thanks for another excellent contribution @Dgramada !

src/main/java/org/springframework/data/redis/core/RedisOperations.java
Dgramada: I believe it should return a literal `null` for blocking or an empty `Mono` for reactive.

PR #3336, opened 2026-03-24
Migrate testing from source to Docker

Dgramada:
### Description:

Replace the existing test infrastructure that downloads and compiles Redis from source with Docker Compose containers.

### Motivation
The current testing setup is not ideal for future features that include Redis modules (JSON, Search, Bloom filter, etc.). When using source, testing would require developers to install extra dependencies like `clang` and would increase the Makefile complexity.

Docker Compose gives us pre-built images with modules already included and faster startup.

### What changed
Makefile `start`, `stop`, and `clean` commands now delegate to docker compose. The `IMAGE` environment variable controls which Redis/Valkey image is used. The test and all-tests targets now properly preserve the Maven exit code before cleanup.

Docker Compose (src/test/resources/docker-env/docker-compose.yml) — defines the same topology as before: 
- 4 standalone nodes (master + 2 replicas + 1 auth-protected)
- 4 sentinels (3 standard + 1 auth-protected)
- 4 cluster nodes with automatic initialization. 

All services use `network_mode: host` so tests connect to the same 127.0.0.1:<port> addresses as before. Healthchecks ensure services are ready before dependents start.

CI action (.github/actions/setup-redis/action.yml) — instead of passing `VERSION`, `PROJECT`, and `GH_ORG` for source compilation, the action now passes a single `IMAGE` env var. The Redis binary cache step was removed since Docker image layers are handled by the container runtime.

Test fix — corrected `hpExpireAtReturnsSuccessAndSetsTTL` which was calling hTtl (seconds) instead of hpTtl (milliseconds) and asserting the wrong range.

**Important:** The CI pipeline will probably need adjustment and CI configs will probably need to be updated from using project version and name to the equivalent testing images.

---


src/test/java/org/springframework/data/redis/connection/AbstractConnectionIntegrationTests.java
```diff
@@ -3733,12 +3711,12 @@ void hpExpireAtReturnsSuccessAndSetsTTL() {
 		long inFiveSeconds = Instant.now().plusSeconds(5L).toEpochMilli();
 
 		actual.add(connection.hpExpireAt("hash-hexpire", inFiveSeconds, "key-2"));
-		actual.add(connection.hTtl("hash-hexpire", "key-2"));
+		actual.add(connection.hpTtl("hash-hexpire", "key-2"));
```
mp911de: I am not quite sure this change meets the description of the pull request description. HTTL is retrieving seconds and assertions check 0 to 5 second range. Now, the check is using milliseconds without actually fixing something.

Dgramada:
This version of the testing works for Linux but it would require Docker Desktop for MacOS. Also it would require enabling shared network mode in Docker Desktop. 

I understand this is not ideal as it requires developers that use MacOS to do additional setup, so I will change my approach and try another solution.

mp911de:

PR #3337, opened 2026-03-25
Fix GH-3080: recover from RedisInvalidSubscriptionException on concurrent pub/sub

### LINKED ISSUE #3080: Pub/Sub in a concurrent scenario may result in the inability to use subscriptions anymore, which can ultimately lead to memory leaks
### Problem Description
- In certain concurrent scenarios, we have observed that the pub/sub mechanism can result in subscription failures and subsequently lead to memory leaks.

#### addListener Method
1. Initially, the listener is added to the `listenerTopics` collection.
2. Subsequently, the `subscribeChannel` method is invoked to establish the subscription.

#### org.springframework.data.redis.connection.util.AbstractSubscription#subscribe Method
1. The method checks if the `alive` flag is set to true. If it is not, a `RedisInvalidSubscriptionException` is thrown.

#### removeMessageListener Method
1. The listener is first removed from the `listenerTopics` collection.
2. Next, it is verified whether `listenerTopics` is empty. If so, the `stopListening` method is called.
3. Following this, the channel is removed from the subscription's channels, and the `closeIfUnsubscribed` method is invoked.

#### org.springframework.data.redis.connection.util.AbstractSubscription#closeIfUnsubscribed Method
1. The method checks if the channels collection is empty. If it is, the `alive` flag is set to false using `alive.compareAndSet(true, false)`.

### Bug Analysis
- Steps 2 and 3 of the `removeMessageListener` method introduce a potential race condition. During the interval between these steps, it is possible for step 2 to detect that `listenerTopics` is not empty while step 3 detects that the channel is empty (resulting in setting `alive` to false).
- Consequently, any subsequent calls to `addListener` will fail because the subscription process is deemed unsuccessful. As a result, applications typically do not invoke `removeMessageListener` again, leading to the listener remaining in the `listenerTopics` collection indefinitely.

#### change code (RedisMessageListenerContainer) 
```java
public class RedisMessageListenerContainer implements InitializingBean, DisposableBean, BeanNameAware, SmartLifecycle {
    // ==============add  code start
	private static final AtomicInteger counter = new AtomicInteger(0);
	private static final AtomicBoolean firstListenerHasRemoved = new AtomicBoolean(false);
    // ==============add  code end

    // ==============add  code start
	private static final AtomicInteger counter = new AtomicInteger(0);
	private static final AtomicBoolean firstListenerHasRemoved = new AtomicBoolean(false);
   
    private void addListener(MessageListener listener, Collection<? extends Topic> topics) {
        // ....
        for (Topic topic : topics) {
            // ...
        }
        //  ============== add code start
        boolean wasListening = isListening();
        counter.incrementAndGet();
        while (counter.get() == 2 && !firstListenerHasRemoved.get()) { }
        // ============== add code end
        if (isRunning()) {
            lazyListen();
            // .....
        }
    }

```

#### add test case to RedisMessageListenerContainerIntegrationTests
```java
	@ParameterizedRedisTest
	void subscribeErrorAfterStart() throws Exception {

		CompositeListener listener1 = (message, pattern) -> { };
		CompositeListener listener2 = (message, pattern) -> { };

		container.start();
		container.addMessageListener(listener1, new ChannelTopic("a"));
		CompletableFuture<Exception> subscribeExceptionFuture = new CompletableFuture<>();
		new Thread(() -> {
			try {
				container.addMessageListener(listener2, new ChannelTopic("b"));
				subscribeExceptionFuture.complete(null);
			} catch (Exception e) {
				subscribeExceptionFuture.complete(e);
			}
		}).start();
		new Thread(() -> {
			container.removeMessageListener(listener1);
		}).start();
		assertThat(subscribeExceptionFuture.get()).isNull();
		container.destroy();
	}

```

suuuuuuminnnnnn:
## Problem

Concurrent `addMessageListener` + `removeMessageListener` calls can cause a race
condition in `RedisMessageListenerContainer`:

1. Thread A calls `addMessageListener(listener2, "b")` → updates `channelMapping`
   + `listenerTopics`, then prepares to call `subscribeChannel(["b"])`
2. Thread B calls `removeMessageListener(listener1, "a")` →
   `unsubscribeChannel(["a"])` → `AbstractSubscription.channels` becomes empty
   → `closeIfUnsubscribed()` → `close()` → **`alive = false`**
3. Thread A resumes → `subscribeChannel(["b"])` → `checkPulse()` →
   `alive == false` → **`RedisInvalidSubscriptionException`** propagates
4. `listener2` remains in `listenerTopics` without an active subscription
   → **memory leak**

This race only manifests with async drivers (Lettuce). Jedis uses
`BlockingSubscriber` which sets `connection = null` after `doSubscribe`,
making incremental `subscribeChannel` a no-op.

## Fix

In the `wasListening` branch of `addListener()`, catch
`RedisInvalidSubscriptionException` and recover:
```java
try {
    if (!channels.isEmpty()) {
        getRequiredSubscriber().subscribeChannel(...);
    }
    if (!patterns.isEmpty()) {
        getRequiredSubscriber().subscribePattern(...);
    }
} catch (RedisInvalidSubscriptionException ex) {
    stopListening();  // dispose dead connection
    lazyListen();     // fresh connection + re-subscribe all from channelMapping/patternMapping
    return;
}
```

`stopListening()` does not change the `started` flag, so the container remains
running. `lazyListen()` re-subscribes all topics including the newly added one
(already in `channelMapping` before the exception).

## Testing

Added unit test `shouldRecoverWhenSubscriptionDiesOnConcurrentRemove` that:
- Simulates the race by making `subscriptionMock.subscribe()` throw
  `RedisInvalidSubscriptionException`
- Verifies no exception propagates to the caller
- Verifies recovery triggers a second `connection.subscribe()` call
- Verifies `listenerTopics` contains the new listener (no memory leak)
- Verifies `isListening() == true` after recovery

Closes #3080

---

### DIFF
--- a/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
+++ b/src/main/java/org/springframework/data/redis/listener/RedisMessageListenerContainer.java
@@ -55,6 +55,7 @@ import org.springframework.data.redis.connection.Message;
 import org.springframework.data.redis.connection.MessageListener;
 import org.springframework.data.redis.connection.RedisConnection;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.data.redis.connection.RedisInvalidSubscriptionException;
 import org.springframework.data.redis.connection.Subscription;
 import org.springframework.data.redis.connection.SubscriptionListener;
 import org.springframework.data.redis.connection.util.ByteArrayWrapper;
@@ -716,8 +717,24 @@ public class RedisMessageListenerContainer implements InitializingBean, Disposab
 
 				getRequiredSubscriber().addSynchronization(new SynchronizingMessageListener.SubscriptionSynchronization(
 						patterns, channels, () -> future.complete(null)));
-				getRequiredSubscriber().subscribeChannel(channels.toArray(new byte[channels.size()][]));
-				getRequiredSubscriber().subscribePattern(patterns.toArray(new byte[patterns.size()][]));
+
+				try {
+					if (!channels.isEmpty()) {
+						getRequiredSubscriber().subscribeChannel(channels.toArray(new byte[channels.size()][]));
+					}
+					if (!patterns.isEmpty()) {
+						getRequiredSubscriber().subscribePattern(patterns.toArray(new byte[patterns.size()][]));
+					}
+				} catch (RedisInvalidSubscriptionException ex) {
+					// Race with removeMessageListener: concurrent unsubscribe made the connection-level
+					// subscription go dead (closeIfUnsubscribed -> alive=false) while we were about to
+					// subscribe new channels. stopListening() disposes the dead connection; lazyListen()
+					// opens a fresh connection and re-subscribes all topics in channelMapping/patternMapping,
+					// which already include the channels/patterns added above.
+					stopListening();
+					lazyListen();
+					return;
+				}
 
 				try {
 					future.join();
--- a/src/test/java/org/springframework/data/redis/listener/RedisMessageListenerContainerUnitTests.java
+++ b/src/test/java/org/springframework/data/redis/listener/RedisMessageListenerContainerUnitTests.java
@@ -19,7 +19,9 @@ import static org.assertj.core.api.Assertions.*;
 import static org.mockito.Mockito.*;
 
 import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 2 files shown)


rferreira:
any chance we can get this merged?

PR #3341, opened 2026-04-09
Use default RedisMessageListenerContainer bean name for annotation-driven Redis listeners

### LINKED ISSUE #3340: Redis listeners should operate on a default RedisMessageListenerContainer, even if multiple beans are present
`@EnableRedisListeners` is somewhat similar to what we're doing for JMS and Kafka. Working on providing support for this feature in Spring Boot, I've noticed a discrepancy that is quite annoying. 

Spring Data Redis relies to a single candidate for the `RedisMessageListenerContainer` to use.  JMS, on the other hand, always relies on a default bean named `jmsListenerContainerFactory`. Ignoring consistency, relying on a bean named `redisMessageListenerContainer` has the following advantages:

* It makes it obvious that the default has to have some sort of well-defined characteristics
* It continues to work if an additional bean is registered

Right now, if you define a bean and have a bunch of listeners with no container reference, adding an additional container would make those one fails. You'd then either need to make the existing bean `@Primary` or add a reference to the existing endpoint. 

Requiring the bean to be primary looks a bit too invasive to me. Please consider harmonizing with how JMS/Kafka behave.

mp911de:
We now use a default bean name to ensure that additional `RedisMessageListenerContainer` bean registrations do not break applications by using a default bean name. Also, we align our behavior with other messaging components in the Spring portfolio.

Closes #3340

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>4.1.0-SNAPSHOT</version>
+	<version>4.1.0-GH-3340-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 	<description>Spring Data module for Redis</description>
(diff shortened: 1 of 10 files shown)


christophstrobl:
LGTM

PR #3345, opened 2026-04-10
Fix sort direction ignored in GeoSearchStoreCommandArgs.sort().

### LINKED ISSUE #3342: Fix sort direction ignored in GeoSearchStoreCommandArgs.sort()
`GeoSearchStoreCommandArgs.sort(Direction)` always assigns `Direction.ASC` regardless of the provided argument, causing `sortDescending()` to have no effect.

It would be better to assign the provided `direction` argument directly, so that the method behaves as documented.

yeonsu00:
## Problem

`GeoSearchStoreCommandArgs.sort(Direction)` always assigned `Direction.ASC`
regardless of the provided argument:

```java
public GeoSearchStoreCommandArgs sort(Direction direction) {
    Assert.notNull(direction, "Sort direction must not be null");
    sortDirection = Direction.ASC; // direction parameter ignored
    return this;
}
```

This caused `sortDescending()` to have no effect, and the SORT DESC flag
was never passed to the underlying Redis GEOSEARCHSTORE command in either
Lettuce or Jedis drivers.

The equivalent method in `GeoSearchCommandArgs` is implemented correctly and
served as a reference.

## Fix

Assign the provided direction argument directly:

```java
sortDirection = direction;
```

## Testing

Added integration test `geoSearchAndStoreShouldRespectSortDirection` that:

- Adds three members at different distances from a reference point (Palermo, Agrigento, Catania)
- Calls `searchAndStore` with `sortDescending().limit(1)`
- Verifies that only the farthest member (Catania) is stored in the destination key

Closes #3342

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/RedisGeoCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/RedisGeoCommands.java
@@ -44,6 +44,7 @@ import org.springframework.util.Assert;
  * @author Ninad Divadkar
  * @author Christoph Strobl
  * @author Mark Paluch
+ * @author Yeonsu Kim
  * @since 1.8
  * @see RedisCommands
  */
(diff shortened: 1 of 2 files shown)


christophstrobl:
LGTM

christophstrobl:
Thank you @yeonsu00 - merged to _main_ development line and back ported to _4.0.x_ & _3.5.x_.

PR #3360, opened 2026-04-27
Add SslOptions support to JedisClientConfiguration

### LINKED ISSUE #3355: Add support for configuring Jedis `SslOptions`
We do support `SslSocketFactory`, `HostnameVerifier`, and `SslParameters` and for consistency, we should also support `SslOptions` for easier trust material configuration.

gggeon96:
## Description

Adds `SslOptions` support to `JedisClientConfiguration` for consistency with the existing `SSLSocketFactory`, `SSLParameters`, and `HostnameVerifier` options, as discussed in #3355.

```java
JedisClientConfiguration.builder()
    .useSsl()
    .sslOptions(SslOptions.builder()
        .truststore("truststore.p12", "secret")
        .sslProtocol("TLSv1.3")
        .build())
    .build();
```

The Mutable cleanup follows your direction: the `sslSocketFactory`, `sslParameters`, and `hostnameVerifier` fields/setters had no invocations, so they are removed; the corresponding getters now return `Optional.empty()`.

## Open question

Javadoc uses `@since 4.2` to match the current 4.2.0-SNAPSHOT main. Happy to retarget if you prefer a different release line.

Closes #3355

---

- [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don't submit any formatting related changes.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/DefaultJedisClientConfiguration.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/DefaultJedisClientConfiguration.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.redis.connection.jedis;
 
+import redis.clients.jedis.SslOptions;
+
 import java.time.Duration;
 import java.util.Optional;
 
@@ -30,6 +32,7 @@ import org.jspecify.annotations.Nullable;
  *
  * @author Mark Paluch
  * @author Christoph Strobl
+ * @author Geonhyeon Kim
  * @since 2.0
  */
 class DefaultJedisClientConfiguration implements JedisClientConfiguration {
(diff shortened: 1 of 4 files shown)


mp911de:
Jedis deprecated their `sslParameters`, `sslSocketFactory`, and `ssl` methods and we should follow suit. Introducing `SslOptions` as replacement API makes sense.

gggeon96:
Added the deprecations. `getSslSocketFactory()`, `getSslParameters()` and the matching builder methods are now `@Deprecated(since = 4.2)` pointing to `sslOptions`. I left `useSsl()` alone since it is the entry point to the ssl builder that `sslOptions` hangs off, and `hostnameVerifier` since Jedis did not deprecate it. The factory still applies the deprecated settings so existing configurations keep working.

PR #3361, opened 2026-04-27
Deprecate StringToPropertiesConverter

### LINKED ISSUE #3020: Duplicate code in StringToPropertiesConverter class
```java

	@Override

			info.load(stringReader);
		} catch (Exception ex) {
		}
		return info;
	}
}
```

answndud:
Closes #3020.

`StringToPropertiesConverter` duplicates `Converters.toProperties(String)`. This change deprecates the standalone converter and delegates to the shared converter implementation so Redis INFO string parsing is kept in one place.

Tests: `./mvnw -Dtest=ConvertersUnitTests test`


mp911de:
Fixed via cc0054fe7713a91090fe37b0671865ce751ca7eb

PR #3365, opened 2026-05-10
Deprecate duplicate String properties converter

### LINKED ISSUE #3020: Duplicate code in StringToPropertiesConverter class

akimoovv:
Deprecates `StringToPropertiesConverter` and delegates its conversion logic to the shared `Converters.toProperties(String)` implementation.

This removes the duplicate String-to-Properties parsing path while preserving the existing converter type for compatibility. A focused converter test was added to cover Redis INFO-style property conversion through the shared converter API.

Closes gh-3020.

Tests:
- `./mvnw -Dtest=ConvertersUnitTests test`
- `./mvnw -DskipTests verify`

Note: `./mvnw verify` was also attempted, but local integration tests require Redis on localhost:6379 and failed because Redis was not available.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/convert/StringToPropertiesConverter.java
+++ b/src/main/java/org/springframework/data/redis/connection/convert/StringToPropertiesConverter.java
@@ -15,29 +15,22 @@
  */
 package org.springframework.data.redis.connection.convert;
 
-import java.io.StringReader;
 import java.util.Properties;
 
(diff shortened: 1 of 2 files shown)


mp911de:
Fixed via cc0054fe7713a91090fe37b0671865ce751ca7eb

PR #3366, opened 2026-05-12
Reuse JsonNode in GenericJackson2JsonRedisSerializer

### LINKED ISSUE #3012: Avoid potential duplicate parse operation within GenericJackson2JsonRedisSerializer.
In order to read a potential type key json source might be parsed multiple times which imposes a computation overhead. This could be avoided by reusing already parsed JsonNodes as outlined in https://github.com/spring-projects/spring-data-redis/pull/2999/commits/8aaf501b73a8c74df7db839793d55b6e83720835.

DragonFSKY:
Closes #3012

This change avoids parsing the same payload twice when `GenericJackson2JsonRedisSerializer` resolves the target type from default typing metadata.

`TypeResolver` now returns the resolved `JavaType` together with the parsed `JsonNode`. For serializer-managed default readers, deserialization can reuse that tree through a `TreeTraversingParser` instead of parsing the original byte array again.

Custom `Jackson2ObjectReader` instances continue to receive the original `byte[]` source, and the existing protected `resolveType(...)` contract remains unchanged.

Tests:
- `./mvnw -s settings.xml -Dtest=GenericJackson2JsonRedisSerializerUnitTests#defaultReaderReusesJsonNodeParsedForTypeResolution test`
- `./mvnw -s settings.xml -Dtest=GenericJackson2JsonRedisSerializerUnitTests,GenericJacksonJsonRedisSerializerUnitTests,Jackson2JsonRedisSerializerTests,JacksonJsonRedisSerializerTests test`

### DIFF
--- a/src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
+++ b/src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
@@ -52,6 +52,7 @@ import com.fasterxml.jackson.databind.jsontype.TypeSerializer;
 import com.fasterxml.jackson.databind.jsontype.impl.StdTypeResolverBuilder;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.TextNode;
+import com.fasterxml.jackson.databind.node.TreeTraversingParser;
 import com.fasterxml.jackson.databind.ser.SerializerFactory;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
 import com.fasterxml.jackson.databind.type.TypeFactory;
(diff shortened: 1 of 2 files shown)


mhalbritter:
Thanks for the PR!

We're going to decline this one. Reusing the `JsonNode` for the value costs number fidelity: building the tree coerces floating-point literals to `DoubleNode` before any type id is applied, and no later type information can recover them.

```
wire: {"@class":"java.util.LinkedHashMap","bigDecimal":["java.math.BigDecimal",1.0000000000000000000000001]}
tree: {"@class":"java.util.LinkedHashMap","bigDecimal":["java.math.BigDecimal",1.0]}
```

`GenericJackson2JsonRedisSerializer` is also deprecated for removal since 4.0, so we'd rather not add reader-identity special-casing to it.

#3012 will hopefully be fixed with #3408: GenericJacksonJsonRedisSerializer scans the token stream for the type hint instead of building a tree: the scan stops at the hint and skips values without materialising them, so there's no tree and no second full parse.

PR #3382, opened 2026-06-10
GH-3099: Parse Redis INFO output without using Properties.load()

### LINKED ISSUE #3020: Duplicate code in StringToPropertiesConverter class

### LINKED ISSUE #3099: Redis INFO output is parsed as (Java) properties, although it is not the same format
The issue is the same as here https://github.com/spring-projects/spring-data-redis/issues/1281
That issue was closed because:

> Redis on Windows is [no longer supported](https://github.com/MicrosoftArchive/redis). As this variant has no future I'm inclined to not include any Windows-specific changes to Spring Data Redis

There are Redis variants for Windows that are newer than version 3.
https://github.com/ponfee/embedded-redis/tree/master currently uses 6.2.7.
https://github.com/redis-windows/redis-windows has version 7.

Although we're not using the Windows versions in production, they are helpful for developers using a Windows machine.

Technically speaking, the INFO output should not be parsed as Properties as it is not encoded as such.

My current workaround is to set the TEMP/TMP environment variables so that they don't point to `c:\users\...`, as the `\u....` raises the exception.
I've tried to start Redis as `c:/.../redis-server.exe` but the changed separator only works for the configuration file, not exe path in the INFO output.

Would you reconsider the previous decision?

won-seoop:
## Summary

Closes #3099. Closes #3020.

`Properties.load()` interprets `\u` sequences as Unicode escapes. Redis INFO output can contain Windows-style paths in fields like `exe_path`, `config_file`, `rdb_filename`, and `aof_filename` that include sequences such as `c:\users\...`. When these paths pass through `Properties.load()`, the `\u` prefix triggers a Unicode parsing error (or silently corrupts the value).

The Redis INFO format is a simple `key:value` text format — it is **not** a Java `.properties` file and should not be parsed as one.

## Changes

**Fixes #3099:** Replaced the `Properties.load()`-based implementation in `Converters.toProperties(String)` with a direct line-by-line parser that:
- Skips blank lines and section-header comments (`# Server`, `# Memory`, etc.)
- Splits each line on the **first** colon only (matching Redis INFO's `key:value` format)
- Does **not** interpret any escape sequences, preserving values exactly as Redis emits them

**Fixes #3020:** `StringToPropertiesConverter` now delegates to `Converters.toProperties(String)` rather than duplicating the parsing logic, eliminating the duplication flagged in that issue.

## Test plan

- [ ] Existing tests pass (no behaviour change for normal INFO output)
- [ ] Manually verify that INFO output containing `c:\users\...` paths (Windows Redis) is parsed without error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
+++ b/src/main/java/org/springframework/data/redis/connection/convert/Converters.java
@@ -15,7 +15,6 @@
  */
 package org.springframework.data.redis.connection.convert;
 
-import java.io.StringReader;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.*;
@@ -30,7 +29,6 @@ import org.springframework.data.geo.GeoResult;
 import org.springframework.data.geo.GeoResults;
 import org.springframework.data.geo.Metric;
 import org.springframework.data.geo.Metrics;
-import org.springframework.data.redis.RedisSystemException;
 import org.springframework.data.redis.connection.ClusterSlotHashUtil;
 import org.springframework.data.redis.connection.DataType;
 import org.springframework.data.redis.connection.RedisClusterNode;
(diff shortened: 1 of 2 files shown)


lh0156:
Nice fix. I looked into this locally and I think a few additional regression cases would make the coverage stronger, since the previous `Properties.load()` path could also apply Java properties escape/continuation semantics.

Cases that seem worth covering:
- Windows-style paths containing `\u`, such as `c:\users\...`
- malformed Unicode-like sequences such as `\u12GZ`
- trailing backslash values, to ensure Java properties line continuation is not applied
- values containing additional colons, e.g. IPv6 endpoints or `host:port`
- invalid INFO lines mixed with valid entries
- duplicate keys preserving `Properties` last-write-wins behavior

I have a local version of these tests if it would be useful.

mp911de:
Thank you for your contribution, that's fixed now through b4295b26d8182ca4ca69064bde351bf45284079b from a combination of suggestions across the related pull requests.

PR #3388, opened 2026-06-24
Jedis fails on TTL parameters exceeding integer range

### LINKED ISSUE #3386: Jedis fails on TTL parameters exceeding integer range
use `long` instead of `int`.

DragonFSKY:
Closes #3386.

Jedis supports `long` TTL values for `RESTORE`. Pass `ttlInMillis` without narrowing in both the normal and `REPLACE` paths, and remove the obsolete integer-range guard.

A focused parameterized test covers both variants with a TTL above `Integer.MAX_VALUE`.

Verification:
- `./mvnw --no-transfer-progress --quiet clean test -Pci` (12,156 tests, 0 failures/errors, using the Redis environment from `make start`)
- `git diff --check`

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisKeyCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisKeyCommands.java
@@ -63,6 +63,7 @@ import org.springframework.util.ObjectUtils;
  * @author Yordan Tsintsov
  * @author Tihomir Mateev
  * @author Tiefang Hu
+ * @author Dongliang Xie
  * @since 2.0
  */
 @NullUnmarked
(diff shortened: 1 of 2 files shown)


renechoi (commented):
This needs a rebase. Merging onto current main auto-merges `JedisKeyCommands` with no conflict but keeps the range check that a2e8ad2 (GH-3437, Sep 8) hoisted above the `replace` branch, while dropping your casts. Only the test file conflicts; keeping your version of it, `restoreShouldPassLongTtlToJedis` errors on both params with `TtlInMillis must be less than Integer.MAX_VALUE` at JedisKeyCommands:413.

Supports the change: Jedis 8.0.1 declares `restore(byte[], long, byte[])`, while `lpop(byte[], int)` is still int, so the b251c6a guards stay needed.

DragonFSKY:
Thanks for the review! I’ve rebased onto `main` and removed the relocated RESTORE TTL check, keeping the `lpop`/`rpop` guards unchanged.

mhalbritter:
Thanks @DragonFSKY !

PR #3390, opened 2026-06-29
Add support for Redis JSON

Dgramada:
This is a follow up pull request to #3327 that handles the new JSON API for Spring Data Redis. 
It builds upon the command layer from the referenced PR and adds API for the template layer.

Notes:
- `BoundJsonOperations` is not included yet. Once a version of `JsonOperations` is approved, the bound operations can be included in this branch.
- `RedisJsonTemplate` initially felt like a good idea. It creates a separation of concerns, which is nice because the new JSON API does not pair well with a specific `V` type unlike the other data types in `RedisTemplate`. However, `RedisJsonTemplate` feels a little off and maybe a more elegant solution could be derived. It feels more like a configuration class. Maybe we could rename it to `RedisJsonConfig`. This would also promote developers to inject `JsonOperations` directly.
- `RedisJsonSerializer` does not extend `RedisSerializer`, due to it bringing a lot of functionality that doesn't add any value to the JSON serialization. Furthermore, keeping them separate gives separation of concerns. Also Jedis and Lettuce do not provide byte[] APIs for the JSON commands in Redis, thus making `RedisSerializer` incompatible.

---


src/main/java/org/springframework/data/redis/connection/json/JsonPath.java
```diff
@@ -0,0 +1,52 @@
...
+		return DefaultJsonPath.ROOT;
+	}
+
+	/**
+	 * JSON path from a {@code String}.
+	 *
+	 * @param path the JSON path. Must not be {@literal null}.
+	 * @return {@link JsonPath} representing JSON path.
+	 */
+	static JsonPath of(String path) {
```
mp911de: This method suggests a proper handling and parsing. I'd name it `raw` to indicate what to expect from the method.

src/main/java/org/springframework/data/redis/core/JsonOperations.java
```diff
@@ -0,0 +1,561 @@
...
+	 */
+	JsonStringSpec string(K key);
+
+	/**
+	 * Start building a JSON value operation for the given {@code key}.
+	 *
+	 * @param key must not be {@literal null}.
+	 * @return a spec for specifying the value operation.
+	 */
+	JsonAtKeySpec key(K key);
```
mp911de: we wanted to call this entry point `value`, right?

src/main/java/org/springframework/data/redis/core/RedisJsonTemplate.java
```diff
@@ -0,0 +1,162 @@
...
+ * @author Yordan Tsintsov
+ * @since 4.2
+ */
+public class RedisJsonTemplate<K> extends RedisAccessor implements BeanClassLoaderAware {
+
+	private boolean enableTransactionSupport = false;
+	private boolean initialized = false;
+	private @Nullable ClassLoader classLoader;
+
+	private @Nullable RedisSerializer<K> keySerializer;
```
mp911de: It would make sense to strengthen the contract and make these `final`

src/main/java/org/springframework/data/redis/core/RedisJsonTemplate.java
```diff
@@ -0,0 +1,162 @@
...
+	private final JsonOperations<K> jsonOperations = new DefaultJsonOperations<>(this);
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public void afterPropertiesSet() {
+
+		super.afterPropertiesSet();
+
+		if (keySerializer == null) {
+			keySerializer = (RedisSerializer<K>) new JdkSerializationRedisSerializer(classLoader != null ? classLoader : this.getClass().getClassLoader());
```
mp911de: Let's default to `String` serializers. Zooming out, a constructor `RedisJsonTemplate(RedisConnectionFactory)` with `String and a default JSON `serializer` (Jackson 2/Jackson 3) makes sense. Also, a constructor `RedisJsonTemplate(RedisConnectionFactory, RedisSerializer<String>, RedisJsonSerializer)` as alternative to provide serializers.

Moving off mutable state is a good thing. We can still keep transaction support but even for transactions I am not sure we should do this.

src/main/java/org/springframework/data/redis/core/RedisJsonTemplate.java
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * Helper class that simplifies Redis JSON data access.
+ * <p>
+ * Keys are serialized with {@link JdkSerializationRedisSerializer} by default, which produces binary (non human-readable)
+ * key representations. Configure a key serializer through {@link #setKeySerializer(RedisSerializer)}.
+ *
+ * @author Yordan Tsintsov
+ * @since 4.2
+ */
+public class RedisJsonTemplate<K> extends RedisAccessor implements BeanClassLoaderAware {
```
mp911de: `RedisJsonTemplate<K> implements RedisJsonOperations`, we can drop the `BeanClassLoaderAware` interface to simplify the design. By moving the config responsibility towards the calling code, we can remove certain requirements and assumptions on our side.

src/main/java/org/springframework/data/redis/serializer/JacksonRedisJsonSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.util.Assert;
+
+/**
+ * Implementation of {@link RedisJsonSerializer} with Jackson 3.
+ *
+ * @author Yordan Tsintsov
+ * @since 4.2
+ */
+public class JacksonRedisJsonSerializer implements RedisJsonSerializer {
```
mp911de: Instead of a new class, we should retrofit `GenericJacksonJsonRedisSerializer` to implement `RedisJsonSerializer`.

src/main/java/org/springframework/data/redis/serializer/RedisJsonSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Serializer for converting Objects to and from their JSON {@link String} representation. It is recommended that
+ * implementations handle {@literal null} objects on serialization (by writing JSON {@code null}) and JSON {@code null}
+ * on deserialization (by returning {@literal null}). Note that Redis does not accept {@literal null} keys or values but
+ * can return {@code null} replies for non-existing keys.
+ *
+ * @author Yordan Tsintsov
+ * @since 4.2
+ */
+public interface RedisJsonSerializer {
```
mp911de: `RedisJsonSerializer extends RedisSerializer<Object>`

mp911de (commented):
Here's an initial round of review feedback.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
```diff
@@ -876,6 +868,23 @@ public static HostAndPort toHostAndPort(RedisNode node) {
...
+	public static JsonSetParams toJsonSetParams(JsonSetCondition condition) {
+		return switch (condition.getPathCondition()) {
+			case UPSERT -> new JsonSetParams();
+			case IF_PATH_NOT_EXISTS -> new JsonSetParams().nx();
+			case IF_PATH_EXISTS -> new JsonSetParams().xx();
+		};
+	}
+
+	public static RedisJsonCommands.JsonType fromJsonType(Class<?> clazz) {
+		if (clazz == String.class) return RedisJsonCommands.JsonType.STRING;
```
mp911de: this can be also `CharSequence`, the checks should be reworked.

src/main/java/org/springframework/data/redis/connection/jedis/JedisConverters.java
```diff
@@ -876,6 +868,23 @@ public static HostAndPort toHostAndPort(RedisNode node) {
...
+		if (clazz == int.class || clazz == float.class) return RedisJsonCommands.JsonType.NUMBER;
+		if (clazz == boolean.class) return RedisJsonCommands.JsonType.BOOLEAN;
+		if (clazz == Object.class) return RedisJsonCommands.JsonType.OBJECT;
+		if (clazz == List.class) return RedisJsonCommands.JsonType.ARRAY;
```
mp911de: same for instance of Collection

src/main/java/org/springframework/data/redis/core/RedisJsonTemplate.java
viktoriya-kutsarova: After discussing with @mp911de and @Dgramada, we agreed to mark transaction support as out of scope for this feature. @Dgramada will remove the related code.

We also agreed that RedisJsonTemplate should default to a string serializer for keys and RedisJsonSerializer for values (the actual JSON). @Dgramada explored the proposed constructor approach, but it led to this line:

`this.keySerializer = (RedisSerializer<K>) RedisSerializer.string();`

which would break if K is anything other than String. Following @mp911de's suggestion, we'll explore a static factory method next and see whether it surfaces any other issues.

src/main/java/org/springframework/data/redis/core/RedisJsonTemplate.java
viktoriya-kutsarova: We clarified with @mp911de that `RedisJsonOperations` should be an interface extending `JsonOperations`, implemented by `RedisJsonTemplate<K>`. This would let us simplify calls like `jsonTemplate.opsForJson().values(...)` to `jsonTemplate.values(...)`. @Dgramada will explore the implementation and determine whether DefaultJsonOperations (and related classes) can be removed.

src/main/java/org/springframework/data/redis/core/RedisJsonTemplate.java
```diff
@@ -0,0 +1,554 @@
...
+		Assert.notNull(jsonSerializer, "JsonSerializer must not be null");
+
+		setConnectionFactory(connectionFactory);
+		this.keySerializer = keySerializer;
+		this.jsonSerializer = jsonSerializer;
+	}
+
+	static RedisJsonSerializer defaultJsonSerializer() {
+
+		if (!ClassUtils.isPresent("tools.jackson.databind.ObjectMapper", RedisJsonTemplate.class.getClassLoader())) {
```
viktoriya-kutsarova: This presence check only detects Jackson 3. Should we also detect Jackson 2 via com.fasterxml.jackson.databind.ObjectMapper and fall back to GenericJackson2JsonRedisSerializer?

mp911de: yes, please. we do have support for Jackson 2 and so we should pick it up.

mhalbritter:
Just a note from my side:

`RedisJsonCommands` has no commands for `JSON.OBJKEYS`, `JSON.OBJLEN`, `JSON.NUMINCRBY` or `JSON.ARRPOP`. This could be something worth considering to add in the future, after this PR has been merged.

Dgramada:
Hi @mhalbritter. Thank you for pointing this out. The commands were omitted on purpose because of some issues with the Jedis driver. We are planning on improving the Jedis JSON API and possibly adding the leftover commands once Jedis releases the API improvements.

mp911de:

PR #3397, opened 2026-07-10
Fix NullPointerException resolving type hints for JSON with duplicate properties

### LINKED ISSUE #3396: `GenericJacksonJsonRedisSerializer` throws NPE deserializing duplicate properties
## Summary

`GenericJacksonJsonRedisSerializer` (the Jackson 3 serializer, with default typing enabled) throws a `NullPointerException` when deserializing **any** JSON object that contains a duplicate property name. The failure is in `GenericJacksonJsonRedisSerializer.TypeResolver.readTree(byte[])`, which builds a `JsonNode` tree using a `DeserializationContext` obtained from `mapper._deserializationContext()` — a context that is never bound to the parser, so its `_readCapabilities` is `null`. As soon as the tree contains a duplicate key, `BaseNodeDeserializer._handleDuplicateProperty` calls `ctxt.isEnabled(StreamReadCapability.DUPLICATE_PROPERTIES)` and NPEs.

Duplicate names are permitted by JSON (RFC 8259) and Jackson's own tree deserializer normally handles them gracefully (last value wins, unless `FAIL_ON_READING_DUP_TREE_KEY` is set). Here they crash instead.

## Affected versions

- spring-data-redis **4.1.0** (Spring Data BOM 2026.0.0), Jackson **3.2.0** (`tools.jackson`).
- The offending code is unchanged on `main` (`TypeResolver.readTree` still uses `mapper._deserializationContext()` directly).

## Minimal reproduction

No custom mapper, no special types — a default-typing serializer and a two-line JSON with a duplicate key:

```java
import org.springframework.data.redis.serializer.GenericJacksonJsonRedisSerializer;
import tools.jackson.databind.jsontype.BasicPolymorphicTypeValidator;

var validator = BasicPolymorphicTypeValidator.builder().allowIfSubType(Object.class).build();
var serializer = GenericJacksonJsonRedisSerializer.builder()
        .enableDefaultTyping(validator)
        .build();

// Any JSON object with a duplicate property name triggers it:
serializer.deserialize("{\"a\":1,\"a\":2}".getBytes());
```

## Actual behaviour

```
org.springframework.data.redis.serializer.SerializationException: Could not read JSON:Cannot invoke
"tools.jackson.core.util.JacksonFeatureSet.isEnabled(tools.jackson.core.util.JacksonFeature)" because "this._readCapabilities" is null
Caused by: java.lang.NullPointerException: Cannot invoke
"tools.jackson.core.util.JacksonFeatureSet.isEnabled(tools.jackson.core.util.JacksonFeature)" because "this._readCapabilities" is null
    at tools.jackson.databind.DeserializationContext.isEnabled(DeserializationContext.java:470)
    at tools.jackson.databind.deser.jackson.BaseNodeDeserializer._handleDuplicateProperty(BaseNodeDeserializer.java:148)
    at tools.jackson.databind.deser.jackson.BaseNodeDeserializer._deserializeContainerNoRecursion(BaseNodeDeserializer.java:375)
    at tools.jackson.databind.deser.jackson.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:94)
    at tools.jackson.databind.deser.jackson.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:12)
    at org.springframework.data.redis.serializer.GenericJacksonJsonRedisSerializer$TypeResolver.readTree(GenericJacksonJsonRedisSerializer.java:514)
    at org.springframework.data.redis.serializer.GenericJacksonJsonRedisSerializer$TypeResolver.resolveType(GenericJacksonJsonRedisSerializer.java:478)
    at org.springframework.data.redis.serializer.GenericJacksonJsonRedisSerializer.resolveType(GenericJacksonJsonRedisSerializer.java:222)
    at org.springframework.data.redis.serializer.GenericJacksonJsonRedisSerializer.deserialize(GenericJacksonJsonRedisSerializer.java:210)
```

## Expected behaviour

The value should deserialize the same way Jackson's tree reader would (last value wins by default), or at least raise a meaningful `SerializationException` — not a `NullPointerException` from an uninitialized `DeserializationContext`.

## Root cause

`TypeResolver.readTree(byte[])`:

```java
private JsonNode readTree(byte[] source) throws IOException {
    BaseNodeDeserializer<?> deserializer = JsonNodeDeserializer.getDeserializer(JsonNode.class);
    DeserializationConfig cfg = mapper.deserializationConfig();
    try (JsonParser parser = createParser(source)) {
        JsonToken t = parser.currentToken();
        if (t == null) {
            t = parser.nextToken();
            if (t == null) {
                return cfg.getNodeFactory().missingNode();
            }
        }
        DeserializationContext ctxt = mapper._deserializationContext();   // <-- not bound to `parser`
        if (t == JsonToken.VALUE_NULL) {
            return cfg.getNodeFactory().nullNode();
        } else {
            return deserializer.deserialize(parser, ctxt);
        }
    }
}
```

`mapper._deserializationContext()` returns a context whose per-read state (including `_readCapabilities`, normally derived from `JsonParser.streamReadCapabilities()`) has not been initialized, because the parser is never assigned to it. Any code path that consults a `StreamReadCapability` then dereferences `null` — `_handleDuplicateProperty` is the first place that happens, so the NPE only surfaces for input containing a duplicate property. A context created and bound to the parser through the normal read path would have `_readCapabilities` populated and would apply the existing duplicate-key handling.

## Suggested fix

Bind the parser to the deserialization context before deserializing (so stream-read capabilities are populated), or read the tree through a parser-aware path (e.g. an `ObjectReader`/`ObjectMapper.readTree(JsonParser)` variant) rather than invoking `deserializer.deserialize(parser, ctxt)` with an unbound `mapper._deserializationContext()`.

## Impact

Any cached value whose JSON contains a duplicate property crashes on read instead of being handled. This is easy to hit in practice, for example:

- values written by an earlier Jackson 2 build (`GenericJackson2JsonRedisSerializer`), which was more lenient about emitting duplicate keys; and
- polymorphic types where the `@JsonTypeInfo(include = As.PROPERTY, property = "…")` type-id name coincides with a bean property of the same name (Jackson 2 emitted both as duplicate keys). Such entries can no longer be read back — even to be evicted — under the Jackson 3 serializer.

ctomc:
## Problem

`GenericJacksonJsonRedisSerializer` (Jackson 3, default typing enabled) throws a `NullPointerException` when deserializing any value whose JSON contains a duplicate property name.

`TypeResolver.readTree(byte[])` builds the JSON node tree with a `DeserializationContext` obtained from `ObjectMapper._deserializationContext()`, which is never bound to the parser. Its stream-read capabilities therefore stay `null`, and the first time `BaseNodeDeserializer._handleDuplicateProperty` consults `StreamReadCapability.DUPLICATE_PROPERTIES` it fails:

```
java.lang.NullPointerException: Cannot invoke
"tools.jackson.core.util.JacksonFeatureSet.isEnabled(...)" because "this._readCapabilities" is null
    at tools.jackson.databind.DeserializationContext.isEnabled(DeserializationContext.java:470)
    at tools.jackson.databind.deser.jackson.BaseNodeDeserializer._handleDuplicateProperty(BaseNodeDeserializer.java:148)
    at ...GenericJacksonJsonRedisSerializer$TypeResolver.readTree(GenericJacksonJsonRedisSerializer.java:509)
```

Duplicate property names are valid JSON (RFC 8259) and occur in practice — for example in cache entries written by the Jackson 2 serializer, or by a polymorphic type whose `@JsonTypeInfo(include = As.PROPERTY)` type-id name coincides with a bean property of the same name (Jackson 2 emitted both as duplicate keys). Such entries can no longer be read back under the Jackson 3 serializer — not even to be evicted.

Reported in #3396, with a minimal reproduction there.

## Fix

Bind the parser to the context before deserializing, so its stream-read capabilities are initialized. This mirrors how `ObjectMapper._readTreeAndClose(...)` binds the parser via `assignAndReturnParser(...)`. The tree scan then applies Jackson's normal duplicate-property handling (last value wins) instead of dereferencing a `null`.

## Test

`GenericJacksonJsonRedisSerializerUnitTests.deserializeShouldHandleDuplicatePropertyWithDefaultTyping` deserializes a value containing a duplicate property. It reproduces the `NullPointerException` above without the fix and passes with it.

---

- [x] I have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x] There is a test case backing the change; no formatting-only changes are included.
- [x] I added myself as an author in the headers of the classes I touched.

Closes #3396

### DIFF
--- a/src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
+++ b/src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
@@ -23,7 +23,6 @@ import tools.jackson.core.TreeNode;
 import tools.jackson.core.Version;
 import tools.jackson.databind.DefaultTyping;
 import tools.jackson.databind.DeserializationConfig;
-import tools.jackson.databind.DeserializationContext;
 import tools.jackson.databind.DeserializationFeature;
 import tools.jackson.databind.JacksonModule;
 import tools.jackson.databind.JavaType;
@@ -31,6 +30,7 @@ import tools.jackson.databind.JsonNode;
 import tools.jackson.databind.ObjectMapper;
 import tools.jackson.databind.SerializationContext;
 import tools.jackson.databind.cfg.MapperBuilder;
+import tools.jackson.databind.deser.DeserializationContextExt;
 import tools.jackson.databind.deser.jackson.BaseNodeDeserializer;
 import tools.jackson.databind.deser.jackson.JsonNodeDeserializer;
 import tools.jackson.databind.json.JsonMapper;
@@ -68,6 +68,7 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
  * {@link JacksonObjectWriter}.
  *
  * @author Christoph Strobl
+ * @author Tomaz Cerar
  * @see JacksonObjectReader
  * @see JacksonObjectWriter
  * @see ObjectMapper
@@ -503,10 +504,12 @@ public class GenericJacksonJsonRedisSerializer implements RedisSerializer<Object
 					}
 				}
 
-				/*
-				 * Hokey pokey! Oh my.
-				 */
-				DeserializationContext ctxt = mapper._deserializationContext();
+				// Bind the parser to the context so its stream-read capabilities are initialized: the node
+				// deserializer consults them (e.g. StreamReadCapability.DUPLICATE_PROPERTIES when a duplicate
+				// field name is encountered) and would otherwise throw a NullPointerException. This mirrors the
+				// way ObjectMapper._readTreeAndClose(...) binds the parser via assignAndReturnParser(...).
+				DeserializationContextExt ctxt = mapper._deserializationContext();
+				ctxt.assignParser(parser);
 
 				if (t == JsonToken.VALUE_NULL) {
 					return cfg.getNodeFactory().nullNode();
--- a/src/test/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializerUnitTests.java
+++ b/src/test/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializerUnitTests.java
@@ -17,6 +17,7 @@ package org.springframework.data.redis.serializer;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatExceptionOfType;
+import static org.assertj.core.api.InstanceOfAssertFactories.map;
 import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
 import static org.springframework.test.util.ReflectionTestUtils.*;
@@ -60,6 +61,7 @@ import com.fasterxml.jackson.annotation.JsonView;
  * @author Christoph Strobl
  * @author Mark Paluch
  * @author John Blum
+ * @author Tomaz Cerar
  */
 class GenericJacksonJsonRedisSerializerUnitTests {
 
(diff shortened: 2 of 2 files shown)


mp911de:
Thank you for your contribution. However, I notice several aspects that suggest this change hasn't been thoroughly engineered before submission. The changes appear to lack context for our specific codebase patterns, and there are inconsistencies that would have been caught otherwise with a proper design approach.
We appreciate community contributions, but we do need them to meet a reasonable baseline for consideration.

> [!TIP]
> This means:
> * Understanding the existing code structure and following established patterns
> * Providing clear rationale for why the change with its specific design improves the framework
> * Ensuring the contribution aligns with Spring's design principles
> * Avoiding code pollution with excessive inline-comments.

When contributions require significant rework from our side, we have to deprioritize or even reject them in favor of changes that are ready for integration as we have to be mindful of our limited time, and we need to focus it where it has the most impact. If you would like us to spend some time working on this pull request, please spend some time making this a substantial and quality contribution.

When describing your changes, please focus on the why rather than the what. We can see what changed from the diff itself, there's no need to enumerate or summarize those changes in prose.

> [!IMPORTANT]
> Instead, help us understand:
>
> * Your reasoning for the approach you've chosen
> * Trade-offs you considered
> * Any areas where you're uncertain and would value guidance

Concise, thoughtful explanations of your decision-making process are far more valuable than detailed restatements of the code changes themselves.

PR #3398, opened 2026-07-11
Fix UnifiedJedis watch transaction state

### LINKED ISSUE #3392: Transaction with a read before MULTI fails with "Incorrect number of transaction results"
### Summary

Upgrading `spring-data-redis` from 4.0.x to **4.1.0** breaks the canonical optimistic-locking transaction pattern (`WATCH` → read → `MULTI` → queue writes → `EXEC`). Any transaction that issues a read command **between `WATCH` and `MULTI`** now throws:

```
java.lang.IllegalArgumentException: Incorrect number of transaction results; Expected: 3 Actual: 2
```

wrapped as `org.springframework.data.redis.RedisSystemException`. The same code works on 4.0.x. Reverting only `spring-data-redis` to 4.0.x fixes it with no other change.

### Affected versions

- **Broken:** 4.1.0
- **Last known good:** 4.0.5 (and earlier 4.0.x)
- Reproduced with the **Jedis** driver (`jedis` 7.5.2). The exception originates in the shared `TransactionResultConverter`, so Lettuce may also be affected — not verified.
- Redis server version is irrelevant (reproduced on Redis 6.2, 7, and 8.x).

### Reproduction (via the recommended `SessionCallback` API)

Dependencies: `spring-data-redis:4.1.0`, `redis.clients:jedis:7.5.2`. A standalone Redis on `localhost:6379`.

```java
JedisConnectionFactory factory =
        new JedisConnectionFactory(new RedisStandaloneConfiguration("localhost", 6379));
factory.afterPropertiesSet();
StringRedisTemplate template = new StringRedisTemplate(factory);
template.afterPropertiesSet();

final String key = "cas:" + java.util.UUID.randomUUID();
List<Object> results = template.execute(new SessionCallback<List<Object>>() {
    @Override
    public List<Object> execute(RedisOperations operations) {
        operations.watch(key);
        operations.opsForValue().get(key);      // read current value BEFORE multi()
        operations.multi();
        operations.opsForValue().setIfAbsent(key, "v1");
        operations.expire(key, Duration.ofSeconds(60));
        return operations.exec();
    }
});
```

**Expected** (and the 4.0.x result): `exec()` returns `[true, true]`.
**Actual on 4.1.0:** throws `IllegalArgumentException: Incorrect number of transaction results; Expected: 3 Actual: 2`.

The same failure occurs when using the low-level `RedisConnection` API directly (`connection.watch()` / `stringCommands().get()` / `connection.multi()` / … / `connection.exec()`), so it is not specific to `SessionCallback`.

### Isolation

Holding everything constant and changing only the `spring-data-redis` version (Jedis pinned at 7.5.2, same Redis container):

| spring-data-redis | Result |
|---|---|
| 4.0.5 | `[true, true]` (OK) |
| 4.1.0 | `Incorrect number of transaction results; Expected: 3 Actual: 2` |

Pinning `spring-data-redis` at 4.0.x while upgrading Jedis to 7.5.2 stays green, so the Jedis version is not the cause.

### Analysis

`exec()` returns 2 results (the two queued writes: `setIfAbsent` + `expire`), but the transaction result converter now expects **3**. The extra expected result corresponds to the `GET` issued between `WATCH` and `MULTI`: on 4.1.0 it appears to be counted as a transaction result even though it executes immediately (its value is returned synchronously) and is not part of the `EXEC` reply. The count mismatch then trips `TransactionResultConverter`.

### Possibly related

This may share a root cause with the 4.1.0 command-layer refactor implicated in #3391 (StackOverflowError in `DefaultedRedisConnection.set` on 4.1.0). Both are regressions introduced in 4.1.0 around the `set`/transaction command changes.

### Downstream impact

This surfaced as a critical regression in Apache NiFi 2.10.0 (which bumped to spring-data-redis 4.1.0): the `DetectDuplicate` processor and `Wait`/`Notify` (via the Redis `DistributedMapCacheClient`) both fail. Tracking issue: https://issues.apache.org/jira/browse/NIFI-16043

### Ask

Could you confirm whether this is a regression in the 4.1.0 transaction handling and whether a fix is planned for a 4.1.x patch? Happy to provide a full runnable reproducer repo if useful.

hutiefang76:
Fixes #3392

`UnifiedJedis.watch()` pins a connection by creating a transaction object before `MULTI`. The connection therefore reported queueing too early, so reads between `WATCH` and `MULTI` polluted the expected `EXEC` result count.

This change distinguishes WATCH-only state from an active MULTI transaction. Commands before MULTI run immediately on the pinned connection, while incompatible pipeline, subscription, and cursor operations are rejected.

Tests cover typed and raw reads, EXEC result alignment, pipeline/subscription guards, and SCAN variants. The focused Jedis suite passes 813 tests with 0 failures or errors (64 skipped).

I used AI-assisted review for edge-case analysis and verified the root cause, implementation, and tests locally.

- [x] I have read the Spring Data contribution guidelines.
- [x] I applied the project formatter/checkstyle without unrelated formatting changes.
- [x] I added unit and integration tests.
- [x] I added myself as author in every touched class.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnection.java
@@ -68,6 +68,7 @@ import org.springframework.util.CollectionUtils;
  * @author Dengliming
  * @author John Blum
  * @author Tihomir Mateev
+ * @author Tiefang Hu
  * @see redis.clients.jedis.Jedis
  * @see redis.clients.jedis.RedisClient
  */
@@ -211,6 +212,14 @@ public class JedisConnection extends AbstractRedisConnection {
 
 		return doWithJedis(it -> {
 
+			if (delegate.isWatchOnly()) {
+
+				Response<Object> response = pipelineFunction.apply(JedisInvoker.createCommands(getRequiredTransaction()));
+				Object result = response.get();
+
+				return result != null ? converter.convert(result) : nullDefault.get();
+			}
+
 			if (isQueueing()) {
 
 				Response<Object> response = pipelineFunction.apply(JedisInvoker.createCommands(getRequiredTransaction()));
@@ -310,15 +319,17 @@ public class JedisConnection extends AbstractRedisConnection {
 
 			ProtocolCommand protocolCommand = () -> JedisConverters.toBytes(command);
 
-			if (isQueueing() || isPipelined()) {
+			if (isQueueing() || isPipelined() || isWatchOnly()) {
 
 				CommandArguments arguments = new CommandArguments(protocolCommand).addObjects(args);
 				CommandObject<Object> commandObject = new CommandObject<>(arguments, BuilderFactory.RAW_OBJECT);
 
 				if (isPipelined()) {
 					pipeline(newJedisResult(getRequiredPipeline().executeCommand(commandObject)));
-				} else {
+				} else if (isQueueing()) {
 					transaction(newJedisResult(getRequiredTransaction().executeCommand(commandObject)));
+				} else {
+					return getRequiredTransaction().executeCommand(commandObject).get();
 				}
 				return null;
 			}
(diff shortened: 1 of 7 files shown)


mhalbritter:
Merged, polished and backported. Thanks a lot @hutiefang76 !

PR #3401, opened 2026-07-24
Fix inconsistent Javadoc parameter references

### LINKED ISSUE #3400: Fix inconsistent Javadoc parameter references
I came across a few small Javadoc mismatches while browsing the APIs.

Some comments use parameter names that don’t match the actual method signatures, like `begin` instead of `start`, or `field` where the method uses `key` or `fields`. I also found a couple of `must not be null` notes on primitive parameters and one duplicated pipeline/transaction return note.

Most of these are in `ListOperations`, `ValueOperations`, and `RedisHashCommands`.

There’s no behavioral change here—just a small cleanup to keep the generated API docs accurate and consistent.

leeworms:
## Summary

This PR fixes several small Javadoc inconsistencies where the documentation does not match the corresponding method signatures.

Changes include:

* Replacing outdated parameter references such as `begin` with `start`
* Correcting `field` references where methods use `key` or varargs `fields`
* Removing `must not be null` descriptions from primitive parameters
* Removing a duplicated pipeline/transaction return note

These are documentation-only changes and do not affect runtime behavior.

## Testing

No tests were added because this change only updates Javadocs. I verified that the updated parameter references match the corresponding method signatures.

Closes #3400 

---

* [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
* [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
* [ ] You submit test cases (unit or integration tests) that back your changes.
* [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/ReactiveHashCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/ReactiveHashCommands.java
@@ -50,6 +50,7 @@ import org.springframework.util.Assert;
  * @author Christoph Strobl
  * @author Mark Paluch
  * @author Tihomir Mateev
+ * @author JaeGeun Lee
  * @since 2.0
  */
 public interface ReactiveHashCommands {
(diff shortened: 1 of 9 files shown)


mhalbritter:
Merged, polished and backported. Thank you @leeworms !

PR #3404, opened 2026-07-28
Use Parser-bound `DeserializationContext` for parsing type hints in `TypeResolver`

### LINKED ISSUE #3396: `GenericJacksonJsonRedisSerializer` throws NPE deserializing duplicate properties

mp911de:
We now use the deserialization context bound to the JSON parser instead of using the unbound and generic Mapper `DeserializationContext` that is actually only visible for testing.

Also, use `StdTypeResolverBuilder` if possible to defer `_deserializationContext()` access.

Closes #3396

### DIFF
--- a/src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
+++ b/src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
@@ -244,11 +244,14 @@ public class GenericJacksonJsonRedisSerializer implements RedisSerializer<Object
 
 			DeserializationConfig deserializationConfig = mapper.deserializationConfig();
 
+			var typer = deserializationConfig.getDefaultTyper(null);
+			if (typer instanceof StdTypeResolverBuilder std) {
+				return std.getTypeProperty();
+			}
+
 			JavaType objectType = mapper.getTypeFactory().constructType(Object.class);
-
-			TypeDeserializer typeDeserializer = deserializationConfig.getDefaultTyper(null)
+			TypeDeserializer typeDeserializer = typer
 					.buildTypeDeserializer(mapper._deserializationContext(), objectType, Collections.emptyList());
-
 			return typeDeserializer.getPropertyName();
 		});
 	}
@@ -488,7 +491,7 @@ public class GenericJacksonJsonRedisSerializer implements RedisSerializer<Object
 		/**
 		 * Lenient variant of ObjectMapper._readTreeAndClose using a strict {@link JsonNodeDeserializer}.
 		 */
-		private JsonNode readTree(byte[] source) throws IOException {
+		private JsonNode readTree(byte[] source) {
 
 			BaseNodeDeserializer<?> deserializer = JsonNodeDeserializer.getDeserializer(JsonNode.class);
 			DeserializationConfig cfg = mapper.deserializationConfig();
--- a/src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
+++ b/src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
@@ -230,7 +230,8 @@ public class GenericJacksonJsonRedisSerializer implements RedisSerializer<Object
 
 	private static Lazy<String> newLazyTypeHintPropertyName(ObjectMapper mapper, Lazy<Boolean> defaultTypingEnabled) {
 
-		Lazy<String> configuredTypeDeserializationPropertyName = getConfiguredTypeDeserializationPropertyName(mapper);
+		Supplier<@Nullable String> configuredTypeDeserializationPropertyName = getConfiguredTypeDeserializationPropertyName(
+				mapper);
 
 		Lazy<String> resolvedLazyTypeHintPropertyName = Lazy
 				.of(() -> defaultTypingEnabled.get() ? configuredTypeDeserializationPropertyName.get() : null);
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/redis/serializer/GenericJacksonJsonRedisSerializer.java
```diff
@@ -244,11 +244,14 @@ private static Lazy<String> getConfiguredTypeDeserializationPropertyName(ObjectM
 
 			DeserializationConfig deserializationConfig = mapper.deserializationConfig();
 
-			JavaType objectType = mapper.getTypeFactory().constructType(Object.class);
+			var typer = deserializationConfig.getDefaultTyper(null);
```
mhalbritter, line 247: `var` is not very often used in the codebase. Use explicit type?

mhalbritter: `typer` can be `null`, should we return `null` for the whole Lazy in that case?

mhalbritter:
Reviewed, looks good to me. I have two small nitpicks.

mhalbritter:
Merged and backported. Thanks @mp911de !

PR #3406, opened 2026-07-29
Fix argument validation in RedisData.setTimeToLive(Long, TimeUnit)

JuHyun419:
## Summary
- `RedisData.setTimeToLive(Long, TimeUnit)` had a copy/paste bug: the second `Assert.notNull(...)` call re-checked `timeToLive` instead of validating `timeUnit`.
- As a result, passing a `null` `TimeUnit` skipped the intended fail-fast `IllegalArgumentException` and instead threw an unguarded `NullPointerException` from `TimeUnit.SECONDS.convert(...)`.
- Fixed the check to validate `timeUnit` as the message already stated.
- Added `RedisDataUnitTests` covering: null `TimeUnit` (new behavior), null `timeToLive` (existing behavior), and the regular seconds-conversion path (regression guard).

## Test plan
- [x] New unit tests in `RedisDataUnitTests` pass (`./mvnw test -Dtest=RedisDataUnitTests`)
- [x] `./mvnw compile` / `test-compile` succeed

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/convert/RedisData.java
+++ b/src/main/java/org/springframework/data/redis/core/convert/RedisData.java
@@ -161,7 +161,7 @@ public class RedisData {
 	public void setTimeToLive(Long timeToLive, TimeUnit timeUnit) {
 
 		Assert.notNull(timeToLive, "TTL must not be null when used with TimeUnit");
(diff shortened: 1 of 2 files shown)


mhalbritter:
Merged, polished and backported. Thanks @JuHyun419 !

PR #3408, opened 2026-07-30
Resolve type hint without parsing the entire payload

mhalbritter:
This is my take at solving https://github.com/spring-projects/spring-data-redis/issues/3012.

I implemented it only for the Jackson 3 case, as the Jackson 2 class is deprecated.

It works by using the parser to get tokens until we reach the first `@class` and use that as the class name instead of parsing the whole JSON in a `JsonNode`.

We don't reuse the parsed `JsonNode`, (as the spike in 8aaf501b73a8c74df7db839793d55b6e83720835 does) as this breaks number handling with `double` and `BigDecimal`, see the `org.springframework.data.redis.serializer.GenericJacksonJsonRedisSerializerUnitTests#deserializeRetainsNumberFidelity()` test.

> the loss happens at tree-build time, before any type information is applied. That's precisely why "parse once into a tree, then convert" cannot be made correct by being smarter about types, and why the commit went with a shallow token scan plus a full re-read instead.

That's also the reason why https://github.com/spring-projects/spring-data-redis/pull/3366 is not suitable for us - this was the PR which sent me down this rabbit hole.

When we merge this, we should clarify that #3012 is about Jackson 3, not Jackson 2.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-redis</artifactId>
-	<version>4.2.0-SNAPSHOT</version>
+	<version>4.2.0-GH-3012-SNAPSHOT</version>
 
 	<name>Spring Data Redis</name>
 	<description>Spring Data module for Redis</description>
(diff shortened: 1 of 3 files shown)


mp911de:

PR #3415, opened 2026-08-14
Document setIfAbsent return value

### LINKED ISSUE #3414: Document conditional set return values
The Javadoc for `org.springframework.data.redis.core.ValueOperations#setIfAbsent(K, V)` is:

```java
	/**
	 * Set the {@code value} for {@code key} if {@code key} is absent.
	 *
	 * @param key must not be {@literal null}.
	 * @param value must not be {@literal null}.
	 * @return {@literal null} when used in pipeline / transaction.
	 * @see <a href="https://redis.io/commands/set">Redis Documentation: SET</a>
	 */
```

This does not mention what the method returns when not used in pipeline / transaction.

AzazelSensei:
- [x] You have read the Spring Data contribution guidelines.
- [x] You use the code formatters provided and have them applied to your changes. Don’t submit any formatting related changes.
- [x] You added yourself as author in the headers of the classes you touched.

`setIfAbsent` javadoc only mentioned the pipeline/transaction null case. Outside of that it is true when the key was set and false when it already existed.

Closes #3414.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
+++ b/src/main/java/org/springframework/data/redis/core/BoundValueOperations.java
@@ -33,6 +33,7 @@ import org.springframework.data.redis.core.types.Expiration;
  * @author Christoph Strobl
  * @author Marcin Grzejszczak
  * @author Yordan Tsintsov
+ * @author Abdullah
  * @author JaeGeun Lee
  */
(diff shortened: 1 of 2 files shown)


AzazelSensei:
Closing. I already spent the time once; I'm not spending it waiting too.

PR #3424, opened 2026-09-01
Migrate legacy Jedis Cluster API to `UnifiedJedis`

viktoriya-kutsarova:
Follow-up to #3315, completing the `UnifiedJedis` migration for the cluster path.

The cluster command callbacks were still typed over `Jedis`, and the node resource provider still handed out `Jedis` instances, which is not a `UnifiedJedis` subtype. This PR moves the remaining cluster internals to `UnifiedJedis` and removes the last uses of the deprecated `sendCommand` API.

## Changes

- `JedisClusterCommandCallback` / `JedisMultiKeyClusterCommandCallback` now extend `ClusterCommandCallback<UnifiedJedis, T>` / `MultiKeyClusterCommandCallback<UnifiedJedis, T>`. Source-incompatible for subclasses that implemented callbacks against `Jedis`.
- `JedisClusterNodeResourceProvider` returns a `UnifiedJedisAdapter` wrapping the node `Connection` instead of `Jedis`.
- Commands missing on `UnifiedJedis` (`CLUSTER *`, `BGREWRITEAOF`, `BGSAVE`, `SAVE`, `LASTSAVE`, `CONFIG RESETSTAT`/`REWRITE`, `TIME`, `CLIENT LIST`/`KILL`, `SHUTDOWN`) use `executeCommand(new CommandArguments(...))`; all remaining `sendCommand` calls (deprecated since Jedis 7.4.0) are migrated as well.
- Raw command replies are decoded through `JedisConverters`.
- Cluster `migrate` now applies `MigrateOption` (`COPY`/`REPLACE`) via `MigrateParams`.
- Cluster `jsonMGet` routes through the typed multi-key callback using the per-node client.

## Testing

Existing integration suite passes against local cluster infrastructure; no new tests.

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/jedis/JedisClusterConnection.java
@@ -69,7 +69,7 @@ import org.springframework.util.Assert;
  * {@link RedisClusterConnection} implementation on top of {@link RedisClusterClient}.
  * <p>
  * Uses the native {@link RedisClusterClient} api where possible and falls back to direct node communication using
- * {@link Jedis} where needed.
+ * {@link UnifiedJedis} where needed.
  * <p>
  * Pipelines and transactions are not supported in cluster mode. This class is not Thread-safe and instances should not
  * be shared across threads.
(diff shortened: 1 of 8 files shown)


viktoriya-kutsarova:
I was thinking we might also want to add a few lines in the upgrade guide regarding the change for the callbacks. If someone extended any of the callbacks, they would need to update them to use UnifiedJedis.

```
[[upgrading.3-to-4.jedis-cluster-callbacks]]
=== Jedis Cluster Command Callbacks

As of 4.2, `JedisClusterConnection.JedisClusterCommandCallback` and
`JedisClusterConnection.JedisMultiKeyClusterCommandCallback` extend `ClusterCommandCallback<UnifiedJedis, T>`
and `MultiKeyClusterCommandCallback<UnifiedJedis, T>` instead of their `Jedis`-typed variants.
Update custom callbacks to accept `UnifiedJedis`.
Commands without a `UnifiedJedis` method can be issued through `UnifiedJedis.executeCommand(CommandArguments)`.
```

mp911de:
Thank you for your contribution. That's merged and polished now. I've updated the release notes accordingly.

PR #3428, opened 2026-09-03
Compare raw TTL values as long in `Expirations.TimeToLive`

### LINKED ISSUE #3425: `Expirations.TimeToLive.of` truncates the raw TTL to `int`, misclassifying large TTL values
`Expirations.TimeToLive.of(Number, TimeUnit)` classifies the raw TTL by its lower 32 bits:

```java
return switch (value.intValue()) {
    case -2 -> MISSING;
    case -1 -> PERSISTENT;
    default -> new TimeToLive(value.longValue(), timeUnit);
};
```

Any 64-bit TTL whose low 32 bits equal `-2` or `-1` is returned as the `MISSING`/`PERSISTENT` sentinel. With millisecond precision this is reachable in practice: an `HPTTL` of `4294967294` ms (≈ 49.7 days) is reported as `MISSING`, and `4294967295` ms as `PERSISTENT`. Millisecond values flow into this method from `DefaultHashOperations.getTimeToLive` / `DefaultReactiveHashOperations.getTimeToLive` (via `hpTtl`). `TimeToLive#toString` has the same `(int) raw()` truncation.

Reproducer:

```java
Expirations.TimeToLive ttl = Expirations.TimeToLive.of(4294967294L, TimeUnit.MILLISECONDS);
assertThat(ttl.isMissing()).isFalse(); // fails: reported as MISSING
```

Expected: the sentinel comparison uses the full `long` value. Reproduced on `main` (4.1 line). I will submit a PR with a fix and test.

harrisleesh:
Closes #3425

`TimeToLive.of(Number, TimeUnit)` classified the raw TTL by `value.intValue()`, so any 64-bit TTL whose low 32 bits equal -2 or -1 — e.g. an `HPTTL` of 4294967294 ms (≈ 49.7 days) — was misclassified as the `MISSING`/`PERSISTENT` sentinel. `TimeToLive#toString` had the same `(int) raw()` truncation.

The sentinel comparison now uses the full `long` value, and `toString` uses `isMissing()`/`isPersistent()`. Regression test added (fails on `main`, passes with the fix); `ExpirationsUnitTest` passes 13/13.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/types/Expirations.java
+++ b/src/main/java/org/springframework/data/redis/core/types/Expirations.java
@@ -234,11 +234,17 @@ public class Expirations<K> {
 		 */
 		public static TimeToLive of(Number value, TimeUnit timeUnit) {
 
-			return switch (value.intValue()) {
-				case -2 -> MISSING;
-				case -1 -> PERSISTENT;
-				default -> new TimeToLive(value.longValue(), timeUnit);
-			};
+			long raw = value.longValue();
+
+			if (raw == MISSING.raw()) {
+				return MISSING;
+			}
+
+			if (raw == PERSISTENT.raw()) {
+				return PERSISTENT;
+			}
+
+			return new TimeToLive(raw, timeUnit);
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


mhalbritter:
Thanks @harrisleesh ! This is merged and backported.

PR #3429, opened 2026-09-03
Fix inverted equals in `RedisStreamCommands.TrimOptions`

### LINKED ISSUE #3426: `TrimOptions.equals` is inverted and ignores `limit`, breaking `XAddOptions` equality
`RedisStreamCommands.TrimOptions.equals` (new in 4.1) inverts its comparisons:

```java
if (this.trimStrategy.equals(that.trimStrategy)) {
    return false;
}
if (this.trimOperator.equals(that.trimOperator)) {
    return false;
}
return ObjectUtils.nullSafeEquals(deletionPolicy, that.deletionPolicy);
```

Equal options compare unequal and different options can compare equal. `limit` participates in `hashCode` but is never compared, and `MaxLenTrimStrategy`/`MinIdTrimStrategy` have no value-based `equals`/`hashCode`, so even a non-inverted comparison would fall back to identity.

This also breaks `XAddOptions.equals`, which delegates via `nullSafeEquals(trimOptions, ...)`:

```java
assertThat(XAddOptions.maxlen(10)).isEqualTo(XAddOptions.maxlen(10));                     // fails
assertThat(TrimOptions.maxLen(10).approximate()).isNotEqualTo(TrimOptions.minId("5-0").exact()); // fails
```

Expected: value-based equality considering strategy (threshold), operator, limit and deletion policy. Reproduced on `main`. I will submit a PR with a fix and tests.

harrisleesh:
Closes #3426

`TrimOptions.equals` returned `false` when the trim strategies or operators were **equal** (inverted conditions), never compared `limit` (though it participates in `hashCode`), and `MaxLenTrimStrategy`/`MinIdTrimStrategy` lacked value-based `equals`/`hashCode`, so equality effectively fell back to identity. `XAddOptions.equals` delegates to `TrimOptions` and was broken the same way: `XAddOptions.maxlen(10)` did not equal `XAddOptions.maxlen(10)`, while `TrimOptions.maxLen(10).approximate()` equaled `TrimOptions.minId("5-0").exact()`.

This change negates the comparisons, includes `limit`, and adds value-based `equals`/`hashCode` to both trim strategies. Three regression tests added (all fail on `main`, pass with the fix); the connection package unit tests pass (119 tests).

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
+++ b/src/main/java/org/springframework/data/redis/connection/RedisStreamCommands.java
@@ -141,6 +141,22 @@ public interface RedisStreamCommands {
 			return threshold;
 		}
 
+		@Override
+		public boolean equals(@Nullable Object o) {
+			if (this == o) {
+				return true;
+			}
+			if (!(o instanceof MaxLenTrimStrategy that)) {
+				return false;
+			}
+			return this.threshold == that.threshold;
+		}
+
+		@Override
+		public int hashCode() {
+			return Long.hashCode(threshold);
+		}
+
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


mhalbritter:
Thanks @harrisleesh ! This is merged and backported.

PR #3432, opened 2026-09-05
getExpirationDuration and getExpirationInstant return wrong result if asked for milliseconds

renechoi:
`getExpirationInstant` and `getExpirationDuration` convert on the `SECONDS` branch but pass the
raw `getExpirationTime()` on the `MILLISECONDS` one.
`seconds(60).getExpirationDuration(MILLISECONDS)` is `PT0.06S`;
`unixTimestamp(1700000000, SECONDS).getExpirationInstant(MILLISECONDS)` is `1970-01-20T16:13:20Z`.
Both now use `getExpirationTimeInMilliseconds()`.

`equals` normalises to millis, so `milliseconds(60_000)` equals `seconds(60)`, but `hashCode`
hashed the raw time and unit, so a `HashSet` of the two keeps both. `hashCode` now hashes what
`equals` compares.

The only in-tree caller derives precision from `isPrecise()` and never hits that branch.

The three new tests fail on the parent commit with the values above; the other six pass.
431 tests across the unit classes touching `Expiration` are green.

### DIFF
--- a/src/main/java/org/springframework/data/redis/core/types/Expiration.java
+++ b/src/main/java/org/springframework/data/redis/core/types/Expiration.java
@@ -17,7 +17,6 @@ package org.springframework.data.redis.core.types;
 
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


mhalbritter:
Thanks for the PR @renechoi, and congrats on your 1st contribution!

This is merged and backported now.

PR #3433, opened 2026-09-07
Refactor Redis JSON API

mhalbritter:
Split JsonResult into JsonResult and JsonPathResult, decouple JsonResult from JsonValue, add exists() accessors, and let GenericJackson2JsonRedisSerializer and GenericJacksonJsonRedisSerializer implement RedisJsonSerializer with raw- value slicing.

Also adds Antora documentation.


src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
```diff
@@ -370,6 +414,144 @@ private JavaType resolveType(byte[] source, ResolvableType type) throws IOExcept
 		}
 	}
 
+	@Override
+	public List<byte[]> splitArray(byte[] source) throws SerializationException {
```
mhalbritter, line 412: This here is interesting. We generally try to slice on the bytes, by getting a parser and finding the offsets. However, in some cases Jackson doesn't give us the offsets. We try to rebuild the parser from the given mapper and make sure the features we need are enabled (see `rawValueFactory`). The features we need are enabled by default, but a user could have deactivated them.

If we detect that we can't get the byte offsets, we fall back to the interface default implementation, which does a deserialization/serialization. This works with every mapper, but loses some benefits the raw byte slicing has.

That's the same for Jackson 2 and Jackson 3.

src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
```diff
@@ -370,6 +414,144 @@ private JavaType resolveType(byte[] source, ResolvableType type) throws IOExcept
...
+		return rawValueFactory.get().createParser(source);
+	}
+
+	/**
+	 * Probe whether {@link #createRawValueParser} reports byte offsets into its source, by parsing a document whose
+	 * offsets are known. See {@link #canSliceRawValues}.
+	 */
+	private boolean probeRawValueOffsets() {
+
+		try (JsonParser parser = createRawValueParser(OFFSET_PROBE)) {
```
mhalbritter, line 502: This here does the detection if we can get accurate byte offsets.

src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java
```diff
@@ -370,6 +414,144 @@ private JavaType resolveType(byte[] source, ResolvableType type) throws IOExcept
...
+			parser.skipChildren();
+		}
+		parser.finishToken();
+
+		long end = parser.currentLocation().getByteOffset();
+
+		// The check for the one thing that probe cannot see: a payload that
+		// is not UTF-8, for which Jackson picks a char-based parser reporting no byte offsets.\
+		// Every RedisJSON reply is UTF-8.
+		if (start < 0 || end < start || end > source.length) {
```
mhalbritter, line 542: Even if we get offsets, when someone passes in JSON which is not UTF-8, the offsets are wrong. This detects that, but should in practice never fire, as Redis gives us UTF-8 json.

mp911de:

PR #3435, opened 2026-09-08
Add support for XAUTOCLAIM to stream operations.

### LINKED ISSUE #2316: Add support for XAUTOCLAIM to stream operations
hi. I'm using Spring Boot 2.6.6. right now, in order to run the [xautoclaim](https://redis.io/commands/xautoclaim/) command, I have to get the native lettuce client and use a lettuce command interface . I would prefer to use Spring's higher-level Reactive/StreamOperations interface, but xautoclaim is not currently supported. please add Reactive/StreamOperations support for xautoclaim. on my Spring version, regular [xclaim](https://redis.io/commands/xclaim/) is also not supported in Reactive/StreamOperations, but it's supported in Reactive/StreamCommands. if, for some reason, there is less resistance to add xautoclaim at the Reactive/StreamCommands level, that would also be preferable to what I'm currently doing. thanks!

### LINKED ISSUE #3434: Add support for XAUTOCLAIM to stream operations
Hello!

Redis 6.2 added [XAUTOCLAIM](https://redis.io/docs/latest/commands/xautoclaim/), which transfers ownership of pending entries that have been idle for longer than a given time, scanning the pending entries list with a cursor instead of requiring the caller to name the record ids up front.

Spring Data Redis exposes `XCLAIM` (`RedisStreamCommands#xClaim`, `StreamOperations#claim`) but not `XAUTOCLAIM`. To recover messages left behind by a dead consumer today you either call `XPENDING` first and feed the ids into `claim`, or drop out of the abstraction entirely:

```java
RedisConnection connection = factory.getConnection();
var nativeCommands = (RedisClusterAsyncCommands<byte[], byte[]>) connection.getNativeConnection();

ClaimedMessages<byte[], byte[]> claimed = nativeCommands
        .xautoclaim("orders".getBytes(),
                XAutoClaimArgs.Builder.xautoclaim(consumer, Duration.ofMinutes(5), "0-0"))
        .get();
```

That ties the recovery code to one driver and bypasses the serializers configured on the template.

Both drivers pinned in `pom.xml` already support the command:

| | cursor | claimed entries |
|---|---|---|
| Lettuce 7.7.0.RELEASE, `xautoclaim(K, XAutoClaimArgs<K>)` -> `ClaimedMessages` | yes | yes |
| Jedis 8.0.1, `xautoclaim` / `xautoclaimJustId` -> `Map.Entry<StreamEntryID, ...>` | yes | yes |

Feature request: expose `XAUTOCLAIM` alongside the existing claim methods, on `RedisStreamCommands` / `ReactiveStreamCommands` and on `StreamOperations` / `ReactiveStreamOperations`, for both drivers.

One design note, since it affects the signatures. `XAUTOCLAIM` returns a cursor in addition to the claimed entries, and a cursor of `0-0` is what tells the caller the scan is finished. Returning a bare `List` or `Flux` of records the way `claim` does would drop that, so I would return a small value object holding the next cursor plus the claimed records, with a `JUSTID` variant holding the ids, and have the reactive side return `Mono<...>` rather than `Flux<...>` so the cursor survives.

Out of scope: the third reply element added in Redis 7.0, the ids of pending entries that no longer exist in the stream. Neither pinned driver exposes it today. The Lettuce side is fixed in redis/lettuce#3901 and ships in 7.8, so I would leave it out for now rather than add an accessor that is always empty, and add it once the driver is upgraded here.

big-cir:
Spring Data Redis exposes `XCLAIM` but not `XAUTOCLAIM`. Recovering unacknowledged messages from a failed consumer currently requires either issuing `XPENDING` first and passing the ids to `claim`, or unwrapping `RedisConnection#getNativeConnection()` to call the underlying driver directly. The latter ties the recovery logic to a specific driver and bypasses the serializers configured on the template.

Both driver versions declared in `pom.xml` already support the command, so this change is only about exposing it: Lettuce 7.7.0 through `xautoclaim(K, XAutoClaimArgs<K>)` returning `ClaimedMessages`, and Jedis 8.0.1 through `xautoclaim` and `xautoclaimJustId`.

`XAutoClaimOptions` holds the min-idle-time, the start id and an optional `COUNT`, mirroring how `XClaimOptions` sits next to `xClaim`.

### Return type

`XAUTOCLAIM` returns a cursor in addition to the claimed entries, and a cursor of `0-0` is what tells the caller the scan is finished. Returning a bare `List` or `Flux` the way `claim` does would discard that, so the command returns `ClaimedRecords` (cursor plus records) and `ClaimedRecordIds` for the `JUSTID` variant. Both implement `Streamable`, like `PendingMessages`. For the same reason the reactive methods return `Mono<ClaimedRecords<...>>` rather than a `Flux` of records. This was the open design question in #3434, so feedback on the shape is welcome.

### Out of scope

The third reply element added in Redis 7.0, the ids of pending entries that no longer exist in the stream, is not exposed. Neither driver reports it today. The Lettuce side is fixed in redis/lettuce#3901 and ships in 7.8, so I left it out rather than add an accessor that is always empty. `ClaimedRecords` and `ClaimedRecordIds` can take the extra list later without changing the existing signatures.

### Driver coverage

Both Lettuce and Jedis, imperative and reactive: `LettuceStreamCommands`, `LettuceReactiveStreamCommands`, and `JedisStreamCommands` (inherited by `JedisClusterStreamCommands`), with conversion handled in the respective `StreamConverters`. The Jedis converter filters out nil entries, which Redis versions before 7.0 return in place of pending entries whose message was deleted.

### Backward compatibility

`xAutoClaim` and `xAutoClaimJustId` are declared as abstract methods on `RedisStreamCommands`, `ReactiveStreamCommands` and `StringRedisConnection`, following the pattern set by `xAckDel` and `xDelEx` in #3247. Custom implementations of those interfaces will need to implement the new methods.

### Documentation and commits

No update to `appendix.adoc` is necessary, as `XAUTOCLAIM` is already listed there as supported, which this change makes true.

Split into two commits, the connection contract first and the template API that builds on it second, to make review easier. Each commit carries its own tests and compiles on its own. Happy to squash if you prefer a single commit.

### Verification

Verified locally against Redis 8.6.2 via `make start` across both drivers, standalone and cluster, imperative and reactive, including the pipeline and transaction paths. 3003 tests across `DefaultStreamOperationsIntegrationTests`, `DefaultReactiveStreamOperationsIntegrationTests`, `LettuceReactiveStreamCommandsIntegrationTests`, the `AbstractConnectionIntegrationTests` subclasses, the converter unit tests and the `DefaultStringRedisConnection` unit tests.

The two failures observed, `xAckDelShouldAcknowledgeAndDeleteEntries` and `xAckDelWithStringIdsShouldWork`, reproduce on unmodified `main` in the same environment and are unrelated to this change. New tests cover the claim itself, cursor pagination with `COUNT` across two calls until the cursor returns to `0-0`, the `JUSTID` variant, and the raw reply shapes of both drivers. Since this introduces new API, those tests do not compile before the change rather than failing against it.

Closes #3434

---
---

### DIFF
--- a/src/main/java/org/springframework/data/redis/connection/DefaultStringRedisConnection.java
+++ b/src/main/java/org/springframework/data/redis/connection/DefaultStringRedisConnection.java
@@ -38,6 +38,8 @@ import org.springframework.data.redis.connection.convert.ListConverter;
 import org.springframework.data.redis.connection.convert.MapConverter;
 import org.springframework.data.redis.connection.convert.SetConverter;
 import org.springframework.data.redis.connection.stream.ByteRecord;
+import org.springframework.data.redis.connection.stream.ClaimedRecordIds;
+import org.springframework.data.redis.connection.stream.ClaimedRecords;
 import org.springframework.data.redis.connection.stream.Consumer;
 import org.springframework.data.redis.connection.stream.MapRecord;
 import org.springframework.data.redis.connection.stream.PendingMessages;
@@ -85,6 +87,7 @@ import org.springframework.util.ObjectUtils;
  * @author Jeonggyu Choi
  * @author Mingi Lee
  * @author Yordan Tsintsov
+ * @author big-cir
  */
 @NullUnmarked
 @SuppressWarnings({ "ConstantConditions", "deprecation" })
@@ -115,6 +118,9 @@ public class DefaultStringRedisConnection implements StringRedisConnection, Deco
 
 	private ListConverter<ByteRecord, StringRecord> listByteMapRecordToStringMapRecordConverter = new ListConverter<>(
 			byteMapRecordToStringMapRecordConverter);
+
+	private Converter<ClaimedRecords<ByteRecord>, ClaimedRecords<StringRecord>> claimedByteRecordsToClaimedStringRecordsConverter = source -> source
+			.mapRecords(byteMapRecordToStringMapRecordConverter::convert);
 
 	@SuppressWarnings("rawtypes") private Queue<Converter> pipelineConverters = new LinkedList<>();
 	@SuppressWarnings("rawtypes") private Queue<Converter> txConverters = new LinkedList<>();
@@ -2950,6 +2956,19 @@ public class DefaultStringRedisConnection implements StringRedisConnection, Deco
 	}
 
 	@Override
+	public ClaimedRecordIds xAutoClaimJustId(String key, String group, String consumer, XAutoClaimOptions options) {
+		return convertAndReturn(delegate.xAutoClaimJustId(serialize(key), group, consumer, options),
+				Converters.identityConverter());
+	}
+
+	@Override
+	public ClaimedRecords<StringRecord> xAutoClaim(String key, String group, String consumer,
+			XAutoClaimOptions options) {
+		return convertAndReturn(delegate.xAutoClaim(serialize(key), group, consumer, options),
+				claimedByteRecordsToClaimedStringRecordsConverter);
+	}
+
+	@Override
 	public Long xDel(String key, RecordId... recordIds) {
 		return convertAndReturn(delegate.xDel(serialize(key), recordIds), Converters.identityConverter());
 	}
@@ -3105,6 +3124,17 @@ public class DefaultStringRedisConnection implements StringRedisConnection, Deco
 	@Override
 	public List<ByteRecord> xClaim(byte[] key, String group, String newOwner, XClaimOptions options) {
 		return delegate.xClaim(key, group, newOwner, options);
+	}
+
+	@Override
+	public ClaimedRecordIds xAutoClaimJustId(byte[] key, String group, String newOwner, XAutoClaimOptions options) {
+		return delegate.xAutoClaimJustId(key, group, newOwner, options);
+	}
+
+	@Override
+	public ClaimedRecords<ByteRecord> xAutoClaim(byte[] key, String group, String newOwner,
+			XAutoClaimOptions options) {
+		return delegate.xAutoClaim(key, group, newOwner, options);
 	}
 
 	@Override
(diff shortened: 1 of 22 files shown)


renechoi (commented):
Ran the branch against Redis 8.6.2 with the pinned drivers. The Lettuce comment on `xAutoClaimJustId` describes something users will hit: the two drivers return different ids once a claimed entry no longer exists in the stream.

3 entries pending for `c1`, middle one `XDEL`ed, then autoclaim to `c2`:

- Lettuce 7.7.0 -> `[a, c, b]`, and `b` is gone from the stream
- Jedis 8.0.1 -> `[a, c]`

redis/lettuce#3901 is merged but not in 7.7.0. Both new JUSTID tests use a single entry and no `XDEL`, so they stay green either way, while `getIds()` is documented as the ids that were claimed.

big-cir:
> Ran the branch against Redis 8.6.2 with the pinned drivers. The Lettuce comment on `xAutoClaimJustId` describes something users will hit: the two drivers return different ids once a claimed entry no longer exists in the stream.
> 

Thanks for running it, I reproduced exactly what you describe. Three entries pending for `c1`, middle one `XDEL`ed, autoclaim to `c2`:

- Lettuce 7.7.0: `[a, c, b]`, with `b` the deleted one
- Jedis 8.0.1: `[a, c]`
- non-JUSTID `autoClaim`: `[a, c]` on both

You are right that the tests do not catch it. All four JUSTID tests use a single entry and no `XDEL`.

This cannot be fixed inside Spring Data Redis. With `JUSTID`, `ClaimedMessagesOutput` appends the third reply element to the same list as the claimed ids, so by the time the converter sees it the boundary is gone. Calling the non-JUSTID form and mapping records to ids would return the right ids, but `JUSTID` is also the variant that does not increment the delivery counter, so that would trade one divergence for another.

redis/lettuce#3901 fixes this along with adding `getDeletedIds()`, so both this and the deferred deleted ids resolve together once the driver here moves to 7.8. It was merged on 2026-08-25, after the 7.7.0 release, and 7.8 is not out yet.

PR #3440, opened 2026-09-13
Check for integer overflow in sort LIMIT

anjeongkyun:
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another year only).

## Motivation

`SortParameters.Range` keeps `start` and `count` as `long`, and `JedisConverters.toSortingParams` passed both to Jedis as `int`:

```java
Range limit = params.getLimit();
if (limit != null) {
    jedisParams.limit((int) limit.getStart(), (int) limit.getCount());
}
```

A count above `Integer.MAX_VALUE` wraps, and a different LIMIT reaches Redis with nothing reported. Asking for three billion:

```java
DefaultSortParameters params = new DefaultSortParameters();
params.limit(0, 3_000_000_000L);
JedisConverters.toSortingParams(params);
```

sends `LIMIT 0 -1294967296`.

`DefaultSortParameters.limit(long, long)` and `SortQuery` both take `long`, so this is reachable from the public API.

## Change

The same guard added for `lPop` and `rPop` in GH-3438, which the sweep around it also applied to `restore`, `sRandMember` and `zRangeByScore`. `toSortingParams` looks like it was the one left over, so I kept the wording consistent with those messages.

Tests are in `JedisConvertersUnitTests` next to the existing converter tests, one for `start` and one for `count`; both fail on current `main`. `Jedis*UnitTests` and the sort tests pass (200 tests).


renechoi:
Three more narrowings of the same kind live in this file. Argv rendered against Jedis 8.0.1 with `Integer.MAX_VALUE + 1`:

- `toScanParams` -> `COUNT -2147483648` (`ScanOptions.count(long)` -> `ScanParams.count(Integer)`)
- `toGeoRadiusParam` -> `COUNT -2147483648` (`GeoRadiusCommandArgs.limit(long)` -> `GeoRadiusParam.count(int)`)
- `toGeoSearchParams` -> `ArithmeticException`, via `Math.toIntExact`
- `toSortingParams` after your fix -> `IllegalArgumentException`

On 7.4.11 the wrapped value does not survive: `SCAN 0 COUNT -2147483648` is `ERR syntax error`, `COUNT 2147483648` is fine.

anjeongkyun:
Thanks, that is a better sweep than mine. I had only grepped for `(int)` casts, so `intValue()` and `Math.toIntExact` slipped past.

Pushed guards for all three. `geosearch` already failed loudly through `Math.toIntExact`, but with an `ArithmeticException`, so I moved it to the same `IllegalArgumentException` the others report rather than leaving one path different.

That makes four of these in `JedisConverters` now. I kept them inline to match the wording of the ones added for `lPop`/`rPop`, but they would fold into a small private helper if you would rather have that.

mhalbritter:
Thanks @anjeongkyun !

